PR #5, opened 2022-10-26
Document i2gw interpretation of Ingress annotations/fields

sunjayBhatia:
Documents some of the *current* interpretation of Ingress fields and annotations, which may change over time

Also documents some "ideals" for conflict resolution, that are open for discussion!

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,35 @@ the following within the repo:
 go run .
 ```
 
+## Conversion of Ingress resources to Gateway API
+
+### Processing Order and Conflicts
+
+Ingress resources will be processed with a defined order to ensure deterministic generated Gateway API configuration.
+This should also determine precedence order of Ingress resources and routes in case of conflicts.
+
+Ingress resources with the oldest creation timestamp will be sorted first and therefore given precedence.
+If creation timestamps are equal, then sorting will be done based on the namespace/name of the resources.
+If an Ingress rule conflicts with another (e.g. same path match but different backends) an error will be reported for the one that sorted later.
+
+Since the Ingress v1 spec does not itself have a conflict resolution guide, we have adopted this one.
+These rules are similar to the [Gateway API conflict resolution guidelines](https://gateway-api.sigs.k8s.io/concepts/guidelines/#conflicts).
+
+### Ingress resource fields to Gateway API fields
+
+Given a set of Ingress resources, `ingress2gateway` will generate a Gateway with various HTTP and HTTPS Listeners as well as HTTPRoutes that should represent equivalent routing rules.
+
+| Ingress Field | Gateway API configuration |
+|---------------|---------------------------|
+| `ingressClassName` | |
+| `defaultBackend` | If present, this configuration will generate a Gateway Listener with no `hostname` specified as well as a catchall HTTPRoute that references this listener. The backend specified here will be translated to a HTTPRoute `rules[].backendRefs[]` element. |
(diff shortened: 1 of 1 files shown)


sunjayBhatia:
Just WIP so far, and added to the README, can move it to somewhere else we think is better if needed!

README.md
```diff
@@ -44,7 +73,7 @@ support for the following annotations:
 * nginx.ingress.kubernetes.io/canary-weight-total
 
 If you are reliant on any annotations not listed above, you'll need to manually
-find a Gateway API equivalent. 
+find a Gateway API equivalent.
```
sunjayBhatia, line 76: still WIP, need to document how annotations should be treated

README.md
```diff
@@ -27,6 +27,35 @@ the following within the repo:
...
+## Conversion of Ingress resources to Gateway API
+
+### Processing Order and Conflicts
+
+Ingress resources will be processed with a defined order to ensure deterministic generated Gateway API configuration.
+This should also determine precedence order of Ingress resources and routes in case of conflicts.
+
+Ingress resources with the oldest creation timestamp will be sorted first and therefore given precedence.
+If creation timestamps are equal, then sorting will be done based on the namespace/name of the resources.
+If an Ingress rule conflicts with another (e.g. same path match but different backends) an error will be reported for the one that sorted later.
```
youngnick, line 39: While really like this process, I feel like I should mention that some users expect the ingress-nginx behavior of merging matching Ingress objects into load-balanced sets. I'm okay with not doing this for now, but we need to remember it for later.

README.md
youngnick: I agree that we shouldn't translate annotations at all - we want to discourage annotation use on Gateway API objects as much as possible.

youngnick (approved):
LGTM

youngnick:
I couldn't remember if we were Prow-ing in this repo. I'll approve and let the next person lgtm

/approve

README.md
```diff
@@ -27,24 +27,53 @@ the following within the repo:
...
-* nginx.ingress.kubernetes.io/canary-by-header
-* nginx.ingress.kubernetes.io/canary-by-header-value
-* nginx.ingress.kubernetes.io/canary-by-header-pattern
-* nginx.ingress.kubernetes.io/canary-weight
+* nginx.ingress.kubernetes.io/canary: If set to `true` will enable weighting backends.
+* nginx.ingress.kubernetes.io/canary-by-header: If specified, the value of this annotation is the header name that will be added as a HTTPHeaderMatch for the routes generated from this Ingress. If not specified, no HTTPHeaderMatch will be generated.
+* nginx.ingress.kubernetes.io/canary-by-header-value: If specified, the value of this annotation is the header value to perform an `HeaderMatchExact` match on in the generated HTTPHeaderMatch.
+* nginx.ingress.kubernetes.io/canary-by-header-pattern: If specified, this is the  pattern to match against for the HTTPHeaderMatch, which will be of type `HeaderMatchRegularExpression`.
+* nginx.ingress.kubernetes.io/canary-weight: If specified and non-zero, this value will be applied as the weight of the backends for the routes generated from this Ingress resource.
 * nginx.ingress.kubernetes.io/canary-weight-total
```
sunjayBhatia, line 73: didn't add anything for this one as I don't see it being used in practice in the code?

robscott: Oh good catch, I'll create an issue to add that.

robscott (approved):
Thanks @sunjayBhatia!

Issue #7, opened 2022-11-08
Create makefile with "make verify"


robscott:
We should have a verify task that runs go tests and golangci-lint. As part of this, we should copy over the following files from Gateway API:

- https://github.com/kubernetes-sigs/gateway-api/blob/main/.golangci.yml
- https://github.com/kubernetes-sigs/gateway-api/blob/main/hack/verify-golint.sh

It may be useful to look at the Makefile from Gateway API when working on this as this one will likely be a subset of that:

- https://github.com/kubernetes-sigs/gateway-api/blob/main/Makefile

shouri007:
I can take this up

shaneutt:
Thanks @shouri007! Let us know if you need any help.

PR #8, opened 2023-02-01
feat: add Makefile for run task

zaunist:
ref issue: https://github.com/kubernetes-sigs/ingress2gateway/issues/7

Add Makefile to run verify task and golangci-lint task.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,47 @@
+run:
+  timeout: 10m
+  issues-exit-code: 1
+  tests: true
+  skip-dirs-use-default: true
+  modules-download-mode: readonly
+  allow-parallel-runners: false
+
+linters:
+  fast: false
+  enable:
+  - gofmt
+  - goimports
+  - revive
+  - govet
+  - misspell
+  - exportloopref
+  disable:
+  - scopelint
+  disable-all: false
+  presets:
+  - bugs
+  - unused
+
+# all available settings of specific linters
+linters-settings:
+  gofmt:
+    # simplify code: gofmt with `-s` option, true by default
+    simplify: true
+  golint:
+    min-confidence: 0.9
+  govet:
+    # report about shadowed variables
+    check-shadowing: true
+  misspell:
+    locale: US
+
+issues:
+  exclude-rules:
+    # Exclude some linters from running on tests files.
+    - path: _test\.go
+      linters:
+        - gocyclo
+        - errcheck
+        - dupl
+  exclude:
+    - Using the variable on range scope `tc` in function literal
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,47 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# We need all the Make variables exported as env vars.
+# Note that the ?= operator works regardless.
+
+# Enable Go modules.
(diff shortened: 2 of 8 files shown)


zaunist:
@shaneutt Please take a look, thanks.

.golangci.yml
```diff
@@ -0,0 +1,47 @@
...
+issues:
+  exclude-rules:
+    # Exclude some linters from running on tests files.
+    - path: _test\.go
+      linters:
+        - gocyclo
+        - errcheck
+        - dupl
+  exclude:
+    - Using the variable on range scope `tc` in function literal
```
shaneutt, line 47: I'm curious, did you copy this `.golangci.yml` from somewhere, or did you hand-select these settings?

Makefile
```diff
@@ -0,0 +1,47 @@
+# Copyright 2019 The Kubernetes Authors.
```
shaneutt, line 1:
```suggestion
# Copyright 2023 The Kubernetes Authors.
```

Makefile
```diff
@@ -0,0 +1,47 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# We need all the Make variables exported as env vars.
+# Note that the ?= operator works regardless.
+
+# Enable Go modules.
+export GO111MODULE=on
+
+all: vet fmt verify test;$(info $(M)...Begin to test and verify this project.) @ ## Test and verify.
```
shaneutt, line 21: I'm not necessarily opposed to having the verification and testing run by default, but the main thing this project attempts to make is a `ingress2gateway` binary, are you up for adding a `build` directive in here at the end of the chain?

.golangci.yml
zaunist: No, I copied it from [here](https://github.com/kubernetes-sigs/gateway-api/blob/main/.golangci.yml), but I removed one of the rules because I didn't think it was working here, so I removed it.

If this is necessary, I can add it back.

https://github.com/kubernetes-sigs/gateway-api/blob/6445cc46d294bc61b44a37315ed51bf1597bcbbd/.golangci.yml#L30

https://github.com/kubernetes-sigs/gateway-api/blob/6445cc46d294bc61b44a37315ed51bf1597bcbbd/.golangci.yml#L39

Makefile
zaunist: Thank you for your careful correction.

Makefile
zaunist: Of course, I began to consider whether to add a `build` command, but I noticed the description in READEME.md: 
```
go run .
```
So I gave up to add `build`. Since adding build is necessary, I'll add it in a later commit

zaunist:
Thanks for your review, I have updated the command executed by default to `help`.

.golangci.yml
zaunist: The reason why I removed `ignore-words` is that my vscode has some lint rules, and `ignore-words` does not take effect here, and it will cause the editor to report an error. 

Of course, if this rule has other effects, it can be restored

shaneutt: No I think this is fine, thank you for the explanation :+1:

Makefile
shaneutt: Awesome, thank you!

Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+# Run go test against code
+.PHONY: test
+test: vet;$(info $(M)...Begin to run tests.)  @ ## Run tests.
+	go test -race -cover ./pkg/...
+
+# Build the binary
+.PHONY: build
+build: vet;$(info $(M)...Build the binary.)  @ ## Build the binary.
+	go build -o  ingress2gateway .
```
shaneutt, line 48:
```suggestion
	go build -o ingress2gateway .
```

Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+# Run static analysis.
+.PHONY: verify
+verify:
+	hack/verify-golint.sh
+
+
```
shaneutt, line 55:
```suggestion
```

shaneutt (approved):
Couple small comments, mostly looking good thank you!

Would like to have @robscott check this out too.

Xunzhuo (commented):
LGTM
I think this is good to go.

shaneutt (approved):
/lgtm

PR #10, opened 2023-03-24
installation: add script to install i2gw easily

Xunzhuo:
/area installation

### DIFF
--- a/tools/hack/install.sh
+++ b/tools/hack/install.sh
@@ -1,0 +1,188 @@
+#!/usr/bin/env bash
+
+# Copyright © 2023 Kubernetes Authors
+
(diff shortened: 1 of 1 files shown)


Xunzhuo:
xref: https://github.com/kubernetes-sigs/ingress2gateway/issues/21

robscott:
I think this is one area both @youngnick and I were pretty hesitant about in the initial PR. I'd rather stick with installing a binary, using go install, or krew, as all of those methods feel safer than running a shell script.

Xunzhuo:
Yes @robscott, I just opened this PR in case if we will need this approach in the future. Or we can close it if we do not consider this way in the future.

PR #31, opened 2023-06-29
Add support for resource filtering by namespace

### LINKED ISSUE #30: Add support for resource filtering by namespace
Add a --namespace / -n flag to support Gateway API object filtering by namespace.
In addition, a --all-namespaces / -A flag to indicate that all namespaces should be used.
If none of the flags is specified, the current namespace will be used.

levikobi:
**What type of PR is this?**

/kind feature
This PR introduces 2 new flags which will enable the user to filter Gateway API objects by namespace.

Fixes #30 

```release-note
Add a --namespace / -n flag to support Gateway API object filtering by namespace.
In addition, a --all-namespaces / -A flag to indicate that all namespaces should be used.
The current namespace will be used if none of the flags is specified.
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -23,12 +23,23 @@ import (
 	"github.com/spf13/cobra"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 var (
 	// outputFormat contains currently set output format. Value assigned via --output/-o flag.
 	// Defaults to YAML.
 	outputFormat = "yaml"
+
+	// namespace contains currently set namespace that should be used. Value assigned via
+	// --namespace/-n flag.
+	// Default behavior is to use the current namespace the user is in.
+	namespace string
+
+	// allNamespaces indicates whether all namespaces should be used. Value assigned via
+	// --all-namespaces/-A flag.
+	// If present, overrides the namespace variable.
+	allNamespaces bool
 )
 
 // printCmd represents the print command. It prints HTTPRoutes and Gateways
@@ -41,7 +52,11 @@ var printCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		i2gw.Run(resourcePrinter)
+		namespaceFilter, err := getNamespaceFilter(namespace, allNamespaces)
+		if err != nil {
+			return err
+		}
+		i2gw.Run(resourcePrinter, namespaceFilter)
 		return nil
 	},
 }
@@ -59,12 +74,46 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
 	}
 }
 
+// getNamespaceFilter returns a namespace filter, taking into consideration whether a specific
+// namespace is requested, or all of them are.
+func getNamespaceFilter(requestedNamespace string, useAllNamespaces bool) (string, error) {
+
+	// When we should use all namespaces, return an empty string.
+	// This is the first condition since it should override the requestedNamespace,
+	// if specified.
+	if useAllNamespaces {
+		return "", nil
+	}
+
+	if requestedNamespace == "" {
+		return getNamespaceInCurrentContext()
+	}
+	return requestedNamespace, nil
+}
+
+// getNamespaceInCurrentContext returns the namespace in the current active context of the user.
+func getNamespaceInCurrentContext() (string, error) {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+	currentNamespace, _, err := kubeConfig.Namespace()
+
+	return currentNamespace, err
+}
+
 func init() {
 	var printFlags genericclioptions.JSONYamlPrintFlags
 	allowedFormats := printFlags.AllowedFormats()
 
 	printCmd.Flags().StringVarP(&outputFormat, "output", "o", "yaml",
 		fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(allowedFormats, ", ")))
+
+	printCmd.Flags().StringVarP(&namespace, "namespace", "n", "",
+		fmt.Sprintf(`If present, the namespace scope for this CLI request`))
+
+	printCmd.Flags().BoolVarP(&allNamespaces, "all-namespaces", "A", false,
+		fmt.Sprintf(`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
+if specified with --namespace.`))
 
 	rootCmd.AddCommand(printCmd)
 }
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -31,14 +31,13 @@ var (
 	// Defaults to YAML.
 	outputFormat = "yaml"
 
-	// namespace contains currently set namespace that should be used. Value assigned via
+	// The namespace used to query Gateway API objects. Value assigned via
 	// --namespace/-n flag.
-	// Default behavior is to use the current namespace the user is in.
+	// On absence, the current user active namespace is used.
 	namespace string
 
 	// allNamespaces indicates whether all namespaces should be used. Value assigned via
 	// --all-namespaces/-A flag.
-	// If present, overrides the namespace variable.
 	allNamespaces bool
 )
 
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -114,5 +114,7 @@ func init() {
 		fmt.Sprintf(`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
 if specified with --namespace.`))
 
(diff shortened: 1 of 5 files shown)


cmd/print.go
```diff
@@ -23,12 +23,23 @@ import (
...
 )
 
 var (
 	// outputFormat contains currently set output format. Value assigned via --output/-o flag.
 	// Defaults to YAML.
 	outputFormat = "yaml"
+
+	// namespace contains currently set namespace that should be used. Value assigned via
+	// --namespace/-n flag.
+	// Default behavior is to use the current namespace the user is in.
```
LiorLieberman, line 36:
```suggestion
	// The namespace used to query Gateway API objects. Value assigned via
	// --namespace/-n flag.
	// On absence, the current user active namespace is used.
```

cmd/print.go
```diff
@@ -23,12 +23,23 @@ import (
...
+	namespace string
+
+	// allNamespaces indicates whether all namespaces should be used. Value assigned via
+	// --all-namespaces/-A flag.
+	// If present, overrides the namespace variable.
```
LiorLieberman, line 41:
```suggestion
```

LiorLieberman: removing this in favour of the bottom comment about `MarkFlagsMutuallyExclusive`

cmd/print.go
```diff
@@ -59,12 +74,46 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
...
 	printCmd.Flags().StringVarP(&outputFormat, "output", "o", "yaml",
 		fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(allowedFormats, ", ")))
 
+	printCmd.Flags().StringVarP(&namespace, "namespace", "n", "",
+		fmt.Sprintf(`If present, the namespace scope for this CLI request`))
+
+	printCmd.Flags().BoolVarP(&allNamespaces, "all-namespaces", "A", false,
+		fmt.Sprintf(`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
+if specified with --namespace.`))
+
```
LiorLieberman, line 117: instead of documenting and checking in the code whether both flags are provided. 
consider adding [MarkFlagsMutuallyExclusive](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) so cobra will do this for you.

cmd/print.go
```diff
@@ -59,12 +74,46 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
 	}
 }
 
+// getNamespaceFilter returns a namespace filter, taking into consideration whether a specific
+// namespace is requested, or all of them are.
+func getNamespaceFilter(requestedNamespace string, useAllNamespaces bool) (string, error) {
```
LiorLieberman, line 79: this func should be changed if you choose to add MarkFlagsMutuallyExclusive (as per my below comment)

pkg/i2gw/ingress2gateway.go
```diff
@@ -42,6 +49,15 @@ func Run(printer printers.ResourcePrinter) {
 		os.Exit(1)
 	}
 
+	if len(ingressList.Items) == 0 {
+		str := "No resources found"
+		if namespace != "" {
+			str += fmt.Sprintf(" in %s namespace.", namespace)
+		}
+		fmt.Println(str)
+		return
```
LiorLieberman, line 58:
```suggestion
		msg := "No resources found"
		if namespace != "" {
			fmt.Printf("%s in %s namespace", msg, namespace)
		} else {
		        fmt.Println(msg)
		}
		return
```

LiorLieberman (changes requested):
Thanks @levikobi

cmd/print.go
levikobi: That's an excellent suggestion. I wasn't aware that Cobra offers this functionality. Kubectl does allow the usage of both flags, so we just need to decide whether we want to follow their behavior or not.

Personally, I believe your suggestion is cleaner and easier to understand. Either way, let me know your preference, and I'll make the necessary adjustments accordingly.

LiorLieberman: interesting.. what happens when you specify both -A and namespace in kubectl? -A is stronger right?

levikobi: yes

LiorLieberman: yeah so I think [MarkFlagsMutuallyExclusive](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive) makes more sense. I do not see a good reason to allow them both

cmd/print.go
levikobi: I've added MarkFlagsMutuallyExclusive, but I don't see how it should change the functionality at getNamespaceFilter...
We'll still need to figure out what is the required namespace filter based on those two flags, regardless of whether or not they both get assigned simultaneously.

Let me know if you got another idea in your mind.

cmd/print.go
```diff
@@ -59,12 +73,48 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
...
 
 	printCmd.Flags().StringVarP(&outputFormat, "output", "o", "yaml",
 		fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(allowedFormats, ", ")))
 
+	printCmd.Flags().StringVarP(&namespace, "namespace", "n", "",
+		fmt.Sprintf(`If present, the namespace scope for this CLI request`))
+
+	printCmd.Flags().BoolVarP(&allNamespaces, "all-namespaces", "A", false,
+		fmt.Sprintf(`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
+if specified with --namespace.`))
```
LiorLieberman, line 115:
```suggestion
		fmt.Sprintf(`If present, list the requested object(s) across all namespaces. Cannot be set together with "--namespace".`))
```

cmd/print.go
```diff
@@ -59,12 +73,48 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
 	}
 }
 
+// getNamespaceFilter returns a namespace filter, taking into consideration whether a specific
+// namespace is requested, or all of them are.
+func getNamespaceFilter(requestedNamespace string, useAllNamespaces bool) (string, error) {
+
+	// When we should use all namespaces, return an empty string.
+	// This is the first condition since it should override the requestedNamespace,
```
LiorLieberman, line 81:
```suggestion
```

cmd/print.go
```diff
@@ -59,12 +73,48 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
...
+	// if specified.
```
LiorLieberman, line 82:
```suggestion
```

cmd/print.go
LiorLieberman: right, added just small nits to the comments

LiorLieberman:
thanks, added few last nits.

/lgtm

levikobi:
> thanks, added few last nits.
> 
> /lgtm

Thanks @LiorLieberman !

cmd/print_test.go
```diff
@@ -58,3 +75,146 @@ func Test_getResourcePrinter(t *testing.T) {
...
+	}
+	if err = f.Close(); err != nil {
+		os.Remove(kubeConfigFile)
+		return nil, err
+	}
+
+	return func() {
+		os.Remove(kubeConfigFile)
+	}, nil
+}
```
robscott, line 201: This feels strange - I don't think we want our tests to depend on docker desktop. Can we just use kubernetes utils for getting kube config? I'm probably missing some context here.

robscott (commented):
Thanks @levikobi!

cmd/print_test.go
```diff
@@ -58,3 +75,146 @@ func Test_getResourcePrinter(t *testing.T) {
...
+				t.Errorf(`getNamespaceFilter("%s", %v) = %v, expected %v`,
+					tc.namespace, tc.allNamespaces, actualNamespaceFilter, tc.expectedNamespaceFilter)
+			}
+		})
+
+	}
+}
+
+func setupKubeConfig() (func(), error) {
+	const kubeConfigPath = "/tmp/i2gw/.kube"
```
sunjayBhatia, line 145: prefer to use something like https://pkg.go.dev/os#MkdirTemp or https://pkg.go.dev/os#CreateTemp to create temp dirs/files for tests

levikobi: Good idea, thanks

cmd/print_test.go
levikobi: The test doesn't depend on docker desktop (or any other provider). I just wanted to configure a custom kube config file in order to test that getNamespaceInCurrentContext() function can successfully extract the namespace out of it.

I didn't want to assume the user running tests already has a file, so I decided to create a new temporary one with non-default values that I can work with. Doesn't matter if the clusters are from Docker, AKS, EKS, or anything else since the format of kube config stays the same.

robscott: That makes sense, maybe replace the references to "docker-desktop" with something like "example" or "test" so the purpose of this is clear.

levikobi: sure, changed it

robscott:
Thanks @levikobi!

/approve

PR #33, opened 2023-07-05
Fix ingress-nginx canary weight

### LINKED ISSUE #6: ingress-nginx: Canary Weight Total Flawed
As caught by @sunjayBhatia, although we parse the `nginx.ingress.kubernetes.io/canary-weight-total` annotation, we have significant issues with the implementation.

1. We [ignore any error when trying to convert the value of this annotation to an int](https://github.com/kubernetes-sigs/ingress2gateway/blob/a2ba9cdbf470a0b32c8040650193897eef2ab3ef/pkg/i2gw/aggregator.go#L415). This should be bubbled up to the list of errors we return when using the tool. 
2. We don't use `weightTotal` anywhere. Although we populate `weightTotal` with the value of this annotation, we don't actually use it anywhere else in the code. This should be used to calculate the weight assigned to other backends in the same HTTPRoute.
3. We don't have any test coverage for this feature.

These items can either all be fixed in the same PR, or smaller individual PRs depending on what's easiest.

levikobi:
**What type of PR is this?**

/kind bug
/kind cleanup
Resolves what @robscott mentioned at #6.

Fixes #6 

```release-note
Display better-formed error messages to the user.
Fixes ingress-nginx canary weight calculation.
```


sunjayBhatia (commented):
looks like the canary weight code could use some test coverage

LiorLieberman:
@levikobi since #6 covers more than error handling, maybe worth removing the "fixes #6 " line so it wont close the issue when this PR is merged

robscott:
Thanks @levikobi! Can you add some basic test coverage so we can avoid bug(s) like this in the future?

levikobi:
Thanks for commenting @sunjayBhatia @LiorLieberman @robscott 

I am intending to add new tests as well as fix the weight distribution (I've written a note on the main PR comment, but I guess it's easy to miss)

pkg/i2gw/aggregator.go
```diff
@@ -284,34 +295,56 @@ func (rg *ingressRuleGroup) toHTTPRoute() (gatewayv1beta1.HTTPRoute, []error) {
...
+
+	return httpRoute, errors
+}
+
+func (rg *ingressRuleGroup) calculateBackendRefWeight(paths []ingressPath) ([]gatewayv1beta1.HTTPBackendRef, field.ErrorList) {
+	var errors field.ErrorList
+	var backendRefs []gatewayv1beta1.HTTPBackendRef
+
+	var numWeightedBackends, totalWeightSet int32
+	var weightTotal = 100
```
LiorLieberman, line 313: Worth adding a comment that this is related to ingress nginx and it is the default for `nginx.ingress.kubernetes.io/canary-weight-total` based on https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#canary

pkg/i2gw/aggregator.go
```diff
@@ -284,34 +295,56 @@ func (rg *ingressRuleGroup) toHTTPRoute() (gatewayv1beta1.HTTPRoute, []error) {
...
+		weightToSet := (int32(weightTotal) - totalWeightSet) / (int32(len(backendRefs)) - numWeightedBackends)
+		if weightToSet < 0 {
+			weightToSet = 0
+		}
+		for i := range backendRefs {
+			if backendRefs[i].Weight == nil {
+				backendRefs[i].Weight = &weightToSet
+			}
+			if *backendRefs[i].Weight > int32(weightTotal) {
+				backendRefs[i].Weight = pointer.Int32(int32(weightTotal))
```
LiorLieberman, line 341: not sure I understand the logic.

If the weight set is higher than weightTotal you set it to `weightTotal` ?

LiorLieberman (commented):
Thanks @levikobi, this is very well written !

@robscott @sunjayBhatia Should we remove support for nginx annotations explicitly in the functions and have some separate logic to handle provider specific annotations instead?

pkg/i2gw/aggregator.go
levikobi: Yes. This is performed in order to prevent the following example (which could happen until now):
You set an ingress directing to a canary deployment with a weight of 130 and a default weight-total of 100. The primary ingress weight will be -30.

pkg/i2gw/aggregator.go
levikobi: Good idea, I'll add the comment.
Anyway, I think we should remove the support for nginx annotations from here and handle it separately with the other provider-specific logic, as you suggested.

robscott:
>@robscott @sunjayBhatia Should we remove support for nginx annotations explicitly in the functions and have some separate logic to handle provider specific annotations instead?

Agree that would be a good follow up, just some effort to make our code more extensible so it's easier to plug in additional annotations in the future. Don't think that needs to block this though.

Thanks @levikobi!

/approve

Will defer to @LiorLieberman or @sunjayBhatia for LGTM.

sunjayBhatia (approved):

sunjayBhatia:
as far as I understand the nginx canary annotation, LGTM 👍🏽

pkg/i2gw/aggregator.go
LiorLieberman: Maybe I misunderstand something, but what happens if you have two ingress, one with `120` and the other one with `120` ? (`weightTotal` is not set - means it is 100 by default)

LiorLieberman: Answering myself - they will both get a 100 weight so it will be 50:50 in the traffic.

Now, what happens when one is `150` and the other one is `100`? again no `weightTotal` is set

LiorLieberman: have a link?

levikobi: I haven't found a statement saying they do not support more than one, but this is what I did:

I followed https://kubernetes.github.io/ingress-nginx/examples/canary/
The traffic split when as expected.

Then I added an additional canary ingress.
Even when setting its weight to 100/100 and the rest of the ingresses to 0/100, the new one would not receive requests. The controller itself didn't have any errors though. It was weird.

tao12345666333: The traffic cannot be segmented as expected due to the logic here.  https://github.com/kubernetes/ingress-nginx/blob/5b35651a1a40f5506d06540c1fcd103dc41a16d5/rootfs/etc/nginx/lua/balancer.lua#L258-L266

LiorLieberman: Well, I am a bit confused by the logic but it looks like like you are right @levikobi . However IIUC it is one canary ingress per ingress rule not per deployment (check the [Known Limitations section](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary))

How the code handles a situation of more than one ingress for the same host with Canary annotations? I think we don't account for this case however I do not think it is very important now as we should deprecate this logic soon and move to a more generic way for provider-specific annotations.

LiorLieberman:
/unhold
/lgtm

Thanks @levikobi

pkg/i2gw/aggregator.go
levikobi: > How the code handles a situation of more than one ingress for the same host with Canary annotations?

I don't think there can be a situation of more than one ingress for the same host with Canary annotations with the ingress-nginx implementation (I tried it out and it just didn't work)... But I may be wrong

> I do not think it is very important now as we should deprecate this logic soon and move to a more generic way for provider-specific annotations.

Agree

PR #35, opened 2023-07-07
Add ability to read yaml config file and convert ingresses to gateways.

Treenhan:
The flag for this is -input_file. Also added tests.

/kind feature
/kind test

Enables the tool to take a file path as input to yaml file. Then convert Ingress resources in the file to Gateways.

```release-note
NONE
```


Treenhan:
@LiorLieberman PTAL

LiorLieberman:
[we discussed offline] It is waiting for a small refactor to use `"k8s.io/apimachinery/pkg/util/yaml"`

Treenhan:
> refactor to use `"k8s.io/apimachinery/pkg`

This is now implemented

pkg/i2gw/ingress2gateway.go
```diff
@@ -70,6 +83,73 @@ func Run(printer printers.ResourcePrinter, namespace string) {
...
+		}
+		if u == nil {
+			continue
+		}
+		objs = append(objs, u)
+	}
+	return objs, nil
+}
+
+func readFile(l *networkingv1.IngressList, inputFile string) error {
```
LiorLieberman, line 105: This function is doing much more than reading a file. Consider changing this to something like `constructIngressesFromFile`. 

Plus a comment on what it does would be good

pkg/i2gw/tests/input-file.json
```diff
@@ -0,0 +1,121 @@
...
+                                    }
+                                }
+                            ]
+                        }
+                    }
+                ]
+            }
+        }
+    ]
+}
```
LiorLieberman, line 121: nit: newline

LiorLieberman (commented):
Thanks @Treenhan !

Posted a few comments. 

Also, should we support the namespace filtering when reading from a file as well? (ref: https://github.com/kubernetes-sigs/ingress2gateway/pull/31)
@levikobi @Treenhan thoughts?

LiorLieberman:
Please add the release notes section (the section in grey) Plus Fixes: #35 to the PR description

levikobi:
> Thanks @Treenhan !
> 
> Posted a few comments.
> 
> Also, should we support the namespace filtering when reading from a file as well? (ref: #31) @levikobi @Treenhan thoughts?

Given the fact that we can only accept one file, I don't think namespace filtering is necessary.

I think a good addition will be to accept multiple files/folder (same as kubectl create -f).
And maybe even add a --recursive flag.

Those don't have to be implemented now though.

levikobi:
> We can have different ingresses with different namespaces on the same file.
That's a good point. Yea, then I guess we should support it. Should be fairly straightforward to implement

> What would be the purpose of this flag ?
This is my intention:
Given: i2gw print -f . -R
I expect to convert everything from the current working directory including subdirectories, recursively.

Treenhan:
> That's a good point. Yea, then I guess we should support it. Should be fairly straightforward to implement
I created additional issue which I will implement in a next few days. https://github.com/kubernetes-sigs/ingress2gateway/issues/37

> I expect to convert everything from the current working directory including subdirectories, recursively.
I think this is a good idea, should I create an issue for this? @LiorLieberman 

FYI @LiorLieberman all comments addressed!

Treenhan:
This is closed in favor of https://github.com/kubernetes-sigs/ingress2gateway/pull/38

Issue #36, opened 2023-07-12
Support translating ingresses from files


LiorLieberman:
**What would you like to be added**:
We only support reading ingresses from the cluster.
We want to support converting ingresses from files as well.

**Why this is needed**:
This can be useful as a lot of configuration files are sitting in files and not necessarily applied in the cluster.

Treenhan:
Hi @LiorLieberman can you assign me to this?

PR #38, opened 2023-07-14
Add ability to read yaml/json config file and convert ingresses to gateways

Treenhan:
The flag for this is -input_file. Also added tests.

/kind feature

What this PR does / why we need it:
Enables the tool to take a file path as input to yaml/json file. Then convert Ingress resources in the file to Gateways.

release-note
Fixes: https://github.com/kubernetes-sigs/ingress2gateway/issues/36


pkg/i2gw/ingress2gateway_test.go
```diff
@@ -0,0 +1,121 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
LiorLieberman, line 2: can you change to 2023 ?

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	networkingv1 "k8s.io/api/networking/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_inputFile(t *testing.T) {
```
LiorLieberman, line 27: The unitests are to check functions.  so ideally we would want something like `Test_<func_name>`

So either you split it into multiple functions unitests or you leave it as one but you change it to `Test_constructIngressesFromFile` (which looks like the function your are testing)

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -0,0 +1,121 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			ingressList := &networkingv1.IngressList{}
+			err := constructIngressesFromFile(ingressList, tc.filePath)
+			if err != nil {
+				t.Errorf("Failed to open test file: %v", err)
+			}
+			for i, got := range ingressList.Items {
+				want := expectIngresses[i]
+				if !apiequality.Semantic.DeepEqual(got, want) {
+					t.Errorf("Expected Ingress %d to be %+v\n Got: %+v\n Diff: %s", i, want, got, cmp.Diff(want, got))
```
LiorLieberman, line 116: maybe we can split this logic to another helper function?

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -0,0 +1,121 @@
...
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_inputFile(t *testing.T) {
+	iPrefix := networkingv1.PathTypePrefix
+	ingress1ClassName := "ingress1-example"
+	ingress2ClassName := "ingress2-example"
+
+	expectIngresses := []networkingv1.Ingress{
```
LiorLieberman, line 32: I am not against having this list but I think it would be better to have it as a map where the name is describing the ingress you expect like: `ingress_with_test1_backend_exposed_in_443` (you could probably think of a better name that describes what you are expecting) and the value is the ingress.

LiorLieberman: Alternatively, instead of looping over them when you compare, have a function that checks whether the `expectedIngresses` are the same as what you got in the `ingressList`

pkg/i2gw/ingress2gateway_test.go
LiorLieberman: I just did it nvm

pkg/i2gw/ingress2gateway_test.go
Treenhan: I made it `Test_constructIngressesFromFile`

pkg/i2gw/ingress2gateway_test.go
Treenhan: I made a helper method that does the internal loop

pkg/i2gw/ingress2gateway_test.go
Treenhan: yes I did this

LiorLieberman:
Thanks @Treenhan 
/lgtm

robscott:
Thanks @Treenhan!

/approve

Issue #37, opened 2023-07-13
Support namespace filtering when reading from files


Treenhan:
The idea is when running ingress2gateway command, we should be able to specify which namespace we want to use. If a namespace is explicitly specified, the command will only process the Ingress resources within that namespace. However, if no namespace is provided, the command will process Ingress resources across all namespaces.

/kind feature
/kind test

levikobi:
Hi @Treenhan,
I think we should keep the same functionality for reading from a file as reading from the cluster.
At the moment, when reading from the cluster without specifying a namespace you will only process objects from the current active namespace, as opposed to all the namespaces. 
If you want to read from all namesapces, you have a --all-namespaces flag.

LiorLieberman:
@Treenhan whats the status of it? do you have any blockers you need help with?

Treenhan:
I'm still deciding what the default behavior should be if no namespace is specified in the YAML/JSON files. I assume that not all machines converting these files will have access to cluster information, or they don't have a cluster at all but just want to convert the files. So they won't be able to grab the default namespace from the cluster context.

My thinking is that the default behavior should be to read everything, unless users want to choose a specific namespace. The reason is that without cluster context, there's no way to know what the default namespace should be.

@LiorLieberman @levikobi any ideas?

Treenhan:
I think I will go with this:
- Without any flag, use the namespace in the active context, if there is no context, it will convert all (this allows users to use this tool without having kubernetes running)
- Use --all-namespaces to convert all
- Use  --namespace X to filter out specific namespaces

levikobi:
> I think I will go with this:
> - Without any flag, use the namespace in the active context, if there is no context, it will convert all (this allows users to use this tool without having kubernetes running)
> - Use --all-namespaces to convert all
> - Use  --namespace X to filter out specific namespaces

Sounds good to me @Treenhan !

LiorLieberman:
+1 sgtm

Treenhan:
This is now done in https://github.com/kubernetes-sigs/ingress2gateway/pull/43

PR #39, opened 2023-07-22
Support namespace filtering for reading resources from local file

Treenhan:
**What type of PR is this?**
/kind feature

Support namespace filtering for reading resources from local file

Fixes https://github.com/kubernetes-sigs/ingress2gateway/issues/37

```release-note
NONE
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -65,7 +65,6 @@ var printCmd = &cobra.Command{
 		// If the input file is specified, and we failed to get the namespace, use all namespaces.
 		i2gw.Run(resourcePrinter, namespaceFilter, inputFile)
 		return nil
-		return err
 	},
 }
 
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -56,14 +56,15 @@ var printCmd = &cobra.Command{
 			return err
 		}
 		namespaceFilter, err := getNamespaceFilter(namespace, allNamespaces)
-		if err == nil {
-			i2gw.Run(resourcePrinter, namespaceFilter, inputFile)
-			return nil
-		} else if clientcmd.IsConfigurationInvalid(err) && inputFile != "" {
-			// process all namespaces if context does not exist while reading from file
-			i2gw.Run(resourcePrinter, "", inputFile)
-			return nil
+		if err != nil && inputFile == "" {
+			// When asked to read from the cluster, but getting the current namespace
+			// failed for whatever reason - do not process the request.
+			return err
 		}
+		// If err is nil we got the right filtered namespace.
+		// If the input file is specified, and we failed to get the namespace, use all namespaces.
+		i2gw.Run(resourcePrinter, namespaceFilter, inputFile)
+		return nil
 		return err
 	},
 }
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -55,7 +55,11 @@ var printCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		namespaceFilter, err := getNamespaceFilter(namespace, allNamespaces)
+		inputFileMode := false
+		if inputFile != "" {
+			inputFileMode = true
+		}
+		namespaceFilter, err := getNamespaceFilter(namespace, allNamespaces, inputFileMode)
 		if err != nil {
 			return err
 		}
@@ -79,7 +83,7 @@ func getResourcePrinter(outputFormat string) (printers.ResourcePrinter, error) {
 
 // getNamespaceFilter returns a namespace filter, taking into consideration whether a specific
 // namespace is requested, or all of them are.
-func getNamespaceFilter(requestedNamespace string, useAllNamespaces bool) (string, error) {
+func getNamespaceFilter(requestedNamespace string, useAllNamespaces bool, inputFileMode bool) (string, error) {
 
 	// When we should use all namespaces, return an empty string.
 	// This is the first condition since it should override the requestedNamespace,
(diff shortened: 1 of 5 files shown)


cmd/print.go
```diff
@@ -55,7 +55,11 @@ var printCmd = &cobra.Command{
...
 		}
-		namespaceFilter, err := getNamespaceFilter(namespace, allNamespaces)
+		inputFileMode := false
+		if inputFile != "" {
+			inputFileMode = true
+		}
+		namespaceFilter, err := getNamespaceFilter(namespace, allNamespaces, inputFileMode)
 		if err != nil {
 			return err
 		}
```
levikobi, line 65: I think it would be cleaner if you returned a predefined error at getNamespaceFilter (something like ErrNoContext), and then checked if that specific error was returned.
If such an error occurs, and we are reading from files return an empty string, otherwise return an error to the user.

Just my opinion.

LiorLieberman: +1

cmd/print.go
```diff
@@ -89,18 +93,23 @@ func getNamespaceFilter(requestedNamespace string, useAllNamespaces bool) (strin
 	}
 
 	if requestedNamespace == "" {
-		return getNamespaceInCurrentContext()
+		getNamespaceInCurrentContext(inputFileMode)
```
LiorLieberman, line 96: seems like the `return` was accidentally dropped.

LiorLieberman (commented):
I less like the namespace moving around between functions. 

I think we got into the stage where we should have a `printRunner` object that holds the data we need; for example all the flags we hold in vars etc.

And break the `i2gw.Run()`. something like:

cmd/print.go
```go
type printRunner struct {
	namespace string
	outputFormat string
	...
}

func (pr *printRunner) PrintGatewaysAndHttpRoutes(cmd *cobra.Command, args []string) error  {
  i2gw.aggregate (something like, we should break i2gw package to smaller things other than Run() and use it here)
  ...
}

func newPrintCommand() *cobra.Command  {
  pr :=  &printRunner{}
  cmd := &cobra.Command{
	Use:   "print .....",
	Short: "print http routes and gateway converted from ingresses...",
	RunE:  pr.PrintGatewaysAndHttpRoutes,
  }
  cmd.Flags().StringVarP(&outputFormat, "output", "o", "yaml",
		fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(allowedFormats, ", ")))
  ... (other flags)
  return cmd		
}

```

and change `cmd/root.go` accordingly like adding `rootCmd.AddCommand(newPrintCommand)` 

Long comment but I hope that makes sense.

Treenhan:
> I less like the namespace moving around between functions.
> 
> I think we got into the stage where we should have a `printRunner` object that holds the data we need; for example all the flags we hold in vars etc.
> 
> And break the `i2gw.Run()`. something like:
> 
> cmd/print.go
> 
> ```go
> type printRunner struct {
> 	namespace string
> 	outputFormat string
> 	...
> }
> 
> func (pr *printRunner) PrintGatewaysAndHttpRoutes(cmd *cobra.Command, args []string) error  {
>   i2gw.aggregate (something like, we should break i2gw package to smaller things other than Run() and use it here)
>   ...
> }
> 
> func newPrintCommand() *cobra.Command  {
>   pr :=  &printRunner{}
>   cmd := &cobra.Command{
> 	Use:   "print .....",
> 	Short: "print http routes and gateway converted from ingresses...",
> 	RunE:  pr.PrintGatewaysAndHttpRoutes,
>   }
>   cmd.Flags().StringVarP(&outputFormat, "output", "o", "yaml",
> 		fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(allowedFormats, ", ")))
>   ... (other flags)
>   return cmd		
> }
> ```
> 
> and change `cmd/root.go` accordingly like adding `rootCmd.AddCommand(newPrintCommand)`
> 
> Long comment but I hope that makes sense.

As discussed offline, I will implement this as a separate issue and PR as this one is focused on implementing the namespace filtering for files

cmd/print.go
```diff
@@ -56,11 +56,15 @@ var printCmd = &cobra.Command{
...
-		if err != nil {
-			return err
+		if err == nil {
+			i2gw.Run(resourcePrinter, namespaceFilter, inputFile)
+			return nil
+		} else if clientcmd.IsConfigurationInvalid(err) && inputFile != "" {
+			// process all namespaces if context does not exist while reading from file
+			i2gw.Run(resourcePrinter, "", inputFile)
+			return nil
 		}
```
levikobi, line 66:
```suggestion
		if err != nil && inputFile == "" {
			// When asked to read from the cluster, but getting the current namespace
			// failed for whatever reason - do not process the request.
			return err
		}
		// If err is nil we got the right filtered namespace.
		// If the input file is specified, and we failed to get the namespace, use all namespaces.
		i2gw.Run(resourcePrinter, namespaceFilter, inputFile)
		return nil
```
I think this could be easier to understand

cmd/print.go
```diff
@@ -56,11 +56,15 @@ var printCmd = &cobra.Command{
...
-		i2gw.Run(resourcePrinter, namespaceFilter, inputFile)
-		return nil
+		return err
```
levikobi, line 67:
```suggestion
```
Comes with my other suggestion

levikobi (commented):
Looks better in my opinion, still a minor change that I think could improve it

cmd/print.go
Treenhan: It is indeed, thanks @levikobi

PR #40, opened 2023-07-24
Refactor project layout to enable multiple providers

### LINKED ISSUE #28: Refactor the code, to be more extensible and testable
Refactor the code, to be more extensible and testable.
**Why this is needed**:

levikobi:
**What type of PR is this?**

/kind cleanup
/kind feature

Adding support for converting multiple Ingress providers, in an extensible and testable way.

Fixes #28 

```release-note
None
```


LiorLieberman (commented):
Thanks @levikobi, thats an important piece of work!

The approach is LGTM but left some comments

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,135 @@ import (
...
+// Provider implementation.
+type ProviderConf struct{}
+
+// The Provider interface specifies the required functionality which needs to be
+// implemented by every concrete Ingress/Gateway-API provider, in order for it to
+// be used.
+type Provider interface {
+	CustomResourceReader
+	ResourceConverter
+}
```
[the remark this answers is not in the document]
levikobi: I thought about adding a providers.go file under providers package, but that will lead to a circular dependency problem.
I am open to separating those declarations as you suggested. If you prefer it, let me know.

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,135 @@ import (
...
+	}
+
+	for name, provider := range providerByName {
+		var customResources interface{}
+
+		if err = provider.ReadResourcesFromCluster(context.Background(), &customResources); err != nil {
+			return fmt.Errorf("failed to read %s resources from the cluster: %w", name, err)
+		}
+
+		// TODO: Open a new issue - the ingress-nginx provider contains conversion logic which
```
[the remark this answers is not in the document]
levikobi: OK. I'll add the common logic in this PR.

pkg/i2gw/ingress2gateway.go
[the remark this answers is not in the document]
LiorLieberman: feels like a better fit but we could ask for more opinions. 

/cc @robscott wdyt?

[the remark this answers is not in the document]
robscott: Yeah I feel like these would be best with some kind of separation, even if that's just a separate file.

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,135 @@ import (
...
+	ReadResourcesFromFiles(ctx context.Context, customResources interface{}, filename string) error
+}
+
+// The ResourceConverter interface specifies all the implemented Gateway API resource
+// conversion functions.
+type ResourceConverter interface {
+
+	// ConvertHTTPRoutes converts the received ingresses and custom resources
+	// associated with the Provider into HTTPRoutes and Gateways.
+	ConvertHTTPRoutes(ingresses []networkingv1.Ingress, customResources interface{}) ([]gatewayv1beta1.HTTPRoute, []gatewayv1beta1.Gateway, field.ErrorList)
```
sunjayBhatia, line 79: should this be more generic (and maybe return a struct containing lists of xRoutes and Gateways) to support adding more Route types in the future (e.g. Contour can convert some HTTPProxies to TCP/TLSRoutes)

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,135 @@ import (
...
+type Provider interface {
+	CustomResourceReader
+	ResourceConverter
+}
+
+type CustomResourceReader interface {
+
+	// ReadResourcesFromCluster reads custom resources associated with
+	// the underlying Provider implementation from the kubernetes cluster.
+	ReadResourcesFromCluster(ctx context.Context, customResources interface{}) error
```
sunjayBhatia, line 66: would be nice to provide a param here or some other mechanism to enable providers to initialize a client to fetch custom resources with (so common boilerplate doesn't have to be repeated)

could make it part of provider init to add to a scheme a common client is created with, or otherwise

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,135 @@ import (
...
+	var gateways []gatewayv1beta1.Gateway
+	var httpRoutes []gatewayv1beta1.HTTPRoute
+	var errs field.ErrorList
+
+	providerByName := constructProviders(&ProviderConf{})
+	if len(providerByName) == 0 {
+		return fmt.Errorf("no providers")
+	}
+
+	for name, provider := range providerByName {
```
sunjayBhatia, line 124: makes sense for now, but as we add providers I would guess we want to add a flag to give users the opportunity to select a specific one? another TODO/issue for that?

pkg/i2gw/ingress2gateway.go
levikobi: I totally agree with you @sunjayBhatia, I thought of it too after the fact.
I will make the adjustment

pkg/i2gw/ingress2gateway.go
levikobi: There is a TODO for this one ;)
Take a look at https://github.com/kubernetes-sigs/ingress2gateway/blob/781a9d9a8858ef0d81f17d10839698fffce04e44/pkg/i2gw/ingress2gateway.go#L158

pkg/i2gw/ingress2gateway.go
levikobi: The idea was to pass the client using ProviderConf, but it may be better to pass it directly as an argument to the function. Not sure yet about this one.

levikobi:
> Thanks @levikobi, thats an important piece of work!
> 
> The approach is LGTM but left some comments.
> 
> Also, any specific reason you wrapped all the errors with `%w` ? I did not find error types in this PR so not sure whats the value of the wrapping

As discussed offline, the %w doesn't add value right now. I will revert to %v.

pkg/i2gw/ingress2gateway.go
levikobi: done

pkg/i2gw/ingress2gateway.go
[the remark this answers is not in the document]
levikobi: done

LiorLieberman:
Havent had time to look after the new commits yet but since we gonna have contour and istio implementing the provider specific logic first I would like another review from @howardjohn and @sunjayBhatia 

/cc @howardjohn @sunjayBhatia

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,92 @@ import (
...
+
+	providerByName := constructProviders(&ProviderConf{})
+	if len(providerByName) == 0 {
+		return fmt.Errorf("no providers")
+	}
+
+	resources := IngressResources{Ingresses: ingresses.Items}
+
+	for name, provider := range providerByName {
+		if err = provider.ReadResourcesFromCluster(context.Background(), &resources.CustomResources); err != nil {
```
LiorLieberman, line 83: maybe init context outside and pass the same one here?

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,92 @@ import (
...
+			return fmt.Errorf("failed to read %s resources from the cluster: %w", name, err)
+		}
+
+		gatewayResources, conversionErrs := provider.IngressToGateway(resources)
```
LiorLieberman, line 87: s/IngressToGateway/ToGateway

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,92 @@ import (
...
+		errs = append(errs, conversionErrs...)
+		for _, gateway := range gatewayResources.Gateways {
+			gateways = append(gateways, gateway)
+		}
+		for _, route := range gatewayResources.HTTPRoutes {
+			httpRoutes = append(httpRoutes, route)
+		}
+	}
+
+	// TODO: Open a new issue - collate gateways objects.
```
LiorLieberman, line 97: please open an issue for it and then put the issue num in the parenthesis. 

TODO(<issue_number>) collate gateways objects

pkg/i2gw/ingress2gateway.go
```diff
@@ -35,70 +35,92 @@ import (
...
+
+	return nil
 }
 
-func ingresses2GatewaysAndHTTPRoutes(ingresses []networkingv1.Ingress) ([]gatewayv1beta1.HTTPRoute, []gatewayv1beta1.Gateway, field.ErrorList) {
-	aggregator := ingressAggregator{ruleGroups: map[ruleGroupKey]*ingressRuleGroup{}}
+// constructProviders constructs a map of concrete Provider implementations
+// by their ProviderName.
+//
+// TODO: Open a new issue - let users filter by provider name.
```
LiorLieberman, line 115: same comment with the issue number

pkg/i2gw/provider.go
```diff
@@ -0,0 +1,107 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
LiorLieberman, line 2: 2023, ditto for all **new** files you created on the PR

pkg/i2gw/provider.go
```diff
@@ -0,0 +1,107 @@
...
+func HTTPRouteToHTTPRouteKey(r gatewayv1beta1.HTTPRoute) HTTPRouteKey {
+	return HTTPRouteKey(r.Namespace + ":" + r.Name)
+}
+
+// FeatureParser is a function that reads the IngressResources, and applies
+// the appropriate modifications to the GatewayResources.
+//
+// Different FeatureParsers will run in undetermined order. The function must
+// modify / create only the required fields of the gateway resources and nothing else.
+type FeatureParser func(IngressResources, *GatewayResources) field.ErrorList
```
LiorLieberman, line 107: Maybe we should think of a different name to `IngressResources` its a bit confusing. 
Essentially since it holds both ingresses and custom resources it is more of "resources we are going to transform"

Maybe "ToGatewayResources" ? Not sure I like that either

pkg/i2gw/providers/ingressnginx/converter.go
```diff
@@ -0,0 +1,59 @@
...
+	// Convert plain ingress resources to gateway resources, ignoring all
+	// provider-specific features.
+	gatewayResources, errs := common.IngressToGateway(resources.Ingresses)
+	if len(errs) > 0 {
+		return i2gw.GatewayResources{}, errs
+	}
+
+	// Apply all patch functions on the gateway resources, one by one.
+	for _, parseFeature := range c.featureParsers {
+		errs = append(errs, parseFeature(resources, &gatewayResources)...)
```
LiorLieberman, line 55: worth splitting this to two lines and adding a comment

LiorLieberman: also maybe `featureFunc` instead of `parseFeature`

pkg/i2gw/ingress2gateway.go
levikobi: done

pkg/i2gw/provider.go
```diff
@@ -0,0 +1,107 @@
...
+	ReadResourcesFromFiles(ctx context.Context, customResources interface{}, filename string) error
+}
+
+// The ResourceConverter interface specifies all the implemented Gateway API resource
+// conversion functions.
+type ResourceConverter interface {
+
+	// ToGateway converts the received IngressResources associated
+	// with the Provider into GatewayResources.
+	ToGateway(resources IngressResources) (GatewayResources, field.ErrorList)
```
Xunzhuo, line 68: `ToGateway` is a bit not clear, `Gateway` is also a resource in GWAPI.

pkg/i2gw/ingress2gateway.go
youngnick: I agree with this one, but I recommend also putting the person who logged the TODO right there if we can.

sunjayBhatia (commented):
besides resolving https://github.com/kubernetes-sigs/ingress2gateway/pull/40#discussion_r1274104021 I think this is looking good, haven't had a change to work on the Contour provider yet but in doing so I'm sure we'll iron out some more bits 👍🏽

cmd/root.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
LiorLieberman, line 2: no need to convert old files, only \*\***new**\*\* files you created need the 2023 since it is the current year.

pkg/i2gw/ingress2gateway.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
LiorLieberman, line 2: same as above comment

cmd/root.go
levikobi: Got it, should I leave it like that or revert back to the original year?

pkg/i2gw/provider.go
levikobi: Would `ToGatewayResources` be better @Xunzhuo?
@LiorLieberman you asked to rename this function too, any thoughts?
Would love it if we could all come to an agreement on a name here

pkg/i2gw/ingress2gateway.go
levikobi: I gave it a second thought, and I'm actually not sure we will need to collate the different resources that providers will generate.
Say we have 2 providers that each returned one `gateway`, they should not be merged into a single `gateway`.

So basically I think we can remove this todo altogether without opening a new issue, but please let me know if you disagree.

pkg/i2gw/provider.go
levikobi: The best next thing I can think about is `IngressRelatedResources` and `GatewayRelatedResources`, but in my opinion, they are unnecessarily long.
I don't really see a problem with the current naming convention, but if you or others have a better suggestion that you prefer, please let me know

levikobi:
> besides resolving [#40 (comment)](https://github.com/kubernetes-sigs/ingress2gateway/pull/40#discussion_r1274104021) I think this is looking good, haven't had a change to work on the Contour provider yet but in doing so I'm sure we'll iron out some more bits 👍🏽

Added the k8s client to the `ProviderConf` struct so that each provider will have access to an initialized client.
If anything else is needed by the providers it can be added in the same way.

pkg/i2gw/providers/ingressnginx/converter.go
levikobi: Sure

pkg/i2gw/provider.go
```diff
@@ -0,0 +1,110 @@
...
+// custom resources.
+type IngressResources struct {
+	Ingresses       []networkingv1.Ingress
+	CustomResources interface{}
+}
+
+// GatewayResources contains all Gateway-API objects.
+type GatewayResources struct {
+	Gateways   map[GatewayKey]gatewayv1beta1.Gateway
+	HTTPRoutes map[HTTPRouteKey]gatewayv1beta1.HTTPRoute
```
howardjohn, line 84: would be nice to allow other types as well

howardjohn: all the types in the API I mean

pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (r *resourceReader) ReadResourcesFromCluster(ctx context.Context, customResources interface{}) error {
+	// ingress-nginx does not have any CRDs.
+	return nil
+}
+
+func (r *resourceReader) ReadResourcesFromFiles(ctx context.Context, customResources interface{}, filename string) error {
+	// ingress-nginx does not have any CRDs.
+	return nil
+}
```
robscott, line 45: It may be easier to save this part of the interface for when we're actually adding support for an implementation that has additional custom resources. Defining it without a clear use case/implementation means that we're probably going to miss some things anyway.

pkg/i2gw/providers/ingressnginx/converter_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+func apiKindPtr(s string) *gatewayv1beta1.Kind {
+	k := gatewayv1beta1.Kind(s)
+	return &k
+}
+
+func gatewayHostnamePtr(s string) *gatewayv1beta1.Hostname {
+	h := gatewayv1beta1.Hostname(s)
+	return &h
+}
```
robscott, line 241: Same comment as above - generic would make this much simpler

robscott (commented):
Thanks @levikobi! Didn't have time to get too deep into this, but it's really great to see this getting split out. Will defer to @LiorLieberman for final approval, but other than the comments here, this seems like a great step forward.

pkg/i2gw/providers/common/converter_test.go
```diff
@@ -0,0 +1,382 @@
...
+
+func portNumberPtr(p int) *gatewayv1beta1.PortNumber {
+	pn := gatewayv1beta1.PortNumber(p)
+	return &pn
+}
+
+func gatewayHostnamePtr(s string) *gatewayv1beta1.Hostname {
+	h := gatewayv1beta1.Hostname(s)
+	return &h
+}
```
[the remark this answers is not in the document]
levikobi: sure, done

pkg/i2gw/providers/common/converter.go
```diff
@@ -14,31 +14,63 @@ See the License for the specific language governing permissions and
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	"k8s.io/utils/pointer"
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
+// ToGateway converts the received ingresses to i2gw.GatewayResources,
+// without taking into consideration any provider specific logic.
+func ToGateway(ingresses []networkingv1.Ingress) (i2gw.GatewayResources, field.ErrorList) {
```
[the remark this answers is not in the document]
levikobi: I went with `ToGatewayResources`, just to kind of make it on the same page with `InputResources`

pkg/i2gw/provider.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// GatewayKey is a unique identifier for a gateway object.
+// Constructed by namespace:name.
+type GatewayKey string
+
+// GatewayToGatewayKey assembles the GatewayKey out of a Gateway.
+func GatewayToGatewayKey(g gatewayv1beta1.Gateway) GatewayKey {
+	return GatewayKey(g.Namespace + ":" + g.Name)
+}
```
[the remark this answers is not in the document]
levikobi: Great suggestion, this is a better approach. Done.

pkg/i2gw/provider.go
levikobi: Sure mate, I added the other resources

pkg/i2gw/provider.go
levikobi: Made the switch to `InputResources` I think it's descriptive enough.
I mark it as resolved but please reopen this discussion if you want to further debate on it.

pkg/i2gw/providers/ingressnginx/resource_reader.go
levikobi: I tend to agree with @LiorLieberman on this one. I do think it will be more convenient to have it as a reference point for further implementations, but if you have strong feelings about it I will remove it @robscott

LiorLieberman (commented):
Thanks very much for you hard work @levikobi !

/lgtm
/approve

Issue #41, opened 2023-07-24
Wrap all flags into a struct so that we don't have to update our methods as more flags are introduced


Treenhan:
The idea is putting all of our flags in an object (something like printRunner) and pass it around instead of adding a new parameter everytime a new flag is introduced.

@LiorLieberman has kindly drafted this idea in the following snippet

cmd/print.go
```
type printRunner struct {
	namespace string
	outputFormat string
	...
}

func (pr *printRunner) PrintGatewaysAndHttpRoutes(cmd *cobra.Command, args []string) error  {
  i2gw.aggregate (something like, we should break i2gw package to smaller things other than Run() and use it here)
  ...
}

func newPrintCommand() *cobra.Command  {
  pr :=  &printRunner{}
  cmd := &cobra.Command{
	Use:   "print .....",
	Short: "print http routes and gateway converted from ingresses...",
	RunE:  pr.PrintGatewaysAndHttpRoutes,
  }
  cmd.Flags().StringVarP(&outputFormat, "output", "o", "yaml",
		fmt.Sprintf(`Output format. One of: (%s)`, strings.Join(allowedFormats, ", ")))
  ... (other flags)
  return cmd		
}
```

/kind refactor

Issue #24, opened 2023-06-17
Add support for different resource mode: local and remote, to support translating Ingress from files besides from the clusters.


Treenhan:
**What would you like to be added:**
Add support for different resource mode: local and remote, to support translating Ingress from files besides from the clusters.

Treenhan:
Can I get assigned @LiorLieberman ?

LiorLieberman:
@Treenhan are you planning to do any additional work on this issue?

mlavacca:
As far as I get from this issue's intention, this feature is already implemented. Do you agree we can close it?

@LiorLieberman 
@levikobi

levikobi:
@mlavacca Yes so it looks

LiorLieberman:
we do not have "remote" support. i.e reading from github(without cloning) . but I dont think anyone ever asked for it so we can close for now

PR #43, opened 2023-07-26
Refactor print.go and ingress2gateway.go

Treenhan:
/kind cleanup

What this PR does / why we need it:
This breaks the Run() method in ingress2gateway.go down to utility methods and use them in print.go. This helps extending the functionalities in the future as we may introduce apply instead of just print for example.

When reviewing this, please take a look only at the second commit, as this is built on top of https://github.com/kubernetes-sigs/ingress2gateway/pull/39

Fixes https://github.com/kubernetes-sigs/ingress2gateway/issues/41

```
NONE
```


cmd/print.go
```diff
@@ -43,55 +47,147 @@ var (
...
-	RunE: func(cmd *cobra.Command, args []string) error {
-		resourcePrinter, err := getResourcePrinter(outputFormat)
+	// resourcePrinter determines how resource objects are printed out
+	resourcePrinter printers.ResourcePrinter
+
+	// Only resources that matches this filter will be processed.
+	namespaceFilter string
+}
+
+func (pr *PrintRunner) PrintGatewaysAndHttpRoutes(cmd *cobra.Command, args []string) error {
```
LiorLieberman, line 58: worth adding a short function comment?

cmd/print.go
```diff
@@ -43,55 +47,147 @@ var (
...
+	cmd.Flags().StringVarP(&pr.namespace, "namespace", "n", "",
+		`If present, the namespace scope for this CLI request`)
+
+	cmd.Flags().BoolVarP(&pr.allNamespaces, "all-namespaces", "A", false,
+		`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
+if specified with --namespace.`)
+
+	cmd.MarkFlagsMutuallyExclusive("namespace", "all-namespaces")
+	pr.InitializeResourcePrinter()
+	pr.InitializeNamespaceFilter()
```
LiorLieberman, line 189: err handling

pkg/i2gw/ingress2gateway.go
```diff
@@ -31,63 +31,18 @@ import (
 	kubeyaml "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/cli-runtime/pkg/printers"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
-func Run(printer printers.ResourcePrinter, namespace string, inputFile string) {
-	conf, err := config.GetConfig()
+func ConstructIngressesFromKubeCluster(cl client.Client, ingressList *networkingv1.IngressList) error {
```
LiorLieberman, line 37: `ConstructINgressesFromCluster`

pkg/i2gw/ingress2gateway.go
```diff
@@ -101,7 +56,7 @@ func ingresses2GatewaysAndHTTPRoutes(ingresses []networkingv1.Ingress) ([]gatewa
 	return aggregator.toHTTPRoutesAndGateways()
 }
 
-func outputResult(printer printers.ResourcePrinter, httpRoutes []gatewayv1beta1.HTTPRoute, gateways []gatewayv1beta1.Gateway) {
+func OutputResult(printer printers.ResourcePrinter, httpRoutes []gatewayv1beta1.HTTPRoute, gateways []gatewayv1beta1.Gateway) {
```
LiorLieberman, line 59: this feels print related. so I would turn this into `func (pr *printRunner) outputResult()`

cmd/print.go
```diff
@@ -43,55 +47,147 @@ var (
...
+		}
+		os.Exit(1)
+	}
+
+	i2gw.OutputResult(pr.resourcePrinter, httpRoutes, gateways)
+
+	return nil
+}
+
+func (pr *PrintRunner) GetNamespacedClient() client.Client {
```
LiorLieberman, line 100: maybe config.go or conf package?

cmd/print.go
```diff
@@ -43,55 +47,147 @@ var (
...
+	if len(ingressList.Items) == 0 {
+		msg := "No resources found"
+		if pr.namespace != "" {
+			fmt.Printf("%s in %s namespace\n", msg, pr.namespace)
+			os.Exit(1)
+		} else {
+			fmt.Print(msg)
+			os.Exit(1)
+		}
+	}
```
LiorLieberman, line 84: this to go into a conf or config.go?

levikobi (commented):
Thank you very much for your time and effort @Treenhan!
I have a couple of a couple of small suggestions

pkg/datasource/datasource.go
```diff
@@ -0,0 +1,77 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package datasource
+
```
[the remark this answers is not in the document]
Treenhan: The idea is after this PR, I will implement reading input files from remote sources like github as per https://github.com/kubernetes-sigs/ingress2gateway/issues/24. So this package will have more logic about fetching remote files very soon. Do you think that makes sense?

cmd/print.go
```diff
@@ -43,55 +46,139 @@ var (
...
+}
+
+// PrintGatewaysAndHttpRoutes performs necessary steps to digest and print
+// converted Gateways and HTTP Routes. The steps includes reading from the source,
+// construct ingresses, convert them, then print them out.
+func (pr *PrintRunner) PrintGatewaysAndHttpRoutes(cmd *cobra.Command, args []string) error {
+	err := pr.initializeResourcePrinter()
+	if err != nil {
+		fmt.Printf("Failed to initialize resrouce printer: %v", err)
+		os.Exit(1)
```
LiorLieberman, line 64: since we move this to a RunE cobra function we should just return errors
`fmt.errorf` maybe?

ditto for all os.exit in this file

cmd/print.go
```diff
@@ -43,55 +46,139 @@ var (
...
+		return nil
 	default:
-		return nil, fmt.Errorf("%s is not a supported output format", outputFormat)
+		return fmt.Errorf("%s is not a supported output format", pr.outputFormat)
 	}
+
 }
 
-// getNamespaceFilter returns a namespace filter, taking into consideration whether a specific
+// initializeNamespaceFilter generate the corret namespace filter, taking into consideration whether a specific
```
LiorLieberman, line 128: Can we add a nice longer comment here about the actual logic. It is not straight forward so a comment could be nice here

LiorLieberman (commented):
Thanks @Treenhan !

cmd/print.go
```diff
@@ -43,55 +46,139 @@ var (
...
+		// If the input file is specified, and we failed to get the namespace, use all namespaces.
+		pr.namespaceFilter = ns
+		return nil
 	}
 
-	if requestedNamespace == "" {
-		return getNamespaceInCurrentContext()
+	pr.namespaceFilter = pr.namespace
+	return nil
+}
```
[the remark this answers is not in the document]
LiorLieberman: I think it is better to have it in a struct for few reasons but here are the main ones; 

1. We initialize a printRunner object and hold most of the necessary things to do the print work in it. 
2. We do not pass (and copy, cause we passed by value) function params, instead we have methods on the object
3. For every new flag/feature we wont not need to change the functions signature

Those are the main things I have in mind. 
@levikobi I am curious to why do you think the testability would be more challenging?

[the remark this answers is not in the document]
levikobi: Well, the issue I had with it is that given this code block:
```go
	err = pr.initializeNamespaceFilter()
	if err != nil {
		return fmt.Errorf("Failed to initialize namespace filter: %v", err)
	}
```
You can't tell if (and which) variables initializeNamespaceFilter needs, and what it returns. I know it says initialize namespaceFilter, but then you have to go and look if there is actually a namespaceFilter variable.
As opposed to:
```go
	pr.namespaceFilter, err = getNamespaceFilter(namespace, allNamespaces)
	if err != nil {
		return fmt.Errorf("Failed to initialize namespace filter: %v", err)
	}
```
That's why, in my opinion, we are losing readability here.

@LiorLieberman regarding what I said about testability, I think it's just a bit harder to test because it's a bit less readable.

But anyways I see that both of you are in favor of this implementation so I'm good with it too 👍

pkg/datasource/datasource.go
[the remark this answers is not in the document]
levikobi: > I will implement reading input files from remote sources like github

That's a cool idea, I like it. Makes more sense to me now

LiorLieberman:
Thanks for working on it @Treenhan ! 

As discussed offline, can you open an issue for testing `constrcutIngressesFromCluster`?

/approve

levikobi:
Thanks @Treenhan!

/lgtm

PR #53, opened 2023-08-19
Support to specify providers to execute convert action

### LINKED ISSUE #45: Add support for running only some of the providers
Add a new flag that enables users to choose which concrete providers should be used when executing the 'print' command.

1. By default, all providers should be used.
2. A user should be able to choose more than one provider.

We can start working on it after #40 is merged.

jeffyjf:
Add a new flag `--provider`, in order to users to choose which concrete providers should be used when executing the 'print' command.

**What type of PR is this?**

/kind feature

Now, we already refactor the layout in order to support multiple providers #40. We also need a mechanism to enables users to choose which concrete providers should be used when executing the 'print' command.

Fixes #45

```release-note
Add a new flag `--provider`, that enables users to choose which concrete providers should be used when executing the 'print' command.
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -54,6 +54,9 @@ type PrintRunner struct {
 
 	// Only resources that matches this filter will be processed.
 	namespaceFilter string
+
+	// providers indicates which providers are used to execute convert action.
+	providers []string
 }
 
 // PrintGatewaysAndHTTPRoutes performs necessary steps to digest and print
@@ -69,7 +72,7 @@ func (pr *PrintRunner) PrintGatewaysAndHTTPRoutes(cmd *cobra.Command, _ []string
 		return fmt.Errorf("failed to initialize namespace filter: %w", err)
 	}
 
-	httpRoutes, gateways, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile)
+	httpRoutes, gateways, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers)
 	if err != nil {
 		return err
 	}
@@ -175,6 +178,9 @@ func newPrintCommand() *cobra.Command {
 	cmd.Flags().BoolVarP(&pr.allNamespaces, "all-namespaces", "A", false,
 		`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
 if specified with --namespace.`)
+
+	cmd.Flags().StringSliceVar(&pr.providers, "providers", i2gw.GetSupportedProviders(),
+		fmt.Sprintf("If present, the tool will try to convert only resources related to the specified providers, supported values are %v", i2gw.GetSupportedProviders()))
 
 	cmd.MarkFlagsMutuallyExclusive("namespace", "all-namespaces")
 	return cmd
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -34,7 +34,7 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
-func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile string) ([]gatewayv1beta1.HTTPRoute, []gatewayv1beta1.Gateway, error) {
+func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile string, providers []string) ([]gatewayv1beta1.HTTPRoute, []gatewayv1beta1.Gateway, error) {
 	conf, err := config.GetConfig()
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to get client config: %w", err)
@@ -50,9 +50,12 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 	var gateways []gatewayv1beta1.Gateway
 	var httpRoutes []gatewayv1beta1.HTTPRoute
 
-	providerByName := constructProviders(&ProviderConf{
+	providerByName, err := constructProviders(&ProviderConf{
 		Client: cl,
-	})
+	}, providers)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	resources := InputResources{}
 
@@ -120,16 +123,19 @@ func ConstructIngressesFromCluster(ctx context.Context, cl client.Client, ingres
 
 // constructProviders constructs a map of concrete Provider implementations
 // by their ProviderName.
-//
-// TODO (levikobi): Issue #45 - let users filter by provider name.
-func constructProviders(conf *ProviderConf) map[ProviderName]Provider {
+func constructProviders(conf *ProviderConf, providers []string) (map[ProviderName]Provider, error) {
 	providerByName := make(map[ProviderName]Provider, len(ProviderConstructorByName))
 
-	for name, newProviderFunc := range ProviderConstructorByName {
-		providerByName[name] = newProviderFunc(conf)
+	for _, requestedProvider := range providers {
+		requestedProviderName := ProviderName(requestedProvider)
+		newProviderFunc, ok := ProviderConstructorByName[requestedProviderName]
+		if !ok {
+			return nil, fmt.Errorf("%s is not a supported provider", requestedProvider)
+		}
+		providerByName[requestedProviderName] = newProviderFunc(conf)
 	}
 
-	return providerByName
+	return providerByName, nil
 }
 
 // extractObjectsFromReader extracts all objects from a reader,
@@ -215,4 +221,13 @@ func aggregatedErrs(errs field.ErrorList) error {
 		errMsg = fmt.Errorf("\n%w # %s", errMsg, err)
 	}
 	return errMsg
+}
+
+// GetSupportedProviders returns the names of all providers that are supported now
+func GetSupportedProviders() []string {
+	supportedProviders := make([]string, 0, len(ProviderConstructorByName))
+	for key := range ProviderConstructorByName {
+		supportedProviders = append(supportedProviders, string(key))
+	}
+	return supportedProviders
 }
--- a/pkg/i2gw/ingress2gateway_test.go
+++ b/pkg/i2gw/ingress2gateway_test.go
@@ -150,3 +150,73 @@ func Test_constructIngressesFromCluster(t *testing.T) {
 		})
 	}
 }
+
+func Test_constructProviders(t *testing.T) {
+	supportProviders := []string{"ingress-nginx"}
+	for _, provider := range supportProviders {
+		ProviderConstructorByName[ProviderName(provider)] = func(conf *ProviderConf) Provider { return nil }
+	}
+	testCases := []struct {
+		name              string
+		providers         []string
+		expectedProviders []string
+		expectedError     error
+	}{{
+		name:              "Test construct providers with default providers",
+		providers:         GetSupportedProviders(),
+		expectedProviders: supportProviders,
+		expectedError:     nil,
+	}, {
+		name:              "Test construct providers with provider that not supported",
+		providers:         []string{"ingress-nginx", "fake-provider"},
+		expectedProviders: []string{},
+		expectedError:     fmt.Errorf("%s is not a supported provider", "fake-provider"),
+	}}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			cl := fake.NewClientBuilder().WithRuntimeObjects([]runtime.Object{}...).Build()
+			providerByName, err := constructProviders(&ProviderConf{
+				Client: cl,
+			}, tc.providers)
+			if tc.expectedError != nil {
+				if err == nil {
+					t.Errorf("Expected error but got none")
+				} else if tc.expectedError.Error() != err.Error() {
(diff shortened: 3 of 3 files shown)


pkg/i2gw/ingress2gateway.go
```diff
@@ -122,11 +123,13 @@ func ConstructIngressesFromCluster(ctx context.Context, cl client.Client, ingres
...
 	providerByName := make(map[ProviderName]Provider, len(ProviderConstructorByName))
 
 	for name, newProviderFunc := range ProviderConstructorByName {
-		providerByName[name] = newProviderFunc(conf)
+		if slices.Contains(providers, string(name)) {
+			providerByName[name] = newProviderFunc(conf)
+		}
 	}
 
 	return providerByName
```
levikobi, line 135: Let's say I misspelled one of the provider names and passed ingres-nginx, instead of ingress-nginx. I expect to get an error saying that the provider I passed doesn't exist. So, we'll probably need to return an error as a second return type in this function.

levikobi: I think this should be unit tested as well. A simple table unit-test would be very nice here.

levikobi (commented):
Thanks @jeffyjf!
These are my main comments for now, but overall looks good to me.

jeffyjf:
@levikobi Thanks for you review. I will submit a commit againt at today later.

pkg/i2gw/ingress2gateway.go
```diff
@@ -120,16 +124,24 @@ func ConstructIngressesFromCluster(ctx context.Context, cl client.Client, ingres
...
 
 	for name, newProviderFunc := range ProviderConstructorByName {
-		providerByName[name] = newProviderFunc(conf)
+		if slices.Contains(providers, string(name)) {
+			providerByName[name] = newProviderFunc(conf)
+		}
 	}
 
-	return providerByName
+	return providerByName, nil
```
levikobi, line 144: You can take advantage of the fact that ProviderConstructorByName is a map to avoid the O(n^2) runtime complexity, and to make the code a bit more concise

```suggestion
	providerByName := make(map[ProviderName]Provider, len(ProviderConstructorByName))

	for _, requestedProvider := range providers {
		requestedProviderName := ProviderName(requestedProvider)
		newProviderFunc, ok := ProviderConstructorByName[requestedProviderName]
		if !ok {
			return nil, fmt.Errorf("%s is not a supported provider", requestedProvider)
		}
		providerByName[requestedProviderName] = newProviderFunc(conf)
	}

	return providerByName, nil
```

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -150,3 +151,47 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
 	}
 }
+
+func Test_constructProviders(t *testing.T) {
+	ingress := ingress(443, "ingress1", "namespace1")
+	testCases := []struct {
+		name           string
+		runtimeObjs    []runtime.Object
+		providers      []string
+		expectingError bool
```
levikobi, line 161: Maybe add an extra field named `expectedProviders`? Then you can compare the results with it.

Also, I don't think you really need the `runtimeObjs` here

pkg/i2gw/ingress2gateway.go
jeffyjf: Greate suggestion! Thanks very much.

pkg/i2gw/ingress2gateway_test.go
jeffyjf: Yes, you are right,  it's redundancy. I will remove it at next commit.

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -150,3 +151,59 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
 	}
 }
+
+func Test_constructProviders(t *testing.T) {
+	testCases := []struct {
+		name           string
+		providers      []string
+		expectingError bool
+	}{{
+		name:           "Test contruct providers with default providers",
```
levikobi, line 161:
```suggestion
		name:           "Test construct providers with default providers",
```

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -150,3 +151,59 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
+		providers:      GetSupportedProviders(),
+		expectingError: false,
+	}, {
+		name:           "Test contruct providers with provider that not supported",
```
levikobi, line 165:
```suggestion
		name:           "Test construct providers with provider that not supported",
```

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -150,3 +151,59 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
+					t.Errorf("Expected no error but got %v", err)
+				}
+				if len(tc.providers) != len(providerByName) {
+					t.Errorf("Expected contructed providers num is %d but got %d", len(tc.providers), len(providerByName))
+				}
+				for key := range providerByName {
+					if !slices.Contains(tc.providers, string(key)) {
+						t.Errorf("An unexpected provider %s was constructed", key)
+					}
+				}
```
levikobi, line 189: Same as the actual `constructProviders` function, you can take advantage of the fact that `providerByName` is a map, and avoid the usage of `slices.Contains`

Secondly, I think this test would be a bit easier to understand if you added an `expectedProviders` field to each test case.

levikobi (commented):
Thanks again @jeffyjf! This almost finished as far as I'm concerned, just couple more small issues

pkg/i2gw/ingress2gateway_test.go
jeffyjf: Thanks very much, I will fix these issues immediately.

cmd/print.go
```diff
@@ -54,6 +54,9 @@ type PrintRunner struct {
 
 	// Only resources that matches this filter will be processed.
 	namespaceFilter string
+
+	// providers indicates use which providers to execute convert action.
```
Xunzhuo, line 58: Maybe:
```
// providers indicates which providers are used to execute convert action.
 ```
?

File: pkg/i2gw/ingress2gateway_test.go [hunk 8]
```diff
@@ -150,3 +150,61 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func Test_GetSupportedProviders(t *testing.T) {
+	t.Run("Test GetSupportedProviders", func(t *testing.T) {
+		allProviders := GetSupportedProviders()
+		if len(allProviders) != len(ProviderConstructorByName) {
+			t.Errorf("The acutal number of the provider we supported is %d but we got the number is: %d",
```
Xunzhuo, line 200: `actual`
`providers`

pkg/i2gw/ingress2gateway.go
```diff
@@ -216,3 +222,12 @@ func aggregatedErrs(errs field.ErrorList) error {
 	}
 	return errMsg
 }
+
+// GetSupportedProviders return the name of all providers that we supported currently
```
Xunzhuo, line 226:
```
// GetSupportedProviders returns the names of all providers that are supported now
```

Xunzhuo (commented):
Some typos

pkg/i2gw/ingress2gateway_test.go
jeffyjf: thanks

cmd/print.go
jeffyjf: thanks

levikobi:
Thanks @jeffyjf!
I see that you addressed everything. If you can resolve the comments it would be great.

/lgtm
/assign @LiorLieberman

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -150,3 +150,61 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
+		expectingError: true,
+	}}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			cl := fake.NewClientBuilder().WithRuntimeObjects([]runtime.Object{}...).Build()
+			providerByName, err := constructProviders(&ProviderConf{
+				Client: cl,
+			}, tc.providers)
+			if tc.expectingError && err == nil {
+				t.Errorf("Expected error but got none")
```
LiorLieberman, line 177: can we check that we indeed get the error we expect?

pkg/i2gw/ingress2gateway_test.go
```diff
@@ -150,3 +150,61 @@ func Test_constructIngressesFromCluster(t *testing.T) {
...
+		expectedProviders []string
+		expectingError    bool
+	}{{
+		name:              "Test construct providers with default providers",
+		providers:         GetSupportedProviders(),
+		expectedProviders: GetSupportedProviders(),
+		expectingError:    false,
+	}, {
+		name:           "Test construct providers with provider that not supported",
+		providers:      []string{"ingress-nginx", "fake-provide"},
```
LiorLieberman, line 167: can we add `expectedProviders: []string{}` ?

pkg/i2gw/ingress2gateway.go
```diff
@@ -120,16 +123,19 @@ func ConstructIngressesFromCluster(ctx context.Context, cl client.Client, ingres
 
 // constructProviders constructs a map of concrete Provider implementations
 // by their ProviderName.
-//
-// TODO (levikobi): Issue #45 - let users filter by provider name.
-func constructProviders(conf *ProviderConf) map[ProviderName]Provider {
+func constructProviders(conf *ProviderConf, providers []string) (map[ProviderName]Provider, error) {
```
LiorLieberman, line 126: `ProviderConf` should not be per provider? I know no one is using it now but IIRC that was the purpose of it.

cmd/print.go
```diff
@@ -176,6 +179,9 @@ func newPrintCommand() *cobra.Command {
 		`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
 if specified with --namespace.`)
 
+	cmd.Flags().StringSliceVarP(&pr.providers, "providers", "p", i2gw.GetSupportedProviders(),
+		`Use these providers specified by the option to execute convert actions`)
```
LiorLieberman, line 183: maybe `fmt.Sprintf("If present, the tool will try to convert only resources related to the specified providers, supported values are %v", i2gw.GetSupportedProviders())`

LiorLieberman (commented):
Thanks @jeffyjf ! 
Left a few comments

pkg/i2gw/ingress2gateway.go
levikobi: I think `ProviderConf` can be shared between all providers, at least for now.
If any provider needs unique settings, it can instantiate them itself.

At some point, we might want to push specific user configurations for a specific provider - then we'll have to rethink it.
But again, for now, IMO it's good enough. Curious what's your opinion about it @LiorLieberman

LiorLieberman: sgtm if you are good with it.

jeffyjf:
@levikobi @LiorLieberman @Xunzhuo Hello, everyone. I need your help. I encountered a troubled issue when I refactor `Test_constructProviders`.  The package "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/ingressnginx" never be impoted when run UT. So the globale variable "i2gw.ProviderConstructorByName" always is null map. But, in order to simulate real scene, "i2gw.ProviderConstructorByName" should be filled someting. Moreover, I can't fill it by import "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/ingressnginx" in "pkg/i2gw/ingress2gateway_test.go", because that  will result in cycle import.

So, How can I resolve this issue? Have similar exemple in this or other project? Thank you very much.

jeffyjf:
I add a [mock provider](https://github.com/jeffyjf/ingress2gateway/blob/providers_flag/pkg/i2gw/mock_provider.go) in [pkg/i2gw](https://github.com/jeffyjf/ingress2gateway/tree/providers_flag/pkg/i2gw) in order to ensure [ProviderConstructorByName](https://github.com/jeffyjf/ingress2gateway/blob/providers_flag/pkg/i2gw/provider.go#L33) is  not null. But, I can't sure this way whether is appropriate. Look forward to your comments.

levikobi:
I think the easiest way to tackle this problem will be to just initialize `ProviderConstructorByName` within your unit test. Simply add "ingress-nginx" and some other providers with a zero value `ProviderConstructor` implementation.

I do have to say I am not totally happy with this solution, but it's going to get the job done for now.
Maybe in the future we will shift away from the global  `ProviderConstructorByName` variable. 

Please let me know if that makes sense to you @jeffyjf, and if you have any further questions

jeffyjf:
@levikobi Got it. Thanks very much.

jeffyjf:
Thanks @levikobi I will try to edit them and add more necessary informations. But, I am not good at english. If I have any mistake, please point out that.

levikobi:
Thanks @jeffyjf 
/lgtm

LiorLieberman:
Thanks for a great contribution @jeffyjf. 
/approve

LiorLieberman:
@jeffyjf can you open a PR to update the README.md file flag table with the new flag your contributed?

jeffyjf:
@LiorLieberman Sure, I'm pleasure.

PR #69, opened 2023-09-28
added ingresses-only flag

### LINKED ISSUE #67: Add a `ingresses-only` flag
Add a `ingresses-only` flag to scope the tool to only convert resources without looking for providers CRDs

**Why this is needed**:
Provider specific support is in early stage support now with only few providers support the conversion. Users might want to scope the tool invocation to only read and convert their ingresses

/label good-first-issue
/help-wanted

camperjett:
**What type of PR is this?**
/kind feature

this adds a new flag `ingresses-only` to scope the tool to only convert resources without looking for providers CRDs

Fixes #67 

```release-note
NONE
```
/cc @LiorLieberman

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,34 +19,39 @@ Note: Ingress2gateway is not intended to copy annotations from Ingress to Gatewa
 ## Build from Source
 
 1. Ensure that your system meets the following requirements:
-- Install Git: Make sure Git is installed on your system to clone the project repository.
-- Install Go: Make sure the go language is installed on your system. You can download it from the official website (https://golang.org/dl/) and follow the installation instructions.
 
-2. Clone the project repository
-```shell
-git clone https://github.com/kubernetes-sigs/ingress2gateway.git && cd ingress2gateway
-```
+   - Install Git: Make sure Git is installed on your system to clone the project repository.
+   - Install Go: Make sure the go language is installed on your system. You can download it from the official website (https://golang.org/dl/) and follow the installation instructions.
 
-3. Build the project
-```shell
-make build
-```
+1. Clone the project repository
+
+   ```shell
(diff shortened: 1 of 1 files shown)


levikobi (changes requested):
Thank you for your contribution @camperjett!

Removing the gateway objects is not exactly what we had in mind here. What we intended to do was to not translate provider-specific functionality by not running provider converters.
Maybe you can take a look at https://github.com/kubernetes-sigs/ingress2gateway#usage for some reference.

Let me know if you have any questions.

camperjett:
Okay. This is similar to passing empty list of providers in function `i2gw.ToGatewayAPIResources(...)` right? Should I do that?

levikobi:
> Okay. This is similar to passing empty list of providers in function `i2gw.ToGatewayAPIResources(...)` right? Should I do that? (In file cmd/print.go, Inside function `PrintGatewaysAndHTTPRoutes(...)`)

Yes that seems about right

One more thing that comes to mind, this flag should probably be mutually exclusive with the providers flag

cmd/print.go
```diff
@@ -182,6 +190,7 @@ if specified with --namespace.`)
 	cmd.Flags().StringSliceVar(&pr.providers, "providers", i2gw.GetSupportedProviders(),
 		fmt.Sprintf("If present, the tool will try to convert only resources related to the specified providers, supported values are %v", i2gw.GetSupportedProviders()))
 
+	cmd.Flags().BoolVar(&pr.ingressesOnly, "ingresses-only", false, "If present, the tool will only print the converted Ingresses.")
```
levikobi, line 193: Should be mutually exclusive with the providers flag

camperjett:
PTAL at 
#71

PR #78, opened 2023-10-19
feat: print all resources

### LINKED ISSUE #73: print all resources when reading from files
When thinking about how this tool would be used in a real environment, we can have a user willing to convert a set of yaml files that use ingress resources into the same set of yaml files using gateway resources. As of today, when the flag `input_file` is used, the tool gets only the ingress resources and prints the converted ones.

When a user has a yaml file containing ingress resources and other kinds of resources (deployment, services, ...), a simple one-liner could enable an immediate conversion to have a ready-to-use file.

For example, by performing an operation similar to the following:
```bash
./ingress2gateway print --input_file old_file.yaml -A --providers=kong --print-all-resources > new_file.yaml
``` 
a user can have a new yaml file ready-to-use.

mlavacca:
**What type of PR is this?**

/kind feature

When reading the resources from a file, it is now possible to display all the resources, not only the Gateway API ones. Besides, the `PROVIDER.md` file has been properly formatted.

Fixes #73 

```release-note
By setting the flag `--all-resources`, it is now possible to print all the resources, not only the Gateway API ones. This flag is available only when reading the resources from a file.
```


pkg/i2gw/provider.go
```diff
@@ -32,6 +34,10 @@ import (
 // func at startup.
 var ProviderConstructorByName = map[ProviderName]ProviderConstructor{}
 
+// ProviderConfByName ProviderConstructorByName is a map of ProviderConf by a
+// provider name. Different Provider implementations should add their conf at startup.
+var ProviderConfByName = map[ProviderName]ProviderConf{}
```
levikobi, line 39: If these configurations will be static and provider-specific, why don't we initialize them as part of each `NewProvider` function?

levikobi (commented):
Thank you @mlavacca!
I have some questions regarding a few things but overall it's seems like another great contribution to me.

pkg/i2gw/provider.go
```diff
@@ -51,13 +70,14 @@ type ProviderConf struct {
...
 	ResourceConverter
+	ResourceFilter
 }
 
 type CustomResourceReader interface {
 
 	// ReadResourcesFromCluster reads custom resources associated with
 	// the underlying Provider implementation from the kubernetes cluster.
-	ReadResourcesFromCluster(ctx context.Context, customResources interface{}) error
+	ReadResourcesFromCluster(ctx context.Context, cl client.Client, customResources interface{}) error
```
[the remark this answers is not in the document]
mlavacca: Because when reading from a file we have no necessity to have a client. Before this change, in such a context and without a cluster up and running the client instantiation was failing, leading to an error in the tool.

pkg/i2gw/provider.go
mlavacca: Yes, it makes total sense. Addressed in e902610

cmd/print.go
```diff
@@ -164,15 +189,24 @@ func newPrintCommand() *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "print",
 		Short: "Prints HTTPRoutes and Gateways generated from Ingress resources",
-		RunE:  pr.PrintGatewaysAndHTTPRoutes,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if pr.allResources && pr.inputFile == "" {
+				return errors.New("--all-resources flag can be set only when --input-file is set")
+			}
+			return nil
+		},
```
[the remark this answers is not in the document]
mlavacca: The idea behind this is to properly convert a yaml file to the equivalent one with ingress resources replaced by Gateway API ones. I do not think that allowing this flag when reading from a cluster would be helpful in any case, as it'd basically be a dump of a cluster state into a unique giant print. Furthermore, we'd need a lot of custom logic to sanitize all the objects (e.g. status, timestamps, uid, etc.). I think we could leave this flag as it is for now, and if some use-case related to applying to cluster-based reading as well we can tackle it in a subsequent PR. What do you think?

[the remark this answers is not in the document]
LiorLieberman: I agree, I also cant think of a good case where printing all from clusters will be useful. 
Also plus one that this can be addresses in a subsequent PR if ever requested.

cmd/print.go
```diff
@@ -60,10 +66,11 @@ type PrintRunner struct {
 	providers []string
 }
 
-// PrintGatewaysAndHTTPRoutes performs necessary steps to digest and print
-// converted Gateways and HTTP Routes. The steps include reading from the source,
-// construct ingresses, convert them, then print them out.
-func (pr *PrintRunner) PrintGatewaysAndHTTPRoutes(cmd *cobra.Command, _ []string) error {
+// PrintResources performs necessary steps to digest and print
+// converted Gateways and HTTP Routes along with other resources.
```
LiorLieberman, line 70:
```suggestion
// converted Gateways and HTTPRoutes along with other resources.
```

cmd/print.go
```diff
@@ -78,12 +85,30 @@ func (pr *PrintRunner) PrintGatewaysAndHTTPRoutes(cmd *cobra.Command, _ []string
...
+			fmt.Printf("# Error printing %s/%s %s with: %v\n",
+				additionalResources[i].GetNamespace(),
+				additionalResources[i].GetName(),
+				additionalResources[i].GetObjectKind().GroupVersionKind().Kind,
+				err)
+		}
+	}
+
 	if len(httpRoutes)+len(gateways) == 0 {
 		msg := "No resources found"
```
LiorLieberman, line 113: two questions; 
1) why do we print the additional resources before the gateways and http routes? Reversing the order would make more sense imo. 

2) I think "No resources found" message needs to be rephrased. Can we have a case where you have `additionalResources` but have no `Gateway` and `HTTPRoutes`?

cmd/print.go
LiorLieberman, line 197: consider [cmd.MarkFlagsRequiredTogether](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsRequiredTogether) instead of PreRunE

pkg/i2gw/provider.go
```diff
@@ -32,17 +34,34 @@ import (
 // func at startup.
 var ProviderConstructorByName = map[ProviderName]ProviderConstructor{}
 
+// ProviderConfByName ProviderConstructorByName is a map of ProviderConf by a
+// provider name. Different Provider implementations should add their conf at startup.
+// var ProviderConfByName = map[ProviderName]ProviderConf{}
```
LiorLieberman, line 39: not sure I understands this?

pkg/i2gw/provider.go
```diff
@@ -32,17 +34,34 @@ import (
...
 type ProviderName string
 
 // ProviderConstructor is a construction function that constructs concrete
 // implementations of the Provider interface.
-type ProviderConstructor func(conf *ProviderConf) Provider
+type ProviderConstructor func() Provider
 
 // ProviderConf contains all the configuration required for every concrete
 // Provider implementation.
 type ProviderConf struct {
```
LiorLieberman, line 50: why did we remove the client from `ProviderConf`?  Initially we assume there can be providers that needs a client with different settings. 

I have no strong feelings though, wdyt? 
@mlavacca @levikobi

pkg/i2gw/provider.go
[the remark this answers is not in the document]
LiorLieberman: Left a comment above related to it. Can we retain the client within ProviderConf and fix the logic so reading from file logic would not use it?

LiorLieberman (commented):
Thanks for all the improvements.

I have a general thought I have after reviewing this.
My understanding is that this PR is driven by the idea that users will want to convert a yaml/json file, with a lot of objects (or a group of files when we get to support it). Then when reading from those files, instead of printing only the Gateway API resources - it will print all the resources plus the new GatewayAPI resources and without the resources that the Gateway API resources replaces.

If this is correct - this filtering logic feels a bit detached to the actual logic. I mean, lets take Kong as an example. If at Kong, UDPIngress and TCPIngress is replaced by the Gateway equivalents, my mind is when the provider is reading the CRD, it decides whether the Gateway resources replace it and based on this decide whether it prints it or not.

thoughts?  I am definitely up for debate here

cmd/print.go
mlavacca: I initially considered using `cmd.MarkFlagsRequiredTogether`, but it turned out it cannot be used here, as `--input-file` can be used without `--all-resource`, while the other way around should be forbidden. As far as I get from `cmd.MarkFlagsRequiredTogether`, both flags are marked as required together, and is not possible to express the above.

pkg/i2gw/provider.go
mlavacca: I have no strong feelings either. I'll add again the client and ensure that it is not instantiated when reading from a file

mlavacca:
> My understanding is that this PR is driven by the idea that users will want to convert a yaml/json file, with a lot of objects (or a group of files when we get to support it). Then when reading from those files, instead of printing only the Gateway API resources - it will print all the resources plus the new GatewayAPI resources and without the resources that the Gateway API resources replaces.

100% accurate

> If this is correct - this filtering logic feels a bit detached to the actual logic. I mean, lets take Kong as an example. If at Kong, UDPIngress and TCPIngress is replaced by the Gateway equivalents, my mind is when the provider is reading the CRD, it decides whether the Gateway resources replace it and based on this decide whether it prints it or not.

Yes, with this PR, the printing logic has been detached from the conversion logic. In the conversion logic for the CRDs, the output will be the corresponding Gateway API objects (let's say for a `TCPIngress` as input, the output will have `Gateway`+`TCPRoute`). In the printing logic, all the input objects are taken as input and after proper filtering displayed. 
Do you see this flow as a problem?

pkg/i2gw/provider.go
mlavacca: good catch, this is a leftover 👍

cmd/print.go
LiorLieberman: fair, thanks

LiorLieberman:
> Yes, with this PR, the printing logic has been detached from the conversion logic. In the conversion logic for the CRDs, the output will be the corresponding Gateway API objects (let's say for a `TCPIngress` as input, the output will have `Gateway`+`TCPRoute`). In the printing logic, all the input objects are taken as input and after proper filtering displayed.
> Do you see this flow as a problem?

It is just that I think it should sit closer to the logic. Detaching them from each other is confusing imo. What benefits you see to keep them detached?

mlavacca:
> It is just that I think it should sit closer to the logic. Detaching them from each other is confusing imo. What benefits you see to keep them detached?

The fact is that they are separated logics, they do different things and are subsequent operations. How would you restructure such a piece of code? It's not clear to me 🤔

mlavacca:
@LiorLieberman I've added again the client in the conf. PTAL

cmd/print.go
LiorLieberman: lets keep a warning or some other indication if the user has no "Gateway resources" (currently only referring to gateway and httproutes)

pkg/i2gw/providers/ingressnginx/ingressnginx.go
```diff
@@ -32,14 +32,17 @@ type Provider struct {
 	conf *i2gw.ProviderConf
 
 	*resourceReader
+	*resourceFilter
 	*converter
 }
 
 // NewProvider constructs and returns the ingress-nginx implementation of i2gw.Provider.
 func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
+	conf.FilteredObjects = filteredObjects
```
levikobi, line 41: If I'm not mistaken, all providers share the same ProviderConf object. So, different providers will override the `FilteredObjects` with their own filteredObjects list, meaning the last constructed provider's filtered objects will last

mlavacca: Nice catch @levikobi! I've changed, now the configuration is passed by value and the objects are not overwritten

LiorLieberman:
> > It is just that I think it should sit closer to the logic. Detaching them from each other is confusing imo. What benefits you see to keep them detached?
> 
> The fact is that they are separated logics, they do different things and are subsequent operations. How would you restructure such a piece of code? It's not clear to me 🤔

I think the provider specific logic should 1) read all the resources it cares about, regardless of whether it is from a file or from the cluster 2) replace with the equivalent GatewayAPI resources 3) output. 

So the workflow for converting from a file: 

1) parse the file and construct a list of resources I am interested in (Ingresses + provider specific CRDs). all other objects unrelated objects - keep in a separate list to be used if `--all-resources` flag is true
2) convert them to the Gateway equivalent
3) print them, if `--all-resources` also print the that list we kept initially

We do not have any implementation that actually reads CRDs from cluster/file and convert them yet, maybe it better to revisit this debate once we have one.

mlavacca:
> > > It is just that I think it should sit closer to the logic. Detaching them from each other is confusing imo. What benefits you see to keep them detached?
> > 
> > 
> > The fact is that they are separated logics, they do different things and are subsequent operations. How would you restructure such a piece of code? It's not clear to me 🤔
> 
> I think the provider specific logic should 1) read all the resources it cares about, regardless of whether it is from a file or from the cluster 2) replace with the equivalent GatewayAPI resources 3) output.
> 
> So the workflow for converting from a file:
> 
> 1. parse the file and construct a list of resources I am interested in (Ingresses + provider specific CRDs). all other objects unrelated objects - keep in a separate list to be used if `--all-resources` flag is true
> 2. convert them to the Gateway equivalent
> 3. print them, if `--all-resources` also print the that list we kept initially

This is a good representation of the current workflow 👍 

> We do not have any implementation that actually reads CRDs from cluster/file and convert them yet, maybe it better to revisit this debate once we have one.

I plan to implement CRD conversion for Kong pretty soon. I think we can bring this discussion there, what do you think?

LiorLieberman:
> I plan to implement CRD conversion for Kong pretty soon. I think we can bring this discussion there, what do you think?

As we discussed offline, lets keep this one open and get back to it once you do the CRD implementation. 

/hold

cmd/print.go
mlavacca: closing thread as no more relevant after the latest changes

mlavacca:
I've revamped this PR after a long time of inactivity. I drastically changed the initial approach after all the changes that have been merged. The PR is 95% complete, there is some documentation missing and some minor cleanup, but I'd like to have the idea validated before adding the last bits.

Please, folks, take another look :)

/cc @LiorLieberman @levikobi

pkg/i2gw/ingress2gateway.go
```diff
@@ -90,6 +133,29 @@ func readProviderResourcesFromFile(ctx context.Context, providerByName map[Provi
...
+	unstructuredObjects, err := ExtractObjectsFromReader(bytes.NewReader(stream), namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract objects: %w", err)
+	}
+	for _, o := range unstructuredObjects {
+		if o.GetNamespace() != namespace {
+			continue
+		}
+		if _, ok := FilteredResources[o.GetObjectKind().GroupVersionKind().GroupKind()]; ok {
+			continue
```
LiorLieberman, line 152: Trying to think if there is a different way to do it. I less like the fact that every provider needs to register its resources and that we depends on it.. I have not thought on anything smart yet. 
But.. we should have all the resources the provider cares about (which are the resources that we need to filter out from the file) in its `storage`.. Maybe we can leverage that?

cmd/print.go
```diff
@@ -75,102 +81,31 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
 		}
 		fmt.Println(msg)
+		return
+	}
+
+	for _, r := range resources {
+		err := pr.resourcePrinter.PrintObj(r, os.Stdout)
+		if err != nil {
+			fmt.Printf("# Error printing %s %s: %v\n", r.GetName(), r.GetObjectKind().GroupVersionKind().Kind, err)
+		}
```
levikobi, line 108: That's a lot of duplicated code that you removed, awesome

pkg/i2gw/ingress2gateway.go
levikobi: I guess we can add a `GetCRDs()` method to the `provider` interface. We have access to all the providers in `i2gw` package, so we can do the registration in this package. Instead of having each one registering his own CRDs

levikobi (commented):
@mlavacca you moved some logic from `common` to `i2gw`, can you please explain why?

cmd/print.go
```diff
@@ -231,6 +166,12 @@ func newPrintCommand() *cobra.Command {
 		Use:   "print",
 		Short: "Prints Gateway API objects generated from ingress and provider-specific resources.",
 		RunE:  pr.PrintGatewayAPIObjects,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
mlavacca: `MarkFlagsRequiredTogether` can't be used as it marks all the flags as mandatory when at least one of them is set. In this case, we want to mark `--input-file` as required when `--all-resources` is set, but not the opposite.

pkg/i2gw/providers/apisix/resource_reader.go
```diff
@@ -39,7 +38,7 @@ func (r *resourceReader) readResourcesFromCluster(ctx context.Context) (*storage
 	// read apisix related resources from cluster.
 	storage := newResourcesStorage()
 
-	ingresses, err := common.ReadIngressesFromCluster(ctx, r.conf.Client, ApisixIngressClass)
+	ingresses, err := i2gw.ReadIngressesFromCluster(ctx, r.conf.Client, ApisixIngressClass)
```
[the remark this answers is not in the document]
mlavacca: I needed to move the `resource-reader` files to the i2gw package to avoid the import cycle as with this change we read resources not only from the provider packages but from i2gw as well, as we want to read resources that do not have to be converted.

pkg/i2gw/ingress2gateway.go
mlavacca: I followed @levikobi's suggestion and added a `GetCRDs` method to the provider interface. Let me know if it looks better now. In that case, I'll add documentation about the new method for the providers to implement.

cmd/print.go
```diff
@@ -246,6 +187,9 @@ func newPrintCommand() *cobra.Command {
 		`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
 if specified with --namespace.`)
 
+	cmd.Flags().BoolVar(&pr.allResources, "all-resources", false,
+		`If present, list all the objects across the selected namespaces. This flag can be set only when reading from a file.`)
```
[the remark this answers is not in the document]
mlavacca: It makes sense, I've reworded the comment to include this information 👍

LiorLieberman (commented):
Thanks @mlavacca, left some comments. 
the move of logic from common to i2gw a bit bothers me because the function moved are actually common functions. Maybe we can come up with different package structure that will still accommodate the changes you introduced?

PR #79, opened 2023-10-20
change httproute name

mlavacca:
**What type of PR is this?**

Fixes #

```release-note

```


pkg/i2gw/providers/common/utils.go
```diff
@@ -99,6 +101,10 @@ func NameFromHost(host string) string {
 	return step2
 }
 
+func HTTPRouteName(ingressName, host string) string {
+	return fmt.Sprintf("%s-%s", ingressName, NameFromHost(host))
+}
```
levikobi, line 106: Should we name this function in a more generic term? I think we will probably use it for other route types in the future

levikobi (commented):
Thanks @mlavacca
Overall looks good to me, one minor question though

pkg/i2gw/providers/common/utils.go
LiorLieberman: +1 
```suggestion
func RouteName(ingressName, host string) string {
	return fmt.Sprintf("%s-%s", ingressName, NameFromHost(host))
}
```

LiorLieberman:
Thanks @mlavacca ! Looks good to me with +1 on Kobi's comment. 

/lgtm

Ping me here once this fixed and I will approve it

pkg/i2gw/providers/common/utils.go
mlavacca: Yeah, it makes sense. I changed it to `RouteName` @LiorLieberman

mlavacca:
@LiorLieberman lgmt label is mssing as there were changes after your lgtm stamp.

PR #80, opened 2023-10-23
fix: version removed from plugin group

mlavacca:
**What type of PR is this?**

/kind bug

The plugin group includes the version, which is wrong. This PR removes the version from the group.

Fixes #

```release-note
NONE
```


Xunzhuo (approved):
/lgtm

PR #86, opened 2023-10-26
Kong TCPIngress translation

### LINKED ISSUE #85: [Kong provider] TCPIngress translation

Kong Ingress controller leverages a CRD called `TCPIngress` to expose TCP routes. Such a resource should be converted into `Gateway` + `TCPRoute`s + `TLSRoute`s.

mlavacca:
**What type of PR is this?**

/kind feature

Fixes #85 

Additionally, go has been bumped to 1.21

```release-note
The Kong `TCPIngress` resources are properly translated into `Gateway`s, `TCPRoute`s, and `TLSRoute`s.
```


pkg/i2gw/providers/common/utils.go
```diff
@@ -95,14 +95,18 @@ func NameFromHost(host string) string {
...
 	return step2
 }
 
 func RouteName(ingressName, host string) string {
-	return fmt.Sprintf("%s-%s", ingressName, NameFromHost(host))
+	routeName := ingressName
+	if host != "" {
+		routeName = fmt.Sprintf("%s-%s", ingressName, NameFromHost(host))
+	}
+	return routeName
```
LiorLieberman, line 109: why?

pkg/i2gw/providers/kong/consts.go
```diff
@@ -26,6 +31,23 @@ const (
 	pluginsKey = "plugins"
 )
 
+const (
+	v1beta1Version string = "v1beta1"
+
+	kongResourcesGroup gatewayv1beta1.Group = "configuration.konghq.com"
+
+	kongPluginKind gatewayv1beta1.Kind = "KongPlugin"
+	tcpIngressKind gatewayv1beta1.Kind = "TCPIngress"
```
LiorLieberman, line 40: why gatewayv1beta1 and not just strings or kubernetes v1 kinds/groups?

pkg/i2gw/providers/kong/kong.go
```diff
@@ -51,25 +54,23 @@ func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
 // ToGatewayAPI converts the received i2gw.InputResources to i2gw.GatewayResources
 // including the kong specific features.
 func (p *Provider) ToGatewayAPI(_ i2gw.InputResources) (i2gw.GatewayResources, field.ErrorList) {
-	return p.converter.convert(p.storage)
+	return p.convert(p.storage)
```
LiorLieberman, line 57: why moving a way from the common approach across the current implementations? 

We have a converter and the the converter is the one to have the method `convert`

pkg/i2gw/providers/kong/kong.go
```diff
@@ -34,15 +34,18 @@ func init() {
...
+
+	*resourceReader
+	*converter
 }
 
 // NewProvider constructs and returns the kong implementation of i2gw.Provider.
 func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
 	return &Provider{
-		storage:        newResourcesStorage(),
+		conf: conf,
```
LiorLieberman, line 47: I dont think we ever use this. 
the resourceReader is the one that holds the conf

pkg/i2gw/providers/kong/storage.go
```diff
@@ -17,16 +17,19 @@ limitations under the License.
...
 )
 
 type storage struct {
-	Ingresses map[types.NamespacedName]*networkingv1.Ingress
+	Ingresses    map[types.NamespacedName]*networkingv1.Ingress
+	TCPIngresses []kongv1beta1.TCPIngress
 }
 
-func newResourcesStorage() *storage {
+func newStorage() *storage {
```
LiorLieberman, line 30: nit: any reason for having a different function name than other implementations?

pkg/i2gw/providers/kong/converter.go
```diff
@@ -43,24 +45,37 @@ func newConverter() *converter {
 	}
 }
 
-func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.ErrorList) {
+func (c *converter) convert(storage storage) (i2gw.GatewayResources, field.ErrorList) {
```
LiorLieberman, line 48: nit: I think when you call convert you do pass a pointer to storage

go.mod
LiorLieberman, line 1: why is this needed? can we revert this?

pkg/i2gw/providers/common/utils.go
mlavacca: because if there is no host set the resulting name is `<ingress-name>-`

LiorLieberman: oh ok, isnt it simpler to just wrap it in trimsuffix?

mlavacca: the conditional needs to be there in any case, hence I do not see any difference.

pkg/i2gw/providers/kong/converter.go
mlavacca: nope, I don't

pkg/i2gw/providers/kong/storage.go
mlavacca: Changed to be coherent with others.

pkg/i2gw/providers/kong/kong.go
mlavacca: yep, removed

pkg/i2gw/providers/kong/kong.go
mlavacca: no particular reason, there are equivalent. Reverted to be coherent with other implementations.

pkg/i2gw/providers/kong/consts.go
mlavacca: it was more convenient as we use these constants in gateway `ObjectRef`s. I agree that `string` fits better though. Changed.

pkg/i2gw/providers/common/utils.go
LiorLieberman: so I am probably missing something, why does it need to be there?

go.mod
LiorLieberman: go.mod and go.sum still appear in files changed

mlavacca: yeah, you are right, they do. The changes are needed by the current implementation.

pkg/i2gw/providers/common/utils.go
mlavacca: oh I did not get your point initially. Yes, we can wrap it in a `trimsuffix`. Changed 👍

LiorLieberman: so before this change, `routeName`used to be <ingressname>-all-hosts when`host==""`. and not it will be <ingressName> ? 
I dont mind just want to check I understand correctly and if we should not change other tests to align with this

LiorLieberman: trimsuffix wont retain the same functionality, whats we want to have when host=="" ?

pkg/i2gw/providers/kong/converter.go
LiorLieberman: any reason not to? We are copying the whole storage struct if we pass by value

pkg/i2gw/ingress2gateway.go
```diff
@@ -127,3 +132,64 @@ func GetSupportedProviders() []string {
...
+			// 16 is the maximum number of addresses a Gateway can have
+			if len(g.Spec.Addresses) > 16 {
+				fieldPath := field.NewPath(fmt.Sprintf("%s/%s", nn.Namespace, nn.Name)).Child("spec").Child("addresses")
+				errs = append(errs, field.Invalid(fieldPath, g, "error while merging gateway listeners: a gateway cannot have more than 16 addresses"))
+			}
+		}
+	}
+
+	return newGateways, errs
+}
```
LiorLieberman, line 195: Mind adding a comment that this behavior is likely to change following https://github.com/kubernetes-sigs/gateway-api/pull/1863/ ?

mlavacca: added

PR #89, opened 2023-10-26
feat: kong implementationSpecific pathType

### LINKED ISSUE #88: [kong] Convert `ImplementationSpecific` path type

Support `ImplementationSpecific` path conversion for Kong provider. The `ImplementationSpecific` path type should be converted into `RegularExpression` within Kong provider conversion functions.

mlavacca:
**What type of PR is this?**

/kind feature

Fixes #88

```release-note
Kong supports `ImplementationSpecific` as `PathType` and converts it into `RegularExpression.`
```

### DIFF
--- a/PROVIDER.md
+++ b/PROVIDER.md
@@ -61,7 +61,10 @@ func (r *resourceReader) ReadResourcesFromFiles(ctx context.Context, customResou
 ```
 3. Create a struct named `converter` which implements the `ResourceConverter` interface in a file named `converter.go`.
 The implemented `ToGatewayAPI` function should simply call every registered `featureParser` function, one by one.
-Take a look at `ingressnginx/converter.go` for example.
+Take a look at `ingressnginx/converter.go` for an example.
+The `ImplementationSpecificOptions` struct contains the handlers to customize native ingress implementation-specific fields.
+Take a look at `kong/converter.go` for an example.
+
 ```go
 package examplegateway
 
@@ -74,6 +77,7 @@ type converter struct {
 	conf *i2gw.ProviderConf
 
 	featureParsers []i2gw.FeatureParser
+	implementationSpecificOptions i2gw.ProviderImplementationSpecificOptions
 }
 
 // newConverter returns an ingress-nginx converter instance.
@@ -82,6 +86,9 @@ func newConverter(conf *i2gw.ProviderConf) *converter {
 		conf: conf,
 		featureParsers: []i2gw.FeatureParser{
 			// The list of feature parsers comes here.
+		},
+		implementationSpecificOptions: i2gw.ProviderImplementationSpecificOptions{
+			// The list of the implementationSpecific ingress fields options comes here.
 		},
 	}
 }
--- a/pkg/i2gw/provider.go
+++ b/pkg/i2gw/provider.go
@@ -73,6 +73,16 @@ type ResourceConverter interface {
 	ToGatewayAPI(resources InputResources) (GatewayResources, field.ErrorList)
 }
 
+// HTTPPathMatchOption is an option to customize the ingress implementationSpecific
+// match type conversion.
+type HTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
+
+// ImplementationSpecificOptions contains all the pointers to implementation-specific
+// customization functions.
+type ImplementationSpecificOptions struct {
+	HTTPPathmatch HTTPPathMatchOption
+}
+
 // InputResources contains all Ingress objects, and Provider specific
 // custom resources.
 type InputResources struct {
--- a/pkg/i2gw/provider.go
+++ b/pkg/i2gw/provider.go
@@ -73,14 +73,15 @@ type ResourceConverter interface {
 	ToGatewayAPI(resources InputResources) (GatewayResources, field.ErrorList)
 }
 
-// HTTPPathMatchOption is an option to customize the ingress implementationSpecific
+// ImplementationSpecificHTTPPathMatchOption is an option to customize the ingress implementationSpecific
 // match type conversion.
-type HTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
+type ImplementationSpecificHTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
 
-// ImplementationSpecificOptions contains all the pointers to implementation-specific
-// customization functions.
-type ImplementationSpecificOptions struct {
-	HTTPPathmatch HTTPPathMatchOption
+// ProviderImplementationSpecificOptions contains all the pointers to implementation-specific
+// customization functions. Such functions will be called by the common package to customize
+// the provider-specific behavior for all the implementation-specific fields of the API.
+type ProviderImplementationSpecificOptions struct {
+	ToImplementationSpecificHTTPPathMatch ImplementationSpecificHTTPPathMatchOption
 }
 
 // InputResources contains all Ingress objects, and Provider specific
--- a/pkg/i2gw/provider.go
+++ b/pkg/i2gw/provider.go
@@ -73,15 +73,15 @@ type ResourceConverter interface {
 	ToGatewayAPI(resources InputResources) (GatewayResources, field.ErrorList)
 }
 
-// ImplementationSpecificHTTPPathMatchOption is an option to customize the ingress implementationSpecific
+// ImplementationSpecificHTTPPathTypeMatchConverter is an option to customize the ingress implementationSpecific
 // match type conversion.
-type ImplementationSpecificHTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
+type ImplementationSpecificHTTPPathTypeMatchConverter func(*gatewayv1beta1.HTTPPathMatch)
 
-// ProviderImplementationSpecificOptions contains all the pointers to implementation-specific
-// customization functions. Such functions will be called by the common package to customize
-// the provider-specific behavior for all the implementation-specific fields of the API.
+// ProviderImplementationSpecificOptions contains customized implementation-specific fields and functions.
+// These will be used by the common package to customize the provider-specific behavior for all the
+// implementation-specific fields of the ingress API.
 type ProviderImplementationSpecificOptions struct {
-	ToImplementationSpecificHTTPPathMatch ImplementationSpecificHTTPPathMatchOption
+	ToImplementationSpecificHTTPPathTypeMatch ImplementationSpecificHTTPPathTypeMatchConverter
 }
 
 // InputResources contains all Ingress objects, and Provider specific
--- a/pkg/i2gw/providers/common/converter.go
+++ b/pkg/i2gw/providers/common/converter.go
@@ -30,9 +30,13 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
+// ToHTTPRouteMatchOption is an option to give the providers the possibility to customize
+// HTTPRouteMatch support.
+type ToHTTPRouteMatchOption func(routePath networkingv1.HTTPIngressPath, path *field.Path) (*gatewayv1beta1.HTTPRouteMatch, *field.Error)
+
 // ToGateway converts the received ingresses to i2gw.GatewayResources,
 // without taking into consideration any provider specific logic.
-func ToGateway(ingresses []networkingv1.Ingress) (i2gw.GatewayResources, field.ErrorList) {
+func ToGateway(ingresses []networkingv1.Ingress, toHTTPRouteMatchCustom ToHTTPRouteMatchOption) (i2gw.GatewayResources, field.ErrorList) {
 	aggregator := ingressAggregator{ruleGroups: map[ruleGroupKey]*ingressRuleGroup{}}
 
 	var errs field.ErrorList
@@ -43,7 +47,7 @@ func ToGateway(ingresses []networkingv1.Ingress) (i2gw.GatewayResources, field.E
 		return i2gw.GatewayResources{}, errs
 	}
 
-	routes, gateways, errs := aggregator.toHTTPRoutesAndGateways()
+	routes, gateways, errs := aggregator.toHTTPRoutesAndGateways(toHTTPRouteMatchCustom)
 	if len(errs) > 0 {
 		return i2gw.GatewayResources{}, errs
 	}
@@ -157,7 +161,7 @@ func (a *ingressAggregator) addIngressRule(namespace, name, ingressClass string,
 	rg.rules = append(rg.rules, ingressRule{rule: rule})
 }
 
-func (a *ingressAggregator) toHTTPRoutesAndGateways() ([]gatewayv1beta1.HTTPRoute, []gatewayv1beta1.Gateway, field.ErrorList) {
+func (a *ingressAggregator) toHTTPRoutesAndGateways(toHTTPRouteMatchCustom ToHTTPRouteMatchOption) ([]gatewayv1beta1.HTTPRoute, []gatewayv1beta1.Gateway, field.ErrorList) {
 	var httpRoutes []gatewayv1beta1.HTTPRoute
 	var errors field.ErrorList
 	listenersByNamespacedGateway := map[string][]gatewayv1beta1.Listener{}
@@ -178,7 +182,7 @@ func (a *ingressAggregator) toHTTPRoutesAndGateways() ([]gatewayv1beta1.HTTPRout
 		}
 		gwKey := fmt.Sprintf("%s/%s", rg.namespace, rg.ingressClass)
 		listenersByNamespacedGateway[gwKey] = append(listenersByNamespacedGateway[gwKey], listener)
-		httpRoute, errs := rg.toHTTPRoute()
+		httpRoute, errs := rg.toHTTPRoute(toHTTPRouteMatchCustom)
 		httpRoutes = append(httpRoutes, httpRoute)
 		errors = append(errors, errs...)
 	}
@@ -269,7 +273,7 @@ func (a *ingressAggregator) toHTTPRoutesAndGateways() ([]gatewayv1beta1.HTTPRout
 	return httpRoutes, gateways, errors
 }
 
-func (rg *ingressRuleGroup) toHTTPRoute() (gatewayv1beta1.HTTPRoute, field.ErrorList) {
+func (rg *ingressRuleGroup) toHTTPRoute(toHTTPRouteMatchCustom ToHTTPRouteMatchOption) (gatewayv1beta1.HTTPRoute, field.ErrorList) {
 	pathsByMatchGroup := map[pathMatchKey][]ingressPath{}
 	var errors field.ErrorList
 
@@ -305,7 +309,13 @@ func (rg *ingressRuleGroup) toHTTPRoute() (gatewayv1beta1.HTTPRoute, field.Error
 	for _, paths := range pathsByMatchGroup {
 		path := paths[0]
 		fieldPath := field.NewPath("spec", "rules").Index(path.ruleIdx).Child(path.ruleType).Child("paths").Index(path.pathIdx)
-		match, err := toHTTPRouteMatch(path, fieldPath)
+		var match *gatewayv1beta1.HTTPRouteMatch
+		var err *field.Error
+		if toHTTPRouteMatchCustom != nil {
+			match, err = toHTTPRouteMatchCustom(path.path, fieldPath)
+		} else {
+			match, err = toHTTPRouteMatch(path.path, fieldPath)
+		}
 		if err != nil {
 			errors = append(errors, err)
 			continue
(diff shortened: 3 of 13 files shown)


pkg/i2gw/providers/kong/converter.go
```diff
@@ -61,3 +66,29 @@ func (c *converter) ToGatewayAPI(resources i2gw.InputResources) (i2gw.GatewayRes
...
+		} else {
+			match.Path.Type = &pmPrefix
+		}
+
+	default:
+		return nil, field.Invalid(path.Child("pathType"), routePath.PathType, fmt.Sprintf("unsupported path match type: %s", *routePath.PathType))
+	}
+
+	return match, nil
+}
```
levikobi, line 94: I would prefer to create a new function for the `PathTypeImplementationSpecific` logic and pass that to the common converter `toHTTPRouteMatchOption` somehow instead of duplicating the entire function.  Given that multiple different providers will have different implementations of the `PathTypeImplementationSpecific`.

I know this is going to be a pretty significant change, but this is what I would have liked to see (Though I'm very interested in other opinions):
1. At the common package, create a new struct named `Converter` and attach the `ToGateway` function to him.
2. Each provider is going to construct his own common converter with provider-specific configuration, for example: `PathTypeImplementationSpecific` function
3. Have the `toHTTPRouteMatchOption` function call the configured `PathTypeImplementationSpecific` function in the case of `PathTypeImplementationSpecific`.

```golang
type Converter struct {
    PathTypeImplementationSpecificFunction func() ...
}

func (c *Converter) ToGateway() ...
```

I think this change will minimize code duplication and allow for easier extensibility in the future instead of passing a new parameter through `ToGateway` each time.

Again, I'm very interested in other opinions regarding this one

levikobi: > 1. why do we need a new `Converter` struct ? we can add this field on the `converter` interface and have the providers initialize it when they initialize the provider.

I don't think this is going to be the only time we're going to have to pass new fields to the `converter`. Creating a new struct from the beginning will make things smoother for the following times, IMO.



> 2. [ToGatewayAPI](https://github.com/kubernetes-sigs/ingress2gateway/blob/38c60b4f573a9e7a52eae580745584f6e6c0d3a7/pkg/i2gw/providers/kong/converter.go#L46) is already a method on the `converter` interface so I do not see a point to change the `ToGateway` as well. we can pass the implementation specific filed down the chain, similar to what Mattia has already did.

`ToGatewayAPI` is a provider-specific method. `ToGateway` is the common one, and it is simply a function, not a method. Again, it comes down to not passing down endless parameters down the chain

Please do let me know if it makes more sense now in your opinion @LiorLieberman @mlavacca

mlavacca: I get @levikobi point on having a common structured way to customize such fields that will be helpful in the future as well. I also agree that passing a long list of parameters to a function is something we should try to avoid as much as possible, and since potentially any implementation-specific parameter could be customized, I think it makes sense to introduce a general enough mechanism to do so.

mlavacca: I have implemented an options-based mechanism. This way it will be easy to add implementation-specific functionalities in the future, keeping the maintenance burden and code duplication very low. Let me know what you think @levikobi @LiorLieberman

pkg/i2gw/providers/kong/converter_test.go
```diff
@@ -33,7 +33,9 @@ import (
 
 func Test_ToGateway(t *testing.T) {
 	iPrefix := networkingv1.PathTypePrefix
+	isPathType := networkingv1.PathTypeImplementationSpecific
```
LiorLieberman, line 36: isPathType is confusing as I expect it to be a `bool` while I think you refer to implementationSpecificPathType
maybe `implementationSpecificPathType` or `implSpecificPathType` ?

pkg/i2gw/providers/kong/README.md
```diff
@@ -17,3 +19,11 @@ Current supported annotations:
...
 
 If you are reliant on any annotations not listed above, please open an issue.
+
+## Implementation-specific features
+
+The following implementation-specific features are supported:
+
+- The ingress `ImplementationSpecific` match type is properly converted to
+  - `RegularExpression` HTTPRoute match type when the path has the prefix `/~`.
+  - `PathPrefix` HTTPRoute match type when there is no prefix `/~`.
```
LiorLieberman, line 29: can we also add a test for Kong `PathPrefix` Implementation-specific ?

pkg/i2gw/providers/ingressnginx/converter_test.go
LiorLieberman, line 1: Thanks! 😊

pkg/i2gw/providers/common/converter.go
```diff
@@ -348,20 +350,22 @@ func getPathMatchKey(ip ingressPath) pathMatchKey {
...
+	switch *routePath.PathType {
 	case networkingv1.PathTypePrefix:
 		match.Path.Type = &pmPrefix
 	case networkingv1.PathTypeExact:
 		match.Path.Type = &pmExact
-	default:
-		return nil, field.Invalid(path.Child("pathType"), ip.path.PathType, fmt.Sprintf("unsupported path match type: %s", *ip.path.PathType))
+	case networkingv1.PathTypeImplementationSpecific:
+		if implementationSpecificPath != nil {
+			implementationSpecificPath(match.Path)
```
LiorLieberman, line 365: this would benefit from a comment. 
Also maybe `applyImplementationSpecificPath` or `toImplementationSpecificPath`?

pkg/i2gw/providers/common/converter.go
```diff
@@ -305,7 +305,9 @@ func (rg *ingressRuleGroup) toHTTPRoute() (gatewayv1beta1.HTTPRoute, field.Error
 	for _, paths := range pathsByMatchGroup {
 		path := paths[0]
 		fieldPath := field.NewPath("spec", "rules").Index(path.ruleIdx).Child(path.ruleType).Child("paths").Index(path.pathIdx)
-		match, err := toHTTPRouteMatch(path, fieldPath)
+		var match *gatewayv1beta1.HTTPRouteMatch
+		var err *field.Error
```
LiorLieberman, line 309: any particular reason for changing to two vars instead of `:=`?

pkg/i2gw/provider.go
```diff
@@ -73,6 +73,16 @@ type ResourceConverter interface {
...
 
+// HTTPPathMatchOption is an option to customize the ingress implementationSpecific
+// match type conversion.
+type HTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
+
+// ImplementationSpecificOptions contains all the pointers to implementation-specific
+// customization functions.
+type ImplementationSpecificOptions struct {
+	HTTPPathmatch HTTPPathMatchOption
+}
```
LiorLieberman, line 84: Can you explain the idea behind the names?

What would be the use of HttpMatchOption? only for implementation specific funcs or there is a vision it could be used for other things? 

Similarly for `ImplementationSpecificOptions` - is this intended to hold various different things/options that the implementation would need during the conversion? If yes, I would prefer be consistent and calling it "provider" and not implementation. Do shout if I got it wrong.

LiorLieberman: Another question, @levikobi feel free to jump here as well. 

Why do we need another struct to hold provider specific options if we have `ProviderConf` cant we just initialize providerConf with the relevant `HTTPPathMatchOption` and pass `c.conf` to [ToGateway](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/kong/converter.go#L50) here instead of the new `ImplementationSpecificOptions` struct?

LiorLieberman (commented):
Thanks Mattia! thats awesome! 

I like this approach. Left a few comments.

pkg/i2gw/provider.go
mlavacca: > What would be the use of HttpMatchOption? only for implementation specific funcs or there is a vision it could be used for other things?

it's intended to be used only on the `implementationSpecific` path type. Probably the name is a bit unfortunate, I'll rename to reflect that 👍 .
All the handlers in `ImplementationSpecificOptions` should be intended to customize the `implementationSpecific` fields of the API. I'll add a comment about that.

> is this intended to hold various different things/options that the implementation would need during the conversion? If yes, I would prefer be consistent and calling it "provider" and not implementation

it's indented to hold all the handlers to customize the `Implementation-specific` fields of the APIs. `Implementation-specific` is the term commonly used in the Gateway API and I would rely on it. Maybe we can rename it `ProviderImplementationSpecificOptions`. WDYT?

> Why do we need another struct to hold provider specific options if we have ProviderConf cant we just initialize providerConf with the relevant HTTPPathMatchOption and pass c.conf to [ToGateway](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/kong/converter.go#L50) here instead of the new ImplementationSpecificOptions struct?

The problem with the ProviderConf is that it is instantiated outside the provider and given to the provider. As it is designed at the moment, it's a good way to pass parameters from outside to the provider (such as the client). The other way around results to be a bit clunky at the moment. The implementation-specific function handlers are intended to be passed from the provider to the common package, and relying on the `ProviderConf` to do so is a bit odd as it is designed today in my opinion.

pkg/i2gw/providers/common/converter.go
mlavacca: nope, probably just a leftover from a previous change, I'll revert it.

pkg/i2gw/providers/common/converter.go
mlavacca: comment added and parameter renamed

pkg/i2gw/providers/kong/README.md
mlavacca: test added

pkg/i2gw/providers/kong/converter_test.go
mlavacca: Yeah, I agree 😄 Renamed

pkg/i2gw/provider.go
LiorLieberman: > it's indented to hold all the handlers to customize the Implementation-specific fields of the APIs. Implementation-specific is the term commonly used in the Gateway API and I would rely on it. Maybe we can rename it ProviderImplementationSpecificOptions. WDYT?

so when you say `implementationSpecific` fields of the API you basically mean "whatever custom thing the implementation will need during the conversion", right? 
If so the project convention is "Provider" instead of implementation. I agree with you we need to be consistent with Gateway API, but this requires further discussion. We better be consistent within the project now, and if we decide to change it we should do the wording change in one PR.
Provider == implementation in this case so I think we need something else than `ProviderImplementationSpecificOptions`. (I think we can just add the `implementationSpecific` function to the ProviderConf`

 

> The problem with the ProviderConf is that it is instantiated outside the provider and given to the provider. As it is designed at the moment, it's a good way to pass parameters from outside to the provider (such as the client). The other way around results to be a bit clunky at the moment. The implementation-specific function handlers are intended to be passed from the provider to the common package, and relying on the ProviderConf to do so is a bit odd as it is designed today in my opinion.

cant you edit the conf [here](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/kong/kong.go#L39C30-L45) and add the `implementationSpecificHTTPPathMatch` function field. Then call new provider with the new updated conf.

I think the idea behind initializing the provider conf in the common package is to provide implementation with the default client, but they can override it. I think `implementationSpecificHTTPPathMatch` should be a common func field to every implementation hence I think it should be in `ProviderConf`

pkg/i2gw/provider.go
```diff
@@ -42,7 +42,8 @@ type ProviderConstructor func(conf *ProviderConf) Provider
 // ProviderConf contains all the configuration required for every concrete
 // Provider implementation.
 type ProviderConf struct {
-	Client client.Client
+	Client  client.Client
+	Options ProviderImplementationSpecificOptions
```
levikobi, line 46: Why do we need the `Options` field here?
From what I've seen, each provider constructs its own `ProviderImplementationSpecificOptions` and passes it to `ToGateway` (which is good, in my opinion).
So I don't see the purpose of having it here too

levikobi (commented):
Thanks for making this change @mlavacca! I think it will make things easier as we advance.

pkg/i2gw/provider.go
mlavacca: 😕 we don't indeed. I added such a field when thinking about Lior's comments and accidentally committed. Just removed, good catch

pkg/i2gw/provider.go
mlavacca: > so when you say implementationSpecific fields of the API you basically mean "whatever custom thing the implementation will need during the conversion", right?

when I say implementation-specific, I refer to the Gateway API support levels [terminology](https://gateway-api.sigs.k8s.io/concepts/conformance/#2-support-levels) (core, extended, implementation-specific). Since core and extended fields can be converted in the common package, as they are implementation-independent and part of the API, all the implementation-specific features, such as the regex path type for HTTPRoute, require a custom modifier to be set. For this reason, I'd keep the `ImplementationSpecific` part of the options, as they refer to the actual terminology. I set `ProviderImplementationSpecificOptions` as these are options related to `implementationSpecific` settings of the provider.

> cant you edit the conf [here](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/kong/kong.go#L39C30-L45) and add the implementationSpecificHTTPPathMatch function field ? Then call new provider with the new updated conf.

This is the clunkiness I was talking about. I don't like the approach of putting everything into the config before creating the provider, but It's a personal opinion, it works either way.

mlavacca: And what if we put such handlers into the converter, along with the `featureParsers`? It looks like the very best place to put such functions: we do not need to change the conf passed from common and can keep such field under the provider's control in a structured way. WDYT?

LiorLieberman: Nice catch! I agree that `featureParser` is the best way to put it. with this featureParser implemented I will suggest as well to change https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/common/converter.go#L364 to print different warning message like "implementationSpecific path type is not supported in generic translation, however providers may added a custom support to translate it. Check your HttpRoute/s to see they reflect the implementationSpecific path type"

Dont take this exact wording, I was just mumbling without thinking too much :)

levikobi: > And what if we put such handlers into the converter, along with the `featureParsers`? It looks like the very best place to put such functions: we do not need to change the conf passed from common and can keep such field under the provider's control in a structured way. WDYT?

+1
This is exactly what I wanted

mlavacca: 👍 options moved to the converter

pkg/i2gw/provider.go
```diff
@@ -73,6 +73,17 @@ type ResourceConverter interface {
...
 
+// ImplementationSpecificHTTPPathMatchOption is an option to customize the ingress implementationSpecific
+// match type conversion.
+type ImplementationSpecificHTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
+
+// ProviderImplementationSpecificOptions contains all the pointers to implementation-specific
+// customization functions. Such functions will be called by the common package to customize
+// the provider-specific behavior for all the implementation-specific fields of the API.
+type ProviderImplementationSpecificOptions struct {
+	ToImplementationSpecificHTTPPathMatch ImplementationSpecificHTTPPathMatchOption
```
LiorLieberman, line 84:
```suggestion
	ToImplementationSpecificHTTPPathTypeMatch ImplementationSpecificHTTPPathTypeMatchConverter
```

pkg/i2gw/provider.go
```diff
@@ -73,6 +73,17 @@ type ResourceConverter interface {
 	ToGatewayAPI(resources InputResources) (GatewayResources, field.ErrorList)
 }
 
+// ImplementationSpecificHTTPPathMatchOption is an option to customize the ingress implementationSpecific
+// match type conversion.
+type ImplementationSpecificHTTPPathMatchOption func(*gatewayv1beta1.HTTPPathMatch)
+
+// ProviderImplementationSpecificOptions contains all the pointers to implementation-specific
+// customization functions. Such functions will be called by the common package to customize
+// the provider-specific behavior for all the implementation-specific fields of the API.
```
LiorLieberman, line 82:
```suggestion
// ImplementationSpecificHTTPPathTypeMatchConverter is an option to customize the ingress implementationSpecific pathType match conversion.
type ImplementationSpecificHTTPPathTypeMatchConverter func(*gatewayv1beta1.HTTPPathMatch)

// ProviderImplementationSpecificOptions contains customized implementation-specific fields and functions. These will be used by the common package to customize the provider-specific behavior for all the implementation-specific fields of the ingress API.
```

pkg/i2gw/providers/common/converter.go
```diff
@@ -348,20 +348,25 @@ func getPathMatchKey(ip ingressPath) pathMatchKey {
 	return pathMatchKey(fmt.Sprintf("%s/%s", pathType, ip.path.Path))
 }
 
-func toHTTPRouteMatch(ip ingressPath, path *field.Path) (*gatewayv1beta1.HTTPRouteMatch, *field.Error) {
+func toHTTPRouteMatch(routePath networkingv1.HTTPIngressPath, path *field.Path, toImplementationSpecificPathMatch i2gw.ImplementationSpecificHTTPPathMatchOption) (*gatewayv1beta1.HTTPRouteMatch, *field.Error) {
```
LiorLieberman, line 351:
```suggestion
func toHTTPRouteMatch(routePath networkingv1.HTTPIngressPath, path *field.Path, toImplementationSpecificPathMatch i2gw.ImplementationSpecificHTTPPathTypeMatchConverter) (*gatewayv1beta1.HTTPRouteMatch, *field.Error) {
```

pkg/i2gw/provider.go
LiorLieberman: `ToImplementationSpecificHTTPPathMatch` --> `ToImplementationSpecificHTTPPathTypeMatch` across the whole PR

pkg/i2gw/providers/kong/converter.go
```diff
@@ -26,7 +26,8 @@ import (
 type converter struct {
 	conf *i2gw.ProviderConf
 
-	featureParsers []i2gw.FeatureParser
+	featureParsers                []i2gw.FeatureParser
+	ImplementationSpecificOptions i2gw.ProviderImplementationSpecificOptions
```
LiorLieberman, line 30: correct me if I am wrong but we dont really need this in this struct, right?
we can just pass the function on line 54

pkg/i2gw/providers/kong/implementation_specific.go
LiorLieberman, line 1: can rename the file to `implementation_specific_pathtype.go`?

LiorLieberman (commented):
Thanks Mattia, thats awesome!
mostly lgtm, just a few last nits

pkg/i2gw/providers/kong/implementation_specific.go
levikobi: If every implementation-specific functionality has its own file, we'll end up with a large list of nearly empty files. I tend to put it all together in the same file, although I don't have strong feelings about it

pkg/i2gw/providers/common/converter.go
```diff
@@ -348,20 +348,25 @@ func getPathMatchKey(ip ingressPath) pathMatchKey {
...
-		return nil, field.Invalid(path.Child("pathType"), ip.path.PathType, fmt.Sprintf("unsupported path match type: %s", *ip.path.PathType))
+	// In case the path type is ImplementationSpecific, the path value and type
+	// will be set by the provider-specific customization function. If such function
+	// is not given by the provider, an error is returned.
+	case networkingv1.PathTypeImplementationSpecific:
+		if toImplementationSpecificPathMatch != nil {
+			toImplementationSpecificPathMatch(match.Path)
+		} else {
+			return nil, field.Invalid(path.Child("pathType"), routePath.PathType, "implementationSpecific path type is not supported in generic translation, and your provider does not provide custom support to translate it")
+		}
```
levikobi, line 369: Shouldn't we keep the default case? Just to keep it bulletproof

pkg/i2gw/providers/kong/converter.go
levikobi: You're right, although we could say the same thing about `featureParsers` field. As far as I can tell, we only use them on the `converter.ToGatewayAPI` function.

We can either initialize both of them in the constructor or on the `converter.ToGatewayAPI` function. I personally like the constructor option better because it makes the other function easier on the eyes.

pkg/i2gw/providers/kong/implementation_specific.go
mlavacca: > If every implementation-specific functionality has its own file, we'll end up with a large list of nearly empty files. I tend to put it all together in the same file, although I don't have strong feelings about it

+1

pkg/i2gw/providers/kong/converter.go
LiorLieberman: hmmm you convinced me :) resolving it

pkg/i2gw/providers/kong/implementation_specific.go
LiorLieberman: good catch. not sure what @levikobi meant one impl_specific for all the implementations OR just to put the function in the kong/conveter.go file, and every implementation will do it on their converter files. I vote for the latter

mlavacca: I'd prefer having a single purpose file for each provider that contains all the implementation-specific options honestly. I think we should keep the converter file as lean as possible.

pkg/i2gw/provider.go
mlavacca: renamed 👍

pkg/i2gw/providers/kong/implementation_specific.go
LiorLieberman: do you anticipate more things to go in their other than implemetationSpecificPathType though?

mlavacca: At the moment no, but since we are adopting a single-purpose file structure, where every file has its own purpose (resource-reader, converter, kong.go/nginx.go, a file for each feature), adding this function to the provider seems to be breaking this approach.

LiorLieberman: I can feel like only implementation specific path type is going to be uses in this purpose, all others should go as featureParsers and have their own file. 

That said i am find with this if you think thats better.

PROVIDER.md
```diff
@@ -83,6 +87,9 @@ func newConverter(conf *i2gw.ProviderConf) *converter {
 		featureParsers: []i2gw.FeatureParser{
 			// The list of feature parsers comes here.
 		},
+		implementationSpecificOptions: i2gw.ProviderImplementationSpecificOptions{
+			// The list of the implementationSpecific options comes here.
```
LiorLieberman, line 91:
```suggestion
			// The list of the implementationSpecific ingress fields options comes here.
```

pkg/i2gw/providers/kong/converter_test.go
```diff
@@ -33,7 +33,9 @@ import (
 
 func Test_ToGateway(t *testing.T) {
 	iPrefix := networkingv1.PathTypePrefix
+	ImplSpecPathType := networkingv1.PathTypeImplementationSpecific
```
LiorLieberman, line 36:
```suggestion
	ImplSpecificPathType := networkingv1.PathTypeImplementationSpecific
```

LiorLieberman (approved):
Thanks, approving with two small nits. 

Will leave @levikobi for lgtm

levikobi (approved):
Great contribution as always. Thanks @mlavacca!
/lgtm

PR #99, opened 2023-11-26
Add init code for the istio provider

dpasiukevich:
**What type of PR is this?**
/kind feature

Adds init code for the istio provider. It would convert Istio API resources (Gateway, VirtualService, DestinationRule etc) into K8S Gateway API

This PR:
1. Reads istio Gateway & VirtualServices
2. Removes `customResources` arg from the Provider interface. Currently is passed to the provider as a `interface{}` (nil value & nil type). Provider code currently cannot modify this struct (because cannot cast to any other type and cannot overwrite the value). 
The PR simplifies interface and removes the need to pass `customResources` as `interface{}` which is also shared across different providers.
Now the provider if wants to handle custom resources, must itself parse them and store in any desired format locally until `ToGatewayAPI` is called on it.
3. Updates `go.mod` to import latest IstioAPI.

cc @howardjohn 

```release-note
NONE
```


LiorLieberman:
Thats great @dpasiukevich ! 
I will book some time today to review this. 

In the meantime, can you edit the release-notes section to reflect the changes (that will then be included in our release notes) 

This is a user facing change assuming now users will be able to invoke the tool with istio provider.

/cc @levikobi @mlavacca

dpasiukevich:
The current PR is just a basic foundation to put the actual code later on. 
I planned to add the release notes to notify customers when the converter would be ready at least on the basic level.

pkg/i2gw/providers/istio/resource_reader.go
```diff
@@ -0,0 +1,147 @@
...
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), &vs); err != nil {
+				return nil, fmt.Errorf("failed to parse istio virtual service object: %w", err)
+			}
+
+			res.VirtualServices[types.NamespacedName{
+				Namespace: vs.Namespace,
+				Name:      vs.Name,
+			}] = &vs
+		default:
+			return nil, fmt.Errorf("unsupported resource kind: %v", objKind)
```
LiorLieberman, line 89: Is there a way where we can warn the user and skip the resource instead of returning an error? I assume some users will have other kinds inside the same file

pkg/i2gw/providers/istio/istio.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type Provider struct {
+	storage   storage
+	reader    reader
+	converter converter
+}
+
+// NewProvider returns the istio implementation of i2gw.Provider.
+func NewProvider(config *i2gw.ProviderConf) i2gw.Provider {
```
LiorLieberman, line 43:
```suggestion
func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
```

nit: for consistency with other implementations

go.mod
```diff
@@ -3,68 +3,71 @@ module github.com/kubernetes-sigs/ingress2gateway
...
+	golang.org/x/sync v0.5.0 // indirect
+	golang.org/x/sys v0.14.0 // indirect
+	golang.org/x/term v0.14.0 // indirect
+	golang.org/x/text v0.14.0 // indirect
+	golang.org/x/time v0.4.0 // indirect
+	google.golang.org/appengine v1.6.8 // indirect
+	google.golang.org/genproto v0.0.0-20231120223509-83a465c0220f // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20231120223509-83a465c0220f // indirect
+	google.golang.org/protobuf v1.31.0 // indirect
+	gopkg.in/evanphx/json-patch.v5 v5.7.0 // indirect
```
LiorLieberman, line 62: I wonder where this comes from as we already have above 
```
	github.com/evanphx/json-patch v5.7.0+incompatible // indirect
	github.com/evanphx/json-patch/v5 v5.7.0 // indirect
```

pkg/i2gw/providers/istio/converter.go
```diff
@@ -0,0 +1,59 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:       gw.Namespace,
+				Name:            gw.Name,
+				Labels:          gw.Labels,
+				Annotations:     gw.Annotations,
+				OwnerReferences: gw.OwnerReferences,
+				Finalizers:      gw.Finalizers,
+			},
+		}
+	}
```
LiorLieberman, line 56: is this intentional to only read Gateways from storage? and only fill in metadata?

dpasiukevich: > is this intentional to only read Gateways from storage?
Yes, the storage would have all istio CRD instances (loaded from cluster or from file).
We don't convert K8S Ingress API in this provider.

> and only fill in metadata?
This converter is just the basis for the following converter logic.

go.mod
dpasiukevich: It may be from `istio.io/api v1.20.0`. It's needed for the new provider.

pkg/i2gw/providers/istio/converter.go
```diff
@@ -0,0 +1,59 @@
...
+
+package istio
+
+import (
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/common"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	k8s "sigs.k8s.io/gateway-api/apis/v1beta1"
```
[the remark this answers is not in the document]
dpasiukevich: done

pkg/i2gw/providers/istio/istio.go
dpasiukevich: done

pkg/i2gw/providers/istio/resource_reader.go
dpasiukevich: added log message for both APIVersion check and Kind

dpasiukevich (commented):
@LiorLieberman updated in regards for the PR comments.

As for this PR -- I just want to have a proper basis and agreement for the conversion logic (imports, proper interface for the Provider).

If everything was done in one PR (this would've been a gargantuan PR). As I cannot reuse any existing conversion logic in the repo (it's for Ingress, while the provider is for Istio).

I'd prefer quicker&smaller iterations.

If needed I can remove the provider from registration for now.

pkg/i2gw/providers/istio/types.go
```diff
@@ -0,0 +1,42 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	APIVersion         = "networking.istio.io/v1beta1"
+	GatewayKind        = "Gateway"
+	VirtualServiceKind = "VirtualService"
+)
+
+type gateway struct {
```
dpasiukevich, line 30: @howardjohn would this be an ok representation of the object to parse into from the `unstructured.Unstructured`? Or is there something similar in the istio API (adapted Istio resources for K8S)? 

I didn't find any, asking just to double check.

pkg/i2gw/provider.go
```diff
@@ -55,13 +55,11 @@ type Provider interface {
 
 type CustomResourceReader interface {
 
-	// ReadResourcesFromCluster reads custom resources associated with
-	// the underlying Provider implementation from the kubernetes cluster.
-	ReadResourcesFromCluster(ctx context.Context, customResources interface{}) error
+	// ReadResourcesFromCluster reads custom resources from the kubernetes cluster.
```
LiorLieberman, line 58: why removing `associated with the underlying Provider implementation from the kubernetes cluster` ?

pkg/i2gw/providers/istio/README.md
```diff
@@ -0,0 +1,5 @@
+# Istio Provider
+
+## WIP
+
+Provider converts networking.istio.io/v1beta1 resources: Gateway, Virtual Service, Destination Rule
```
LiorLieberman, line 5:
```suggestion
Follow #100 for updates on the progress 
```

pkg/i2gw/providers/istio/istio.go
```diff
@@ -0,0 +1,88 @@
...
+		converter: newConverter(),
+	}
+}
+
+// ToGatewayAPI converts stored Istio API entities to i2gw.GatewayResources
+// K8S Ingress resources are not needed, only Istio-based are converted
+// TODO(#100) - add istio conversion logic
+func (p *Provider) ToGatewayAPI(_ i2gw.InputResources) (i2gw.GatewayResources, field.ErrorList) {
+	p.converter.convert(p.storage)
+	return i2gw.GatewayResources{}, field.ErrorList{field.Forbidden(field.NewPath(""), "conversion is WIP")}
```
LiorLieberman, line 56:
```suggestion
	return i2gw.GatewayResources{}, field.ErrorList{field.Forbidden(field.NewPath(""), "istio conversion is WIP")}
	p.converter.convert(p.storage)
```

pkg/i2gw/providers/istio/converter.go
```diff
@@ -0,0 +1,60 @@
...
+				Name:            gw.Name,
+				Labels:          gw.Labels,
+				Annotations:     gw.Annotations,
+				OwnerReferences: gw.OwnerReferences,
+				Finalizers:      gw.Finalizers,
+			},
+		}
+	}
+
+	return gatewayResources, field.ErrorList{field.Forbidden(field.NewPath(""), "conversion is WIP")}
```
LiorLieberman, line 59: since you added this error in `ToGatewayAPI()`, it is not needed here.
```suggestion
	return gatewayResources, nil
```

pkg/i2gw/providers/istio/converter.go
```diff
@@ -0,0 +1,60 @@
...
+
+type converter struct{}
+
+func newConverter() converter {
+	return converter{}
+}
+
+// TODO(#100): istio conversion support is WIP
+func (c *converter) convert(storage storage) (i2gw.GatewayResources, field.ErrorList) {
+	gatewayResources := i2gw.GatewayResources{
```
LiorLieberman, line 36:
```suggestion
func (c *converter) convert(storage storage) (i2gw.GatewayResources, field.ErrorList) {
        // TODO(#100): This logic only do a simple metadata conversion. Need to implement istio conversion logic. 
	gatewayResources := i2gw.GatewayResources{
```

LiorLieberman (commented):
Thanks! This looks good to me. Left a few more nits. 

As this PR is significant, I am waiting for another couple of eyes to look at it.

pkg/i2gw/providers/istio/istio.go
dpasiukevich: linter would reject this with "unreachable code"

pkg/i2gw/providers/istio/converter.go
dpasiukevich: linter would reject this with `return value field.ErrorList is always nil`

and I'd implement the method in the next PRs

dpasiukevich:
@levikobi could I ask for the review? @LiorLieberman already lgtm'd but per https://github.com/kubernetes-sigs/ingress2gateway/pull/99#pullrequestreview-1750942804 asking for an extra review.

I could start the work on full istio provider support in this repo after this PR is merged.

pkg/i2gw/providers/istio/converter.go
LiorLieberman: curios what linter are you using?

dpasiukevich: `./hack/verify-golint.sh` which runs `golangci-lint:v1.55.2`

```
 % ./hack/verify-golint.sh
pkg/i2gw/providers/istio/converter.go:34:70: (*converter).convert - result 1 (k8s.io/apimachinery/pkg/util/validation/field.ErrorList) is always nil (unparam)
func (c *converter) convert(storage storage) (i2gw.GatewayResources, field.ErrorList) {
```

.gitignore
```diff
@@ -26,3 +26,6 @@ go.work
 
 # GolangCI-Lint cache
 /cache
+
+# Intellij
+.idea/
```
mlavacca, line 31: I'd prefer not to have a setup-based entry in our `.gitignore`. You could add `.idea/` to your global `.gitignore`, so that it never gets added to the staging area.

hack/verify-golint.sh
```diff
@@ -18,7 +18,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-readonly VERSION="v1.46.2"
+readonly VERSION="v1.55.2"
```
mlavacca, line 21: 👍

mlavacca (commented):
Hi @dpasiukevich, thanks for this PR, it's great to have another provider onboard!

I have a comment about the storage approach introduced in this PR. I think it goes against the current approach we have at the moment, as we are treating the resources differently: the networkingv1 resources are fetched and returned to the caller, while the provider -specific resources are stored internally in the provider. I'd honestly prefer using the same approach for both the categories, as meshing the two approaches implies losing clarity and structure in the code in my opinion. 

I have [this PR](https://github.com/kubernetes-sigs/ingress2gateway/pull/86) that addresses the same problem by using an approach (the CRDs are gathered, mapped per GVK, and returned to the caller), coherent with the approach we currently have.

I think we should decide which design to keep and go coherently (store everything - return everything). 

@LiorLieberman @levikobi opinions about it?

dpasiukevich:
My suggestion is to keep the providers individual and separate. Each provider should implement 3 methods:
1. read from file
2. read from cluster
3. to gateway api

Each provider knows what resources it needs, what types, how to read and how to convert them. This is a modular approach which supports any providers without any effort.

While with the sharing and passing around (`networking1.Ingress` + map[GVK]interface{}) has some donwsides:
1. some future providers (e.g. as Istio) they don't need `networking1.Ingress` at all. They only need their specific CRDs (as Istio).
2. sharing the same `map[GVK]interface{}`  around all providers lead to non-deterministic results when different providers would modify input resources.
3. removal the need of `interface{}`, with modular providers each provider knows how to read and what types it needs instead of passing interface{} around.

For me both options are fine, although I genuinely see having modular independent provider as a better way -- this better supports any type of provider.
1. If this approach is fine, I can move `ReadIngress` code to common, and update the interface method to be `ToGatewayAPI()`
2. If sharing `map[schema.GroupVersionKind]interface{}` is needed, then I will only add changes after https://github.com/kubernetes-sigs/ingress2gateway/pull/86 submitted to avoid merge conflicts.

levikobi (commented):
Regarding the custom resources and storage discussion @mlavacca @dpasiukevich, I have to say the design of the interfaces was supposed to support something similar to what @mlavacca did in #86, although not precisely (my fault for not expressing myself well enough). What I had in mind is that each provider will have their own struct containing maps of CRDs (a map of `VirtualService`, a different map for another CRD, etc...) - Make the casting once at the beginning of the conversion, and that's it. But I have no particular problem with @mlavacca approach either.

We should align all providers to only one of these options.
I prefer the existing way because it gives the caller all the resources (Ingress + CRDs), and he can do whatever he wants with them.

> sharing the same map[GVK]interface{} around all providers leads to non-deterministic results when different providers would modify input resources.

I agree. Each provider should receive a dedicated instance of `InputResources`, in my opinion.
Let's keep the separation between the providers as much as possible.

LiorLieberman:
@dpasiukevich @mlavacca @levikobi Lets leave behind for a second what was initially designed. We are still in a position where we can be agile enough to change things. 
I agree with some of the benefits @dpasiukevich mentioned in his comment. What I miss is what are the benefits of sticking with the current design over the suggested one?

Another important question - which design is better for executing on multiple providers? we are going to have many real-world cases like this (e.g istio+gke). Also what are the benefits of sharing resources vs not sharing.

I can start a google doc each of you can put the suggested design and we can go from there.

WDYT? @dpasiukevich @levikobi

mlavacca:
> @dpasiukevich @mlavacca @levikobi Lets leave behind for a second what was initially designed. We are still in a position where we can be agile enough to change things. I agree with some of the benefits @dpasiukevich mentioned in his comment. What I miss is what are the benefits of sticking with the current design over the suggested one?
> 
> Another important question - which design is better for executing on multiple providers? we are going to have many real-world cases like this (e.g istio+gke). Also what are the benefits of sharing resources vs not sharing.

Something not clear to me is: what's the use-case to use more than one provider at a time? For example, the input is one ingress and the output is n Gateways + n routes, for n providers. When should it be useful?

> 
> I can start a google doc each of you can put the suggested design and we can go from there.
> 
> WDYT? @dpasiukevich @levikobi

The main discriminator that I see is keeping isolation between providers on the CRDs side of things. By using this approach, isolation is taken by design, while with the other approach, we should reach the same by giving each provider its own structure to fill.
I have no strong opinion about the approach to pick, both suit our needs.

dpasiukevich:
1.
I'd say main thing with the `interfaces{}` would be casting them. All providers that use CRDs would be polluted with similar logic of casting `interface{}` and handling errors.

Only provider itself has knowledge what types of CRDs it needs. So why leak it to the external package in form of `interface{}` just to pass back to the same provider in `ToGateway`.

2.
> what's the use-case to use more than one provider at a time?
I somehow expect that 99.9% cases the user would run just one provider. At least I don't see a scenario where >1 provider run would be useful.

3.
With "modular" approach each provider would fetch ingress resources by itself (although maybe using "common" library). Given that ingress resources would be needed by majority of providers now (except for istio), this would make multiple same calls to get ingress from different providers (for multiprovider run). So at least sharing the passing of ingress could make sense.

LiorLieberman:
Summarizing our offline discussion from today: 

1. We are going to purse with the approach introduced here. Essentially every provider would be isolated have its storage.
2. We are going to follow up on the need whether Ingresses should also move to be fetched by the provider (perhaps every provider will query the ingresses with the relevant ingress classes it needs) 

3. @dpasiukevich  will change the function names on the interface to reflect what it actually does.

pkg/i2gw/ingress2gateway.go
```diff
@@ -64,15 +64,15 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 			return nil, nil, fmt.Errorf("failed to read ingresses from file: %w", err)
 		}
 		resources.Ingresses = ingresses.Items
-		if err = readProviderResourcesFromFile(ctx, providerByName, &resources, inputFile); err != nil {
+		if err = readProviderResourcesFromFile(ctx, providerByName, inputFile); err != nil {
```
mlavacca, line 67: I think we need to change `Read` to `Fetch` also for this one and the `FromCluster` one

LiorLieberman (commented):
Looks good to me! thanks. 
/approve 

@mlavacca and @levikobi for LGTM

pkg/i2gw/providers/common/converter.go
```diff
@@ -136,7 +136,7 @@ func (a *ingressAggregator) addIngress(ingress networkingv1.Ingress) field.Error
 			backend:      *ingress.Spec.DefaultBackend,
 		})
 	}
-	return nil
+	return
```
mlavacca, line 139: this return can be removed at all

pkg/i2gw/provider.go
```diff
@@ -55,13 +55,13 @@ type Provider interface {
 
 type CustomResourceReader interface {
```
mlavacca, line 56: Maybe we should rename this to `CustomResourceFetcher`? Just thinking loud, I am not sure about it.

pkg/i2gw/providers/istio/resource_reader.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read virtual services: %w", err)
+	}
+
+	res.VirtualServices = virtualServices
+
+	return &res, nil
+}
+
+func (r *reader) readUnstructuredObjects(objects []*unstructured.Unstructured) (*storage, error) {
```
mlavacca, line 60: What's the purpose of this function? As far as I see, it is unused

dpasiukevich: This func will be used to fetch all required Istio CRDs. Now not used as this PR just prepared the code to make actual Istio conversion. But this will be updated in future PRs.

pkg/i2gw/provider.go
dpasiukevich: Good idea, thanks for catching that! Updated.

fyi @LiorLieberman

pkg/i2gw/ingress2gateway.go
dpasiukevich: Done.

pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
 )
 
-// converter implements the i2gw.CustomResourceReader interface.
+// converter implements the i2gw.CustomResourceFetcher interface.
 type resourceReader struct {
```
mlavacca, line 26:
```suggestion
type resourceFetcher struct {
```

pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -34,12 +34,12 @@ func newResourceReader(conf *i2gw.ProviderConf) *resourceReader {
 	}
 }
 
-func (r *resourceReader) ReadResourcesFromCluster(ctx context.Context, customResources interface{}) error {
+func (r *resourceReader) FetchResourcesFromCluster(_ context.Context) error {
```
mlavacca, line 37:
```suggestion
func (r *resourceFetcher) FetchResourcesFromCluster(_ context.Context) error {
```

pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -34,12 +34,12 @@ func newResourceReader(conf *i2gw.ProviderConf) *resourceReader {
...
 	// ingress-nginx does not have any CRDs.
 	return nil
 }
 
-func (r *resourceReader) ReadResourcesFromFiles(ctx context.Context, customResources interface{}, filename string) error {
+func (r *resourceReader) FetchResourcesFromFile(_ context.Context, _ string) error {
```
mlavacca, line 42:
```suggestion
func (r *resourceFetcher) FetchResourcesFromFile(_ context.Context, _ string) error {
```

pkg/i2gw/providers/ingressnginx/resource_reader.go
mlavacca: We should rename all the `CustomResourceReader`s to `CustomResourceFetcher`. The same goes for Kong and Nginx and for all the `resource_reader.go` files (should be `resource_fetcher.go`)

dpasiukevich: Yep, sorry for not catching this first time. Now I've renamed all references I believe.

mlavacca (approved):
Thanks for it @dpasiukevich, lgtm!

Will defer to @levikobi for the final lgtm stamp.

go.mod
howardjohn: Its from sigs.k8s.io/kustomize/api which was updated.

Istio API has no new deps:

```
        github.com/golang/protobuf v1.5.3
        google.golang.org/genproto/googleapis/api v0.0.0-20230920204549-e6e6cdab5c13
        google.golang.org/genproto/googleapis/rpc v0.0.0-20231009173412-8bfb1ae86b6c
        google.golang.org/grpc v1.58.3
        google.golang.org/protobuf v1.31.0
        k8s.io/api v0.28.0
        k8s.io/apimachinery v0.28.0
```

just upgraded to 0.28

pkg/i2gw/providers/istio/types.go
howardjohn: istio/client-go repo already has these that can be used

dpasiukevich: Cool, exactly what's needed. Thanks!

pkg/i2gw/providers/istio/istio.go
```diff
@@ -0,0 +1,88 @@
...
+// The ProviderName returned to the provider's registry.
+const ProviderName = "istio"
+
+func init() {
+	i2gw.ProviderConstructorByName[ProviderName] = NewProvider
+}
+
+type Provider struct {
+	storage   storage
+	reader    fetcher
```
LiorLieberman, line 38:
```suggestion
	fetcher    fetcher
```

pkg/i2gw/providers/istio/istio.go
```diff
@@ -0,0 +1,88 @@
...
+	converter converter
+}
+
+// NewProvider returns the istio implementation of i2gw.Provider.
+func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
+	return &Provider{
+		storage:   newResourcesStorage(),
+		reader:    newResourceFetcher(conf.Client),
```
LiorLieberman, line 46:
```suggestion
		fetcher:    newResourceFetcher(conf.Client),
```

mlavacca (approved):

pkg/i2gw/providers/istio/istio.go
dpasiukevich: returned naming back to readers

pkg/i2gw/providers/istio/types.go
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package istio
+
+const (
+	APIVersion         = "networking.istio.io/v1beta1"
```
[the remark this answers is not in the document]
dpasiukevich: I'd keep as is to indicate the only currently supported API version is `networking.istio.io/v1beta1`  for istio converter

PR #104, opened 2023-12-12
fix: avoid duplicate `BackendRefs`

### LINKED ISSUE #102: Duplicate `BackendRefs` in `HTTPRoute` rules
**What happened**:

When converting Ingresses with the same rules into Gateway API objects, `HTTPRoute` Matches are properly grouped, while rule `BackendRefs` are duplicated.

the conversion output of the following yaml file: 
```yaml
---
kind: Ingress
metadata:
  name: ingressa
  namespace: migration
  annotations:
    konghq.com/strip-path: "true"
spec:
  ingressClassName: kong
  rules:
  - http:
      paths:
      - path: /mia
        pathType: Prefix
        backend:
          service:
            name: mia
            port:
              number: 80
---
kind: Ingress
metadata:
  name: ingressb
  namespace: migration
  annotations:
    konghq.com/strip-path: "true"
spec:
  ingressClassName: kong
  rules:
  - http:
      paths:
      - path: /mia
        pathType: Prefix
        backend:
          service:
            name: mia
            port:
              number: 80
```
is
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  creationTimestamp: null
  name: kong
  namespace: migration
spec:
  gatewayClassName: kong
  listeners:
  - name: http
    port: 80
    protocol: HTTP
status: {}
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  creationTimestamp: null
  name: ingressA-all-hosts
  namespace: migration
spec:
  parentRefs:
  - name: kong
  rules:
  - backendRefs:
    - name: mia
      port: 80
    - name: mia
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /mia
status:
  parents: []
```

pottekkat:
**What type of PR is this?**

/kind bug

This PR adds a check for duplicates while adding `BackendRefs` to the `HTTPRoute` rules.

See #102 for previous output. Current output with the same Ingress resource:

```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: HTTPRoute
metadata:
  creationTimestamp: null
  name: ingressa-all-hosts
  namespace: migration
spec:
  parentRefs:
  - name: kong
  rules:
  - backendRefs:
    - name: mia
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /mia
status:
  parents: []
```

The duplicate `mia:80` `backendRef` is avoided.

Fixes #102

```release-note
Duplicate `BackendRefs` are removed from the `HTTPRoute` rules.
```


pkg/i2gw/providers/common/converter.go
```diff
@@ -319,7 +319,23 @@ func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv
 	var errors field.ErrorList
 	var backendRefs []gatewayv1beta1.HTTPBackendRef
 
+	// track unique backends to avoid duplicates
+	uniqueBackends := make(map[string]bool)
```
mlavacca, line 323: I'd suggest using a `map[string]interface{}` and assigning `nil` to the values, as the values here do not really matter.

mlavacca (commented):
Thanks for this PR @pottekkat!

pkg/i2gw/providers/common/converter.go
pottekkat: Okay done: https://github.com/kubernetes-sigs/ingress2gateway/pull/104/commits/42cf5aa90af62979b2d263c2c86287041b033e4e

LiorLieberman (commented):
I think we should make the check after we call `toBackendRef`. `toBackendRef` function does a bit more checks and can also return a refernece to a CRD (not service). 
By checking for dups after this function we can benefit from a complete dedup of backend refs and avoid bugs

pkg/i2gw/providers/common/converter.go
levikobi: Since the intended functionality here is like a set, I suggest using `map[string]struct{}` to reduce memory usage (0 vs 16 bytes per entry).

pkg/i2gw/providers/common/converter.go
```diff
@@ -319,7 +319,23 @@ func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv
 	var errors field.ErrorList
 	var backendRefs []gatewayv1beta1.HTTPBackendRef
 
+	// track unique backends to avoid duplicates
+	uniqueBackends := make(map[string]interface{})
```
levikobi, line 323: Since the intended functionality here is like a set, I suggest using map[string]struct{} to reduce memory usage (0 vs 16 bytes per entry).

pkg/i2gw/providers/common/converter.go
```diff
@@ -319,7 +319,23 @@ func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv
...
+
 	for i, path := range paths {
+		if path.path.Backend.Service != nil {
+
+			backend := fmt.Sprintf("%s/%d", path.path.Backend.Service.Name, path.path.Backend.Service.Port.Number)
+
```
levikobi, line 329: Just to be bulletproof for invalid user input, I think we may want to check if `path.path.Backend.Service.Port` isn't nil

pottekkat:
@LiorLieberman I moved the check.

Ideally the if condition should be something like:

```go
if backendRef.BackendObjectReference.Kind != nil && *backendRef.BackendObjectReference.Kind == "Service" && backendRef.BackendObjectReference.Port != nil { }
```

But it seems like we don't set the `Kind` in the `toBackendRef` function if it is a Service.

pkg/i2gw/providers/common/converter.go
pottekkat: Added the check.

pkg/i2gw/providers/common/converter.go
```diff
@@ -319,12 +319,28 @@ func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv
...
+			backend := fmt.Sprintf("%s/%d", backendRef.BackendObjectReference.Name, *backendRef.BackendObjectReference.Port)
+
+			// skip duplicates
+			if _, exists := uniqueBackends[backend]; exists {
+				continue
+			}
+
+			// add new unique backend
+			uniqueBackends[backend] = nil
+		}
```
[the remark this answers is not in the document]
pottekkat: I tried with this:

```go
func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv1beta1.HTTPBackendRef, field.ErrorList) {
	var errors field.ErrorList
	var backendRefs []gatewayv1beta1.HTTPBackendRef

	// track unique backends to avoid duplicates
	uniqueBackendRefs := make(map[*gatewayv1beta1.HTTPBackendRef]interface{})
	for i, path := range paths {
		backendRef, err := toBackendRef(path.path.Backend, field.NewPath("paths", "backends").Index(i))
		if err != nil {
			errors = append(errors, err)
			continue
		}

		if _, exists := uniqueBackendRefs[&gatewayv1beta1.HTTPBackendRef{BackendRef: *backendRef}]; exists {
			continue
		}
		uniqueBackendRefs[&gatewayv1beta1.HTTPBackendRef{BackendRef: *backendRef}] = nil

		backendRefs = append(backendRefs, gatewayv1beta1.HTTPBackendRef{BackendRef: *backendRef})
	}

	return backendRefs, errors
}
```

But just checking for duplication does not work as the backendref has pointers. So even if the value of the ports are same, it fails to be identified as a duplicate because the pointers are different.

So I think we should create a new key for the uniquebackendrefs that has all the values needed from the backendref. But it might be difficult to maintain in the future where we might have to update the fields as backendrefs's spec changes.

What do you suggest we do here?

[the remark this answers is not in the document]
LiorLieberman: you are right, nice catch ! :)

This is what I came up with. (indeed if the spec change we will need to change but i dont anticipate it will change that often)

Also I a different solution -mine is probably not that good

```go
func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv1beta1.HTTPBackendRef, field.ErrorList) {
	var errors field.ErrorList
	type BackendRefMapKey struct {
		Name      gatewayv1beta1.ObjectName
		Namespace gatewayv1beta1.Namespace
		Port      gatewayv1beta1.PortNumber
		Group     gatewayv1beta1.Group
		Kind      gatewayv1beta1.Kind
	}

	mapKey := func(backendRef gatewayv1beta1.BackendRef) BackendRefMapKey {
		if backendRef.Port != nil {
			return BackendRefMapKey{
				Name: backendRef.Name,
				Port: *backendRef.Port,
			}
		}
		return BackendRefMapKey{
			Name:  backendRef.Name,
			Group: *backendRef.Group,
			Kind:  *backendRef.Kind,
		}
	}

	uniqueBackendRefs := map[BackendRefMapKey]gatewayv1beta1.BackendRef{}

	for i, path := range paths {
		backendRef, err := toBackendRef(path.path.Backend, field.NewPath("paths", "backends").Index(i))
		if err != nil {
			errors = append(errors, err)
			continue
		}
		uniqueBackendRefs[mapKey(*backendRef)] = *backendRef
	}
	var backendRefs []gatewayv1beta1.HTTPBackendRef
	for _, br := range uniqueBackendRefs {
		backendRefs = append(backendRefs, gatewayv1beta1.HTTPBackendRef{BackendRef: br})
	}
	return backendRefs, errors
```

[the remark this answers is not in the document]
pottekkat: Okay, if the spec does not change, this is the solution. I could also think of a similar solution. I will push the changes.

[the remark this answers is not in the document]
LiorLieberman: awesome, please also add a test for duplicated backend refs

dpasiukevich:
Folks, could I ask to move out the deduplication logic to its own func to make it reusable?

Use case: in the istio provider I'd like to use the backendRef deduplication logic. But there's no concept of `ingress` so I cannot use this logic if it's hidden in the method for `ingressRuleGroup`.

pottekkat:
@dpasiukevich Sounds good 👍🏽

pottekkat:
I moved the logic to a reusable function `DeDuplicateBackendRefs`. I'm unsure about writing tests because of the current output.

This is the current output. It appears to be repeating, which is a different issue:

```yaml
kind: Gateway
metadata:
  creationTimestamp: null
  name: kong
  namespace: migration
spec:
  gatewayClassName: kong
  listeners:
  - name: http
    port: 80
    protocol: HTTP
status: {}
---
kind: Gateway
metadata:
  creationTimestamp: null
  name: kong
  namespace: migration
spec:
  gatewayClassName: kong
  listeners:
  - name: http
    port: 80
    protocol: HTTP
status: {}
---
kind: HTTPRoute
metadata:
  creationTimestamp: null
  name: ingressa-all-hosts
  namespace: migration
spec:
  parentRefs:
  - name: kong
  rules:
  - backendRefs:
    - name: mia
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /mia
status:
  parents: []
---
kind: HTTPRoute
metadata:
  creationTimestamp: null
  name: ingressa-all-hosts
  namespace: migration
spec:
  parentRefs:
  - name: kong
  rules:
  - backendRefs:
    - name: mia
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /mia
status:
  parents: []
```

When testing with this input file:

```yaml
---
kind: Ingress
metadata:
  name: ingressa
  namespace: migration
  annotations:
    konghq.com/strip-path: "true"
spec:
  ingressClassName: kong
  rules:
  - http:
      paths:
      - path: /mia
        pathType: Prefix
        backend:
          service:
            name: mia
            port:
              number: 80
---
kind: Ingress
metadata:
  name: ingressb
  namespace: migration
  annotations:
    konghq.com/strip-path: "true"
spec:
  ingressClassName: kong
  rules:
  - http:
      paths:
      - path: /mia
        pathType: Prefix
        backend:
          service:
            name: mia
            port:
              number: 80
```

pkg/i2gw/providers/common/utils.go
```diff
@@ -155,3 +155,38 @@ func groupIngressPathsByMatchKey(rules []ingressRule) orderedIngressPathsByMatch
...
+			k.Name = backendRef.Name
+			k.Group = *backendRef.Group
+			k.Kind = *backendRef.Kind
+		}
+
+		if _, exists := uniqueBackendRefs[k]; exists {
+			continue
+		}
+
+		uniqueBackendRefs[k] = struct{}{}
```
mlavacca, line 188:
```suggestion
		uniqueBackendRefs[k] = {}
```

mlavacca:
> This is the current output. It appears to be repeating, which is a different issue (update: this seems to be happening for every conversion. Is this intended somehow?):

Regarding this duplicated output, this is a strange behavior I never saw. Could you please repeat the conversion setting the `--providers==kong` flag? If this does not happen in that case, I think it may be due to other providers converting the same (and it should not happen).

pottekkat:
> > This is the current output. It appears to be repeating, which is a different issue (update: this seems to be happening for every conversion. Is this intended somehow?):
> 
> Regarding this duplicated output, this is a strange behavior I never saw. Could you please repeat the conversion setting the `--providers==kong` flag? If this does not happen in that case, I think it may be due to other providers converting the same (and it should not happen).

@mlavacca That seems to be the issue.

mlavacca:
> > > This is the current output. It appears to be repeating, which is a different issue (update: this seems to be happening for every conversion. Is this intended somehow?):
> > 
> > 
> > Regarding this duplicated output, this is a strange behavior I never saw. Could you please repeat the conversion setting the `--providers==kong` flag? If this does not happen in that case, I think it may be due to other providers converting the same (and it should not happen).
> 
> @mlavacca That seems to be the issue. The output is fine with just enabling the Kong provider.

Ok, do you mind creating an issue for that? It'd be helpful to track such an error.

pkg/i2gw/providers/common/utils.go
```diff
@@ -155,3 +155,38 @@ func groupIngressPathsByMatchKey(rules []ingressRule) orderedIngressPathsByMatch
...
+		var k uniqueBackendRefsKey
+
+		if backendRef.Port != nil {
+			k.Name = backendRef.Name
+			k.Port = *backendRef.Port
+		} else {
+			k.Name = backendRef.Name
+			k.Group = *backendRef.Group
+			k.Kind = *backendRef.Kind
+		}
```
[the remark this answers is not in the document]
pottekkat: I went with @LiorLieberman 's suggestion as shown here: https://github.com/kubernetes-sigs/ingress2gateway/pull/104#discussion_r1425115085

levikobi:
> @mlavacca That seems to be the issue. The output is fine with just enabling the Kong provider.

@mlavacca @pottekkat It doesn't seem like providers are filtering resources that they aren't responsible for...
I think we can solve it by having each provider call `common.ToGateway` with his name as an argument and let the `common.ToGateway` function filter ingress resources based on the name.

pottekkat:
Opened issue for duplicate resources: https://github.com/kubernetes-sigs/ingress2gateway/issues/109

pkg/i2gw/providers/common/utils.go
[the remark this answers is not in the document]
mlavacca: I see, what I suggest is to always enforce group and kind, and check their validity to improve robustness
```suggestion
		group := gatewayv1.Group("")
		kind := gatewayv1.Kind("Service")
		// "" and "core" are equivalent, and here we need consistency
		if backendRef.Group != nil && *backendRef.Group != "core" {
			group = *backendRef.Group
		}
		if backendRef.Kind != nil {
			kind = *backendRef.Kind
		}

		k.Name = backendRef.Name
		k.Group = group
		k.Kind = kind
		if backendRef.Port != nil {
			k.Port = *backendRef.Port
		}
```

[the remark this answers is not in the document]
LiorLieberman: @mlavacca the suggestion was aligned with the way construct BackendRefs in [toBackendRef](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/common/converter.go#L371) function.

I think you have a valid point but we need to keep this too consistent. 

Also for the `group` logic you suggested. cant we simply do this ?

```go
if backendRef.Group != nil  {
	group = *backendRef.Group
} else {
        group = gatewayv1.Group("")
}
```

pkg/i2gw/providers/common/converter.go
```diff
@@ -328,7 +328,9 @@ func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv
 		backendRefs = append(backendRefs, gatewayv1.HTTPBackendRef{BackendRef: *backendRef})
 	}
 
-	return backendRefs, errors
+	DeDuplicatedBackendRefs := DeDuplicateBackendRefs(backendRefs)
```
LiorLieberman, line 331: can we do `uniqueBackendRefs`? I think this name sounds a bit better.  
This comment applies to all of the `DeDuplicatedBackendRefs` mentions in this PR.

Also, consider renaming the function to `removeBackendRefsDuplicates` which is slightly more descriptive.

pkg/i2gw/providers/common/converter.go
```diff
@@ -328,7 +328,9 @@ func (rg *ingressRuleGroup) configureBackendRef(paths []ingressPath) ([]gatewayv
...
+
+	return DeDuplicatedBackendRefs, errors
 }
```
LiorLieberman, line 334:
```suggestion
	return DeDuplicateBackendRefs(backendRefs), errors
}
```

LiorLieberman:
I am OOO next week. To avoid blocking this, approving this and will let @mlavacca and @levikobi give the final lgtm.

/approve
/hold

pkg/i2gw/providers/common/converter.go
pottekkat: Fixed.

pkg/i2gw/providers/common/converter.go
pottekkat: Fixed.

pkg/i2gw/providers/common/utils.go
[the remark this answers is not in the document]
pottekkat: @mlavacca @LiorLieberman, what should we do here? Should we keep it as it is?

pottekkat:
Added a test case and made other changes as suggested.

levikobi (commented):
Thank you for your efforts here, @pottekkat! This lgtm too, but I'll defer to @mlavacca for final approval in case he has anything else to add.

/label tide/merge-method-squash

pkg/i2gw/providers/common/utils.go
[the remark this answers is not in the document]
mlavacca: > @mlavacca the suggestion was aligned with the way construct BackendRefs in [toBackendRef](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/common/converter.go#L371) function.
> 
> I think you have a valid point but we need to keep this too consistent.
> 

I see, I agree we should be consistent. @pottekkat do you mind applying the same logic I suggested to the place @LiorLieberman pointed out?

> Also for the `group` logic you suggested. cant we simply do this ?
> 
> ```go
> if backendRef.Group != nil  {
> 	group = *backendRef.Group
> } else {
>         group = gatewayv1.Group("")
> }
> ```

We can't simply do this because "core" and "" are semantically the same, but if we mix them we end up having duplicated backendrefs. So, we need consistency and force "core" to be "" or "" to be "core", and this is just a matter of choice.

mlavacca (commented):
There is only one comment that still needs to be addressed. After that, lgtm!

LiorLieberman:
Hi @pottekkat, any updates on this? let us know if we can help unblocking this

pottekkat:
Hi @LiorLieberman I was caught up in some other work. I will complete this next week. Would that be okay?

LiorLieberman:
Thanks @pottekkat - do you mind aligning [toBackendRef](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/common/converter.go#L371) function as [requested](https://github.com/kubernetes-sigs/ingress2gateway/pull/104#discussion_r1429751236) ?

pottekkat:
@LiorLieberman Is it okay now? I am not sure if this is what you had in mind.

https://github.com/kubernetes-sigs/ingress2gateway/pull/104/commits/8fe2d6e0124d81dda30915b0a9d782c23e4f012a

LiorLieberman:
I reverted the last commit that I suggested to do, it think we dont really need to repeat the logic tbh. 

Sorry for the confusion, this looks good to me.

/approve

Issue #100, opened 2023-11-27
Add Istio provider support


LiorLieberman:
**What would you like to be added**:
Adding support for converting istio custom resources to Gateway API equivalent objects

**Why this is needed**:
Supporting istio provider in the tool will help and ease the migration of istio users to Gateway API.

/assign @dpasiukevich

```[tasklist]
### Tasks
- [ ] Add initial code istio implementaion and reading istio CRDs
- [ ] Add istio conversion logic
```

PR #110, opened 2023-12-20
Print generated GatewayClasses, TLSRoutes, TCPRoutes and ReferenceGrants.

dpasiukevich:
**What type of PR is this?**

/kind feature

Print generated GatewayClasses, TLSRoutes, TCPRoutes and ReferenceGrants.

These resources are already present in the provider conversion logic but are not propagated to the printer yet. The PR fixes this.

Related to the #100. Istio provider would generate TLSRoutes, TCPRoutes and ReferenceGrants for users.

```release-note
Print generated GatewayClasses, TLSRoutes, TCPRoutes and ReferenceGrants in addition to Gateways and HTTPRoutes
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -26,7 +26,6 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/client-go/tools/clientcmd"
-	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	// Call init function for the providers
 	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/ingressnginx"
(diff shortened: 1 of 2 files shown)


cmd/print.go
```diff
@@ -73,39 +72,103 @@ func (pr *PrintRunner) PrintGatewaysAndHTTPRoutes(cmd *cobra.Command, _ []string
...
+	for _, r := range gatewayResources {
+		resourceCount += len(r.GatewayClasses)
+		for _, gatewayClass := range r.GatewayClasses {
+			gatewayClass := gatewayClass
+			err := pr.resourcePrinter.PrintObj(&gatewayClass, os.Stdout)
+			if err != nil {
+				fmt.Printf("# Error printing %s GatewayClass: %v\n", gatewayClass.Name, err)
+			}
+		}
+	}
```
levikobi, line 97: It's just a nitpick, but is there any reason you repeated the outer for loop instead of having it once?

dpasiukevich: Kept the original logic: print all gateways from all providers, then all HTTPRoutes etc.

If it's in one loop, then the logic would change and we'd print all resources per provider.

levikobi: 100%

levikobi (approved):
Thanks @dpasiukevich!
/lgtm

LiorLieberman:
Thanks @dpasiukevich !
/approve

PR #112, opened 2023-12-28
Change ingress fetching to be isolated per provider

### LINKED ISSUE #109: Duplicate resources when using implementation-specific annotations
**What happened**:

When converting Ingress resources with implementation-specific annotations, the output contains duplicated `Gateway` and `HTTPRoute` resources.

See https://github.com/kubernetes-sigs/ingress2gateway/pull/104#issuecomment-1856463361 for details.

Also see: https://github.com/kubernetes-sigs/ingress2gateway/pull/104#issuecomment-1857538831 and https://github.com/kubernetes-sigs/ingress2gateway/pull/104#issuecomment-1857580813

See https://github.com/kubernetes-sigs/ingress2gateway/pull/104#issuecomment-1856463361 for reproducing.

No.

LiorLieberman:
**What type of PR is this?**
/kind feature

Currently we fetch all the ingresses in the main package and we pass them to all the providers. 

This is not only redundant (as every provider only cares about its own ingresses) but also creates some bugs. (https://github.com/kubernetes-sigs/ingress2gateway/issues/109)

This also increases consistency as we already fetch CRDs at the provider level and store them in a local storage.

Note: This PR is likely to yield some more issues and TODOs that wont be addressed in this PR.
For example changing `ToGatewayAPI` interface function signature that wont need `i2gw.InputResources` anymore.
Another thing would be to revisit `i2gw.InputResources` struct and check if we need it or we will change it to just a list of Ingresses as it is the only things it holds now.

I structured the commits to ease the review so you could review each commit independently

Fixes #109 

```release-note
Move ingress fetching logic to be isolated, per provider
```

### DIFF
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -48,7 +48,8 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 	var ingresses networkingv1.IngressList
 
 	providerByName, err := constructProviders(&ProviderConf{
-		Client: cl,
+		Client:    cl,
+		Namespace: namespace,
 	}, providers)
 	if err != nil {
 		return nil, err
@@ -136,7 +137,8 @@ func constructProviders(conf *ProviderConf, providers []string) (map[ProviderNam
 // ExtractObjectsFromReader extracts all objects from a reader,
 // which is created from YAML or JSON input files.
 // It retrieves all objects, including nested ones if they are contained within a list.
-func ExtractObjectsFromReader(reader io.Reader) ([]*unstructured.Unstructured, error) {
+// The function takes a namespace parameter to optionally return only namespaced resources.
+func ExtractObjectsFromReader(reader io.Reader, namespace string) ([]*unstructured.Unstructured, error) {
 	d := kubeyaml.NewYAMLOrJSONDecoder(reader, 4096)
 	var objs []*unstructured.Unstructured
 	for {
@@ -148,6 +150,9 @@ func ExtractObjectsFromReader(reader io.Reader) ([]*unstructured.Unstructured, e
 			return objs, fmt.Errorf("failed to unmarshal manifest: %w", err)
 		}
 		if u == nil {
+			continue
+		}
+		if namespace != "" && u.GetNamespace() != namespace {
 			continue
 		}
 		objs = append(objs, u)
@@ -187,15 +192,12 @@ func ConstructIngressesFromFile(l *networkingv1.IngressList, inputFile string, n
 	}
 
 	reader := bytes.NewReader(stream)
-	objs, err := ExtractObjectsFromReader(reader)
+	objs, err := ExtractObjectsFromReader(reader, namespace)
 	if err != nil {
 		return err
 	}
 
 	for _, f := range objs {
-		if namespace != "" && f.GetNamespace() != namespace {
-			continue
-		}
 		if !f.GroupVersionKind().Empty() && f.GroupVersionKind().Kind == "Ingress" {
 			var i networkingv1.Ingress
 			err = runtime.DefaultUnstructuredConverter.
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -45,8 +45,6 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 	}
 	cl = client.NewNamespacedClient(cl, namespace)
 
-	var ingresses networkingv1.IngressList
-
 	providerByName, err := constructProviders(&ProviderConf{
 		Client:    cl,
 		Namespace: namespace,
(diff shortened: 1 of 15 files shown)


pkg/i2gw/providers/istio/resource_reader.go
```diff
@@ -58,6 +60,25 @@ func (r *reader) readResourcesFromCluster(ctx context.Context) (*storage, error)
 	return &res, nil
 }
 
+func (r *reader) readResourcesFromFile(_ context.Context, filename string) (*storage, error) {
```
dpasiukevich, line 63: I'd suggest to remove `_ context.Context` from params, it's not used and not needed (this method doesn't satisfy any interface that requires such param).

pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -34,12 +41,51 @@ func newResourceReader(conf *i2gw.ProviderConf) *resourceReader {
 	}
 }
 
-func (r *resourceReader) ReadResourcesFromCluster(_ context.Context) error {
-	// ingress-nginx does not have any CRDs.
-	return nil
+func (r *resourceReader) readResourcesFromCluster(ctx context.Context) (*storage, error) {
```
dpasiukevich, line 44: I'd suggest to move out `readResourcesFromCluster` logic to a `common` package with the filter option, because this logic is used by both `ingressnginx` & `kong` providers exactly.

pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -34,12 +41,51 @@ func newResourceReader(conf *i2gw.ProviderConf) *resourceReader {
...
+		storage.Ingresses[types.NamespacedName{Namespace: ingress.Namespace, Name: ingress.Name}] = &ingressList.Items[i]
+	}
+
+	return storage, nil
 }
 
-func (r *resourceReader) ReadResourcesFromFile(_ context.Context, _ string) error {
-	// ingress-nginx does not have any CRDs.
-	return nil
+func (r *resourceReader) readResourcesFromFile(_ context.Context, filename string) (*storage, error) {
```
dpasiukevich, line 63: I'd suggest to move out `readResourcesFromFile` logic to a `common` package with the filter option, because this logic is used by both `ingressnginx` & `kong` providers exactly.

dpasiukevich:
/hold

to ensure it's not automerged

levikobi:
I too fell for it before @dpasiukevich 😅

pkg/i2gw/providers/ingressnginx/resource_reader.go
LiorLieberman: I thought about it, we can move this logic to common as `readIngressesFromCluster`. readResourcesFromCluster should still be a local function on the provider to orchestrate the read of whatever resources they need. 

wdyt?

pkg/i2gw/providers/ingressnginx/resource_reader.go
LiorLieberman: same comment as above

pkg/i2gw/providers/ingressnginx/resource_reader.go
dpasiukevich: sgtm

pkg/i2gw/ingress2gateway.go
```diff
@@ -136,7 +137,8 @@ func constructProviders(conf *ProviderConf, providers []string) (map[ProviderNam
 // ExtractObjectsFromReader extracts all objects from a reader,
 // which is created from YAML or JSON input files.
 // It retrieves all objects, including nested ones if they are contained within a list.
-func ExtractObjectsFromReader(reader io.Reader) ([]*unstructured.Unstructured, error) {
+// The function takes a namespace parameter to optionally return only namespaced resources.
+func ExtractObjectsFromReader(reader io.Reader, namespace string) ([]*unstructured.Unstructured, error) {
```
levikobi, line 130: If providers will read ingress resources themselves, I don't see the value in having this function here anymore... Am I missing something?

pkg/i2gw/providers/istio/resource_reader.go
```diff
@@ -58,6 +60,25 @@ func (r *reader) readResourcesFromCluster(ctx context.Context) (*storage, error)
...
+	stream, err := os.ReadFile(filename)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file %v: %w", filename, err)
+	}
+
+	unstructuredObjects, err := i2gw.ExtractObjectsFromReader(bytes.NewReader(stream), r.conf.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract objects: %w", err)
+	}
```
levikobi, line 72: As a followup to my last comment, I see we're using `ExtractObjectsFromReader` here. Maybe move it to the `common` package?

pkg/i2gw/ingress2gateway.go
LiorLieberman: The value is that they will share the code.

pkg/i2gw/providers/istio/resource_reader.go
LiorLieberman: can move it to common yeah, good idea

LiorLieberman:
resolved all the comments here https://github.com/kubernetes-sigs/ingress2gateway/pull/116/commits/835e917876c1a9056ed4cbfb3d071b12e3db07b5

PR #114, opened 2023-12-29
Revert "per-provider-ingress-fetching-isolation"

LiorLieberman:
Reverts kubernetes-sigs/ingress2gateway#112

It was merged accidentally and still waiting further review

PR #115, opened 2023-12-29
[Revert the revert] Change ingress fetching to be isolated per provider

### LINKED ISSUE #109: Duplicate resources when using implementation-specific annotations

LiorLieberman:
Reverts kubernetes-sigs/ingress2gateway#114 which is the revert of kubernetes-sigs/ingress2gateway#112

Better to review kubernetes-sigs/ingress2gateway#112 instead as the commits there are structured and easy the review so you could review each commit independently

I copy pasted the below PR description from kubernetes-sigs/ingress2gateway#112

**What type of PR is this?**
/kind feature

Currently we fetch all the ingresses in the main package and we pass them to all the providers. 

This is not only redundant (as every provider only cares about its own ingresses) but also creates some bugs. (https://github.com/kubernetes-sigs/ingress2gateway/issues/109)

This also increases consistency as we already fetch CRDs at the provider level and store them in a local storage.

Note: This PR is likely to yield some more issues and TODOs that wont be addressed in this PR.
For example changing `ToGatewayAPI` interface function signature that wont need `i2gw.InputResources` anymore.
Another thing would be to revisit `i2gw.InputResources` struct and check if we need it or we will change it to just a list of Ingresses as it is the only things it holds now.

I structured the commits to ease the review so you could review each commit independently

Fixes #109 

```release-note
Move ingress fetching logic to be isolated, per provider
```

### DIFF
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -45,30 +45,19 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 	}
 	cl = client.NewNamespacedClient(cl, namespace)
 
-	var ingresses networkingv1.IngressList
-
 	providerByName, err := constructProviders(&ProviderConf{
-		Client: cl,
+		Client:    cl,
+		Namespace: namespace,
 	}, providers)
 	if err != nil {
 		return nil, err
 	}
 
-	resources := InputResources{}
-
 	if inputFile != "" {
-		if err = ConstructIngressesFromFile(&ingresses, inputFile, namespace); err != nil {
-			return nil, fmt.Errorf("failed to read ingresses from file: %w", err)
-		}
-		resources.Ingresses = ingresses.Items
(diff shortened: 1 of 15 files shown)


LiorLieberman:
/cc @dpasiukevich @levikobi @mlavacca 
/hold

Since @dpasiukevich  already left comments on kubernetes-sigs/ingress2gateway#112 lets leave all further comments there and I will make the changes to this PR

Issue #122, opened 2024-01-15
--input_file requires an available cluster, does not support yaml files with multiple resources, does not give errors


simonfelding:
**What happened**:
I wanted to convert ingresses in a multi-yaml file usin --input_file, without access to the cluster.

`./ingress2gateway print --input_file harbor.yaml`
resulted in
`Error: failed to create client: Get "http://localhost:6443/api?timeout=32s": dial tcp 127.0.0.1:6443: connect: connection refused`

because the cluster is airgapped and unavailable for my laptop. With connection to the cluster, it returns
`No resources found in default namespace`.

**What you expected to happen**:
ingress yamls are converted to httproutes.

**How to reproduce it (as minimally and precisely as possible)**:
`helm template harbor oci://registry-1.docker.io/bitnamicharts/harbor --set exposureType=ingress -set ingress.core.ingressClassName=nginx -n harbor > harbor.yaml`

`./ingress2gateway print --input_file harbor.yaml`

the above gives no meaningful output, but should either give an error or the expected output.

simonfelding:
related to https://github.com/kubernetes-sigs/ingress2gateway/pull/78

dpasiukevich:
Should be fixed by https://github.com/kubernetes-sigs/ingress2gateway/pull/128

Now for reading local files the k8sclient creation is skipped.

LiorLieberman:
@simonfelding I believe this issue is fixed now.
Can you confirm?

LiorLieberman:
closing this as fixed. Feel free to reopen if needed.

davinkevin:
@LiorLieberman I have still this error with 0.2.0: 

<img width="426" alt="image" src="https://github.com/kubernetes-sigs/ingress2gateway/assets/1970922/d6e42ef7-f30a-4b8e-96fe-9031c9baad69">

My only wish is to use the cli offline, on our `gitops` repository

levikobi:
Hi @davinkevin, I guess the resources in your `ingress.yaml` file do not specify a namespace (and if they do, it isn't the default one). Please use -A (--all-namespaces) flag

davinkevin:
Thank you for your answer @levikobi . Result is different but not really better.

<img width="871" alt="image" src="https://github.com/kubernetes-sigs/ingress2gateway/assets/1970922/3f9a61ff-eefb-4f8b-b727-b14e8412d06a">

My file is just this: 

```
apiVersion: networking.k8s.io/v1
[...]

PR #128, opened 2024-01-31
Skip k8s client creation when reading local file.

### LINKED ISSUE #127: Tool fails to translate local file when there's incompatible .kube_config file
**What happened**:

When trying to translate locally stored file, converter fails if there's incompatible .kube_config file with the tools dependencies versions.

```
ingress2gateway print --providers=istio --input_file=helloworld-gateway.yaml
Error: failed to create client: no Auth Provider found for name "oidc"
```

ingress2gateway should run correctly and return output from the local input file. The tool should run correctly irregardless of any .kube_config state during local file conversion.

@frankbu may provide more details as he has experienced such case.

dpasiukevich:
**What type of PR is this?**

/kind bug

Make the tool more robust. Don't read/run k8s client config when reading local file. E.g. user may have incorrect/incompatible .kube_config but the user tries to convert a local file. `.kube_config` file should not affect the tool run in such case.

I've checked all providers, none needs a k8sclient to translate a local file, which is expected.

Fixes #127

```release-note
Skip k8s client creation when reading local file.
```

### DIFF
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -26,19 +26,23 @@ import (
 )
 
 func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile string, providers []string) ([]GatewayResources, error) {
-	conf, err := config.GetConfig()
-	if err != nil {
-		return nil, fmt.Errorf("failed to get client config: %w", err)
-	}
+	var clusterClient client.Client
 
-	cl, err := client.New(conf, client.Options{})
-	if err != nil {
-		return nil, fmt.Errorf("failed to create client: %w", err)
+	if inputFile == "" {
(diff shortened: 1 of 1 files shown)


levikobi (commented):
Thanks @dpasiukevich 
/lgtm

LiorLieberman:
@dpasiukevich I think it also fixes #122 right?

dpasiukevich:
/hold

for the controlled merge

dpasiukevich:
> @dpasiukevich I think it also fixes #122 right?

Yes, by the looks it should fix this issue as well. Let me double check that case later today and close that issue if it's the case.

LiorLieberman:
> Yes, by the looks it should fix this issue as well. Let me double check that case later today and close that issue if it's the case.

SG
/unhold
/approve

PR #132, opened 2024-02-13
Added IngressClass objects support #117

### LINKED ISSUE #117: Add support for `IngressClass` objects
Add support for reading and taking into account `IngressClass` objects.

**Why this is needed**:
If users do not specify `ingressClassName,` we need to understand the default `IngressClass` that should be used for the conversion.

Varun1300211:
**What type of PR is this?**
This pull request implements IngressClass support.

Add one of the following kinds:
/kind feature

It retrieves the default IngressClass based on its API group (networking.k8s.io) and a dedicated annotation (ingressclass.kubernetes.io/is-default-class).
If no explicit IngressClass is specified in the Ingress object or via annotation, the default class is used for resource processing.

Fixes #117

```NONE

``` 
Please do let me know if it was done in the correct way.

### DIFF
--- a/pkg/i2gw/providers/common/utils.go
+++ b/pkg/i2gw/providers/common/utils.go
@@ -18,21 +18,74 @@ package common
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"log"
 	"regexp"
+	"strings"
 
 	networkingv1 "k8s.io/api/networking/v1"
 	networkingv1beta1 "k8s.io/api/networking/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
 var k8sClientset *kubernetes.Clientset
 
-func GetDefaultIngressClass() (string, error) {
-	ingressClasses, err := k8sClientset.NetworkingV1().IngressClasses().List(context.Background(), metav1.ListOptions{})
+func init() {
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Fatal(errors.New("unable to create config"))
+	}
+
+	k8sClientset, err = kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func getFileExtension(fileName string) string {
+	parts := strings.Split(fileName, ".")
+	if len(parts) > 1 {
+		return parts[len(parts)-1]
+	}
+	return ""
+}
+
+func readIngressClassesFromFile(data []byte, extension string) (*networkingv1.IngressClassList, error) {
+	ingressClasses := &networkingv1.IngressClassList{}
+	var err error
+
+	switch extension {
+	case "yaml":
+		err = yaml.Unmarshal(data, ingressClasses)
+	case "json":
+		err = json.Unmarshal(data, ingressClasses)
+	default:
+		return nil, fmt.Errorf("unsupported file type: %s", extension)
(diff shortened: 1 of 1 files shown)


Varun1300211:
Done all the changes requested by you @LiorLieberman, please let me know if anything else is required.

pkg/i2gw/providers/common/utils.go
```diff
@@ -17,24 +17,50 @@ limitations under the License.
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/client-go/kubernetes"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+var k8sClientset *kubernetes.Clientset
+
+func GetDefaultIngressClass() (string, error) {
+	ingressClasses, err := k8sClientset.NetworkingV1().IngressClasses().List(context.Background(), metav1.ListOptions{})
```
levikobi, line 35: We support 2 modes for reading resources: from the cluster or a local file.
Here it seems that you're always reading from the cluster. We should come up with a way to take into account the user's choice

pkg/i2gw/providers/common/utils.go
```diff
@@ -17,24 +17,50 @@ limitations under the License.
...
 	if ingress.Spec.IngressClassName != nil && *ingress.Spec.IngressClassName != "" {
 		ingressClass = *ingress.Spec.IngressClassName
 	} else if _, ok := ingress.Annotations[networkingv1beta1.AnnotationIngressClass]; ok {
 		ingressClass = ingress.Annotations[networkingv1beta1.AnnotationIngressClass]
 	} else {
-		ingressClass = ingress.Name
+		defaultIngressClass, err := GetDefaultIngressClass()
+		if err != nil {
+			return ""
+		}
```
levikobi, line 62: We should probably return an error in this case

levikobi (commented):
Thanks @Varun1300211 and welcome!
I left a few comments regarding some things that need to be taken into account.
If you have any questions or need any help feel free to let us know

Varun1300211:
Hi @levikobi, I made the changes suggested by you. Please take a look.

pkg/i2gw/providers/common/utils.go
```diff
@@ -17,24 +17,104 @@ limitations under the License.
...
+	var err error
+
+	switch extension {
+	case "yaml":
+		err = yaml.Unmarshal(data, ingressClasses)
+	case "json":
+		err = json.Unmarshal(data, ingressClasses)
+	default:
+		return nil, fmt.Errorf("unsupported file type: %s", extension)
+	}
```
levikobi, line 72: Maybe I missed something but it seems like you're always sending an empty slice of data

pkg/i2gw/providers/common/utils.go
```diff
@@ -17,24 +17,104 @@ limitations under the License.
...
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Fatal(errors.New("unable to create config"))
+	}
+
+	k8sClientset, err = kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
```
levikobi, line 51: This init function will always try to create a k8s client, even if the user doesn't have a cluster and want to convert some local file.

You'll need to pass the user's choice of where he wants to read resources from (cluster or file) and the k8s client, which was initialized in the i2gw package.

levikobi (changes requested):
Thanks again @Varun1300211!
I think it may help you to try playing around with the tool for a bit and see the different options it provides. Afterwards maybe try to compile your version and see what changes/breaks.

This new addition needs to take into account different options the user may pass.

LiorLieberman:
@Varun1300211 just checking in, do you need help with moving this forward?

LiorLieberman:
@Varun1300211 Just checking in, please let us know if you need any help with moving this forward

LiorLieberman:
@Varun1300211 a friendly ping, are you still planning to work on this?

PR #134, opened 2024-02-19
added ability to respect the kubebuilder maxlimit

### LINKED ISSUE #124: Make sure generated Gateway-API object fields obey kubebuilder max item validation
Add a mechanism that considers the kube-builder max item validation on Gateway-API list fields. See [this for example](https://github.com/kubernetes-sigs/gateway-api/blob/ee42a6d5d81283d9f41a85627b93e4492b876638/apis/v1/httproute_types.go#L223). 

One option is to validate the constraint, and if it is exceeded - throw an appropriate message.
Another option is to add conversion logic that splits objects by these constraints, which will be more challenging.

**Why this is needed**:
To help migrating users who violate those constraints. See https://github.com/kubernetes-sigs/gateway-api/issues/2501 as an example.

swastik959:
**What type of PR is this?**

/kind feature
adds the ability to limit of objects to respect the kubebuilder max count

Fixes #124 

```release-note

```

### DIFF
--- a/pkg/i2gw/providers/kong/plugins.go
+++ b/pkg/i2gw/providers/kong/plugins.go
@@ -17,6 +17,7 @@ limitations under the License.
 package kong
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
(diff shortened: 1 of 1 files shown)


swastik959:
hi @levikobi I have tried to create a pr for the feedback if this is what was meant from the issue ?

levikobi (commented):
Hi @swastik959! Thank you for taking on this one.

The maxlimit constant isn't provider-specific limitation. We either need to add this as a common validation to each provider, or (a better approach IMO) check that after the translation already occurred.

Also, please avoid the use of panic.

LiorLieberman:
@swastik959 Do you still work on this? Let us know if we can help you to move this forward

LiorLieberman:
I am going to close the PR un-assign this now, so others can pick it up. Feel free to reply if you still want to work on this.

PR #139, opened 2024-03-04
fix flakly ingressnginx test

LiorLieberman:
**What type of PR is this?**
/kind bug

Ingress-nginx support mergind different ingresses to one HTTPRoute, but since we used a map to store the ingresses, the order is not consistent, causing tests to be flaky

Fixes #

```release-note
Fixed ingress-nginx conversion tests
```

/cc @levikobi 
FYI @YTGhost

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/converter.go
+++ b/pkg/i2gw/providers/ingressnginx/converter.go
@@ -19,7 +19,6 @@ package ingressnginx
 import (
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/common"
-	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/i2gw/providers/ingressnginx/resource_reader.go
```diff
@@ -42,7 +44,14 @@ func (r *resourceReader) readResourcesFromCluster(ctx context.Context) (*storage
...
 	}
-	storage.Ingresses = ingresses
+	ingNames := []types.NamespacedName{}
+	for ing := range ingresses {
+		ingNames = append(ingNames, ing)
+	}
+	sort.Slice(ingNames, func(i, j int) bool {
+		return ingNames[i].Name < ingNames[j].Name
+	})
+	storage.Ingresses = OrderedIngressMap{ingressNames: ingNames, ingressObjects: ingresses}
```
levikobi, line 54: I think having this functionality as a method/constructor of the `OrderedIngressMap` type will make things easier to maintain, especially as we already duplicate it later in this file.

pkg/i2gw/providers/ingressnginx/converter.go
```diff
@@ -41,8 +41,8 @@ func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.Erro
 
 	// TODO(liorliberman) temporary until we decide to change ToGateway and featureParsers to get a map of [types.NamespacedName]*networkingv1.Ingress instead of a list
 	ingressList := []networkingv1.Ingress{}
-	for _, ing := range storage.Ingresses {
-		ingressList = append(ingressList, *ing)
+	for _, ing := range storage.Ingresses.ingressNames {
+		ingressList = append(ingressList, *storage.Ingresses.ingressObjects[ing])
```
levikobi, line 45: Consider creating a method on the `OrderedIngressMap` (something like `OrderedIngressMap.Ingresses()`) that gives you the ordered list of ingresses, to abstract away the fact that we must access the `ingressObjects` using the ordered `ingressNames`

levikobi (commented):
Overall looks good to me, I have some suggestions surrounding the maintainability of this addition

pkg/i2gw/providers/ingressnginx/resource_reader.go
LiorLieberman: Done, Thanks!

levikobi (commented):
/lgtm

PR #141, opened 2024-03-08
feat: deprecate i2gw.InputResources

### LINKED ISSUE #113: Find a good use of i2gw.InputResources or deprecate it
We need to revisit the decision to use `i2gw.InputResources` struct or deprecate it altogether.

**Why this is needed**:
Currently this struct only holds a list of Ingresses, I couldn't think of a reason we will need other resources as input resources. 
I think this was added originally to be able to support CRD propagation to providers but since we move to a "fetch and store" per provider logic we need to revisit whether this is needed. 

Another sub task that comes with it is changing `ToGatewayAPI` function signature  on the Provider interface as the input resources are no longer in use after https://github.com/kubernetes-sigs/ingress2gateway/pull/112

- [ ] Find a good use of i2gw.InputResources or deprecate it
- [ ] Remove input resources from [ToGatewayAPI](https://github.com/kubernetes-sigs/ingress2gateway/blob/654e76fa2ec341b9eaf9f7ba42f12e3305c7b993/pkg/i2gw/provider.go#L73C2-L73C14) function.

YTGhost:
**What type of PR is this?**
/kind feature

- [x] Deprecate i2gw.InputResources
- [x] Remove input resources from [ToGatewayAPI](https://github.com/kubernetes-sigs/ingress2gateway/blob/654e76fa2ec341b9eaf9f7ba42f12e3305c7b993/pkg/i2gw/provider.go#L73C2-L73C14) function.

Fixes #113 

```release-note
Deprecate `i2gw.InputResources` and remove input resources from `ToGatewayAPI` function
```

### DIFF
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -68,8 +68,7 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 		errs             field.ErrorList
 	)
 	for _, provider := range providerByName {
-		// TODO(#113) Remove input resources from ToGatewayAPI function
-		providerGatewayResources, conversionErrs := provider.ToGatewayAPI(InputResources{})
+		providerGatewayResources, conversionErrs := provider.ToGatewayAPI()
 		errs = append(errs, conversionErrs...)
 		gatewayResources = append(gatewayResources, providerGatewayResources)
 	}
--- a/pkg/i2gw/provider.go
+++ b/pkg/i2gw/provider.go
@@ -69,9 +69,9 @@ type CustomResourceReader interface {
 // conversion functions.
 type ResourceConverter interface {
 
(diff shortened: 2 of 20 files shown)


YTGhost:
Hi @LiorLieberman PTAL

YTGhost:
> Thats awesome, thank you @YTGhost ! /ok-to-test
> 
> In the meantime, please update the release notes

Sure

LiorLieberman (commented):
This PR looks good to me. 
/cc @levikobi @mlavacca can you please take a look?

levikobi (commented):
Thanks @YTGhost 
This is quite a big change so I played around with this for a bit. I Couldn't find something that broke and tests are passing so lgtm.

@mlavacca if you could take a look as well it would be great
/lgtm

mlavacca (approved):
Thanks YTGhost!

/lgtm

PR #144, opened 2024-03-20
fix: execution fails when istio CRD is not installed in the cluster

### LINKED ISSUE #138: Execution fails when CRDs are not installed in the cluster
**What happened**:
When running the tool without `--providers` flag, it defaults to all providers. Then, when istio provider attempts to read its CRs, it fails the execution

If I am not specifying istio as the desired provider, execution should work regardless of whether I have the CRDs installed or not

Run `ingress2gateway print` in a cluster without istio CRDs intalled.

Here is the error
```
./ingress2gateway print
Error: failed to read istio resources from the cluster: failed to read resources from cluster: failed to read gateways: failed to list istio gateways: failed to get API group resources: unable to retrieve the complete list of server APIs: networking.istio.io/v1beta1: the server could not find the requested resource
```

Two suggestion in mind for how to address this:
1) Enforce to specify `--providers`, and not default to all providers
2) Fail soft with a warning if the provider specified is not the provider whose CRDs are not installed

/assign @dpasiukevich 
/cc @levikobi 
/cc @mlavacca

YTGhost:
**What type of PR is this?**
/kind bug

1. Enforce to specify --providers, and not default to all providers
2. Fail soft with a warning if the provider specified is not the provider whose CRDs are not installed

Fixes #138 

```release-note

```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -67,7 +67,7 @@ type PrintRunner struct {
 func (pr *PrintRunner) PrintGatewaysAndHTTPRoutes(cmd *cobra.Command, _ []string) error {
 	err := pr.initializeResourcePrinter()
 	if err != nil {
-		return fmt.Errorf("failed to initialize resrouce printer: %w", err)
+		return fmt.Errorf("failed to initialize resource printer: %w", err)
 	}
 	err = pr.initializeNamespaceFilter()
 	if err != nil {
@@ -245,10 +245,11 @@ func newPrintCommand() *cobra.Command {
 		`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
 if specified with --namespace.`)
 
-	cmd.Flags().StringSliceVar(&pr.providers, "providers", i2gw.GetSupportedProviders(),
+	cmd.Flags().StringSliceVar(&pr.providers, "providers", nil,
 		fmt.Sprintf("If present, the tool will try to convert only resources related to the specified providers, supported values are %v", i2gw.GetSupportedProviders()))
 
 	cmd.MarkFlagsMutuallyExclusive("namespace", "all-namespaces")
+	cmd.MarkFlagRequired("providers")
 	return cmd
 }
 
--- a/pkg/i2gw/providers/istio/resource_reader.go
+++ b/pkg/i2gw/providers/istio/resource_reader.go
@@ -26,6 +26,7 @@ import (
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/common"
 	istiov1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -43,6 +44,12 @@ func newResourceReader(conf *i2gw.ProviderConf) reader {
 
 func (r *reader) readResourcesFromCluster(ctx context.Context) (*storage, error) {
 	res := newResourcesStorage()
+
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	if err := r.conf.Client.Get(ctx, types.NamespacedName{Name: CRDName}, crd); err != nil {
+		log.Printf("CRD '%s' is not installed in the cluster, will not continue reading resources", CRDName)
+		return res, nil
+	}
 
 	gateways, err := r.readGatewaysFromCluster(ctx)
 	if err != nil {
--- a/README.md
+++ b/README.md
@@ -61,14 +61,14 @@ This above command will:
 
 ### `print` command
 
-| Flag           | Default Value           | Required | Description                                                  |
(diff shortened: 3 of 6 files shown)


YTGhost:
@dpasiukevich @LiorLieberman PTAL

cmd/print.go
```diff
@@ -65,9 +65,12 @@ type PrintRunner struct {
 // converted Gateways and HTTP Routes. The steps include reading from the source,
 // construct ingresses, convert them, then print them out.
 func (pr *PrintRunner) PrintGatewaysAndHTTPRoutes(cmd *cobra.Command, _ []string) error {
+	if pr.providers == nil || len(pr.providers) == 0 {
+		return fmt.Errorf("you must specify at least one provider with --providers")
+	}
```
LiorLieberman, line 70: consider using [MarkFlagRequired](https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired) instead

YTGhost: That's a good idea, thanks!

LiorLieberman (commented):
Thanks, left some comments. 
A general comment, you made two proposed solutions to the same problem. if we change the code to fail soft we might not need to enforce specifying providers.

pkg/i2gw/providers/istio/resource_reader.go
```diff
@@ -44,6 +45,12 @@ func newResourceReader(conf *i2gw.ProviderConf) reader {
 func (r *reader) readResourcesFromCluster(ctx context.Context) (*storage, error) {
 	res := newResourcesStorage()
 
+	crd := &apiextensionsv1.CustomResourceDefinition{}
+	if err := r.conf.Client.Get(ctx, types.NamespacedName{Name: CRDName}, crd); err != nil {
+		log.Printf("CRD '%s' is not installed in the cluster, will not continue reading resources", CRDName)
+		return res, nil
+	}
+
```
LiorLieberman, line 53: Sorry this was not added to the previous review. 
The intention was to change the call to read ANY CRD. In this case you only cover Istio Gateway but it is also relevant for VirtualServices and any other provider's CRD that will be added in the future.

So either you change istio calls to "fail soft" or you create a general helper that doing it

mlavacca (commented):
I lean toward not requiring any provider when calling the CLI, instead, I propose the fail soft as @LiorLieberman proposed. I think the soft fail logic should be introduced one level abve. If a provider fails to read resources, we might want to just log the error and continue with other providers. We should implement this outside the istio provider as we might have the same issue with other Providers (Kong just implemented `TCPIngress`es conversion, and I think we have the same issue there). 

At the moment, logging messages is problematic because of https://github.com/kubernetes-sigs/ingress2gateway/issues/145, but that's a separate issue.

levikobi (commented):
+1 to what @mlavacca and @LiorLieberman said. I also would not want to see an enforcement of specifying the `providers` flags, as this would be one more friction point that users have to figure out.

Alongside Kong, Apisix also faces this issue now. We need to solve this problem for all the providers (one level above, again, as @mlavacca said).

I think a reasonable solution is to have each provider check if his CRDs are installed; if they're not, return a typed error that specifies that. This error will result in a "fail soft" behavior one level above.

At the end of the run, we can specify which providers participated in the translation instead of describing which didn't. I think that's a better choice since a naive user who has some manifests but doesn't specify providers shouldn't receive multiple error messages about missing CRDs from different providers (which he might not even be familiar with).

LiorLieberman:
@YTGhost This is a very important fix to add. 
Do you still have the bandwidth to work on this?

levikobi:
> This is a very important fix to add. 

+1

YTGhost:
> @YTGhost This is a very important fix to add. Do you still have the bandwidth to work on this?

@LiorLieberman Sorry for the delay, Let me fix it tonight.

And let me confirm what we need to do now: 
- [ ] Not enforcing the specification of `--providers`
- [ ] Implement a general helper that checks whether the list of CRDs passed to it are installed in the cluster. If not installed, it will return a typed error. For this typed error, we will use a soft fail logic to handle it at a higher level.
- [ ] Introduce this helper to check each current provider.

mlavacca:
> > @YTGhost This is a very important fix to add. Do you still have the bandwidth to work on this?
> 
> @LiorLieberman Sorry for the delay, I will fix it as soon as possible.
> 
> And let me confirm what we need to do now:
> 
> * [ ]  Not enforcing the specification of `--providers`

👍 

> * [ ]  Implement a general helper that checks whether the list of CRDs passed to it are installed in the cluster. If not installed, it will return a typed error. For this typed error, we will use a soft fail logic to handle it at a higher level.

I don't think we need a helper to check the CRDs are installed in the cluster. This is the flow I have in mind:
-  The providers' `readResourcesFromCluster` functions returns a typed error when the CRD is not installed in the cluster (instead of a general error)
- the [caller](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/ingress2gateway.go#L93-L100) of `readResourcesFromCluster` checks the error type. If it is related to missing CRDs, we add the provider to the list of the ones who did not provide translation, as @levikobi suggested.

LiorLieberman:
@YTGhost can we help to speed this up?
It is also ok if you have had limited bandwidth recently, just that this feature is really important so if you did, someone else would implement this.

YTGhost:
> @YTGhost can we help to speed this up? It is also ok if you have had limited bandwidth recently, just that this feature is really important so if you did, someone else would implement this.

Yes, sorry, I've been busy with some school matters recently, and don't have enough bandwidth. It would be great if someone could help implement it. Apologies once again.

mlavacca:
closing as superseded by #153

PR #152, opened 2024-04-10
Improve error handling across kong & ingress-nginx

### LINKED ISSUE #151: Unnecessary panic calls causes unclear errors
**What happened**:
We have few places in the codebase where we call panic for no good reason.
For example, When we get calculation errors [here](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/ingressnginx/canary.go#L44-L45) we never see them and panic is called. 

We should not call panic
One example would be to run the tool on an ingress with `port.name` (which is unsupported)
Pasting an ingress.yaml example for simplicity:

```yaml
kind: Ingress
metadata:
  name: podinfo.ingress.k8s.local
spec:
  ingressClassName: kong
  rules:
    - host: podinfo.ingress.k8s.local
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: podinfo
                port:
                  name: http
```

Seems all occurrences were copy pasted from the original code in https://github.com/kubernetes-sigs/ingress2gateway/pull/40.

/assign @levikobi

levikobi:
**What type of PR is this?**

/kind bug

Improves error handling across Kong and ingress-nginx providers by removing panic calls and stopping conversions if an error occurs.
Also, prevents the tool from crashing in case no `pathType` is specified.

Fixes #151 

```release-note
* Improves error handling for Kong and Ingress Nginx providers, making them more understandable.
* Prevents the tool from crashing in case no `pathType` is specified
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/canary.go
+++ b/pkg/i2gw/providers/ingressnginx/canary.go
@@ -47,7 +47,9 @@ func canaryFeature(ingressResources i2gw.InputResources, gatewayResources *i2gw.
 			key := types.NamespacedName{Namespace: path.ingress.Namespace, Name: common.RouteName(rg.Name, rg.Host)}
 			httpRoute, ok := gatewayResources.HTTPRoutes[key]
 			if !ok {
-				panic("HTTPRoute not exists - this should never happen")
+				// If there wasn't an HTTPRoute for this Ingress, we can skip it as something is wrong.
+				// All the available errors will be returned at the end.
+				continue
 			}
 
 			patchHTTPRouteWithBackendRefs(&httpRoute, backendRefs)
--- a/pkg/i2gw/providers/kong/converter.go
+++ b/pkg/i2gw/providers/kong/converter.go
@@ -51,31 +51,35 @@ func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.Erro
 		ingressList = append(ingressList, *ingress)
 	}
 
-	globalErrs := field.ErrorList{}
+	errorList := field.ErrorList{}
 
 	// Convert plain ingress resources to gateway resources, ignoring all
 	// provider-specific features.
 	gatewayResources, errs := common.ToGateway(ingressList, c.implementationSpecificOptions)
 	if len(errs) > 0 {
-		globalErrs = append(globalErrs, errs...)
+		errorList = append(errorList, errs...)
 	}
 
 	tcpGatewayResources, errs := crds.TCPIngressToGatewayAPI(storage.TCPIngresses)
 	if len(errs) > 0 {
-		globalErrs = append(globalErrs, errs...)
+		errorList = append(errorList, errs...)
+	}
+
+	if len(errorList) > 0 {
+		return i2gw.GatewayResources{}, errorList
 	}
 
 	gatewayResources, errs = i2gw.MergeGatewayResources(gatewayResources, tcpGatewayResources)
 	if len(errs) > 0 {
-		globalErrs = append(globalErrs, errs...)
+		return i2gw.GatewayResources{}, errs
 	}
 
 	for _, parseFeatureFunc := range c.featureParsers {
 		// Apply the feature parsing function to the gateway resources, one by one.
 		errs = parseFeatureFunc(i2gw.InputResources{Ingresses: ingressList}, &gatewayResources)
 		// Append the parsing errors to the error list.
-		globalErrs = append(globalErrs, errs...)
+		errorList = append(errorList, errs...)
 	}
 
-	return gatewayResources, globalErrs
+	return gatewayResources, errorList
 }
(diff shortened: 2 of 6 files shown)


pkg/i2gw/providers/kong/converter.go
```diff
@@ -51,25 +51,25 @@ func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.Erro
...
 	}
 
-	globalErrs := field.ErrorList{}
-
 	// Convert plain ingress resources to gateway resources, ignoring all
 	// provider-specific features.
 	gatewayResources, errs := common.ToGateway(ingressList, c.implementationSpecificOptions)
 	if len(errs) > 0 {
-		globalErrs = append(globalErrs, errs...)
+		return i2gw.GatewayResources{}, errs
```
levikobi, line 58: Returning the error early here is a better approach in my opinion, as otherwise we get cascading error which are hard to understand.
Fox example, this was before:
```
# Encountered 4 errors # paths.backends[0].service.port: Invalid value: "name": named ports not supported: http # <nil>: Internal error: HTTPRoute does not exist # <nil>: Internal error: HTTPRoute does not exist # <nil>: Internal error: HTTPRoute does not exist
```

While this is after:
```
# Encountered 1 errors # paths.backends[0].service.port: Invalid value: "name": named ports not supported: http
```

LiorLieberman: Good point. I think it might be better to do it after `crds.TCPIngressToGatewayAPI` though. 
The reason is, that if you fail here, and then there is a problem reading in `TCPIngressToGatewayAPI` then you'd need the second run to know about it. However if you aggregate the errs, the use can fix them both in the first iteration.

Similarly, I think it is useful to aggregate the featureParse errors for the same reason.

LiorLieberman: So my proposal is to aggregate `toGateway` and `TCPIngressToGatewayAPI`, then `MergeGateways` - leave as it is now, and lastly aggregate the featureParser errors, also kinda as it is now - maybe only `globalErrs` can have a better name

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -47,7 +47,7 @@ func canaryFeature(ingressResources i2gw.InputResources, gatewayResources *i2gw.
 			key := types.NamespacedName{Namespace: path.ingress.Namespace, Name: common.RouteName(rg.Name, rg.Host)}
 			httpRoute, ok := gatewayResources.HTTPRoutes[key]
 			if !ok {
-				panic("HTTPRoute not exists - this should never happen")
+				return field.ErrorList{field.InternalError(nil, fmt.Errorf("HTTPRoute does not exist"))}
```
LiorLieberman, line 50: On lines 44-45 we get errors from `calculateBackendRefWeight`. Although you replaced panic with returning an error, the errors above are still invisible and the user can not know whats wrong.

The concern is relevant for changes in Kong as well.

levikobi: That's a good catch. In those cases, instead of panicking or returning uninformative errors such as "HTTPRoute does not exist," we could `continue` to the next iteration. All the informative errors will be collected and returned at the end of the for-loop

WDYT @LiorLieberman @mlavacca?

pkg/i2gw/providers/kong/converter.go
levikobi: sounds good to me

pkg/i2gw/providers/ingressnginx/canary.go
LiorLieberman: `continue` plus a comment on top of this to explain sounds good to me. 
Also since you changed `kong/converter.go` to the logic [proposed](https://github.com/kubernetes-sigs/ingress2gateway/pull/152#discussion_r1558876461), I think is ok to log and return an error because in kong this indeed should never happen.

levikobi: > continue plus a comment on top of this to explain sounds good to me.

Done

> Also since you changed kong/converter.go to the logic https://github.com/kubernetes-sigs/ingress2gateway/pull/152#discussion_r1558876461, I think is ok to log and return an error because in kong this indeed should never happen.

I'm not exactly sure what you mean here

LiorLieberman: for all kong featureParsers changes you have added, (returning error instead of panic) maybe we should add that this should never happen? or something more clear to the user?
Because this indeed should never happen now with the change you added to check the errors and return before calling the featureParsers in `kong/converter.go`

I hope this is a bit more clear

levikobi: OK got it now. I added back the "this should never happen" message. If we indeed encounter this error at some point for whatever reason we'll figure out how to address it better

LiorLieberman:
thanks, 
/lgtm

PR #153, opened 2024-04-10
fail soft when crds are not installed

### LINKED ISSUE #138: Execution fails when CRDs are not installed in the cluster

LiorLieberman:
**What type of PR is this?**
/kind bug

Execution should work regardless of whether CRDs are installed or not

Fixes #138 

```release-note
Fix errors when CRDs are not installed in the cluster
```

### DIFF
--- a/pkg/i2gw/providers/istio/resource_reader.go
+++ b/pkg/i2gw/providers/istio/resource_reader.go
@@ -29,6 +29,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type reader struct {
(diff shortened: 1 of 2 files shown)


mlavacca (commented):
Can we move this check to the caller logic and return a typed error from the provider, instead of having the same logic in each provider? Something like what I suggested [here](https://github.com/kubernetes-sigs/ingress2gateway/pull/144#issuecomment-2027374613)

LiorLieberman:
> Can we move this check to the caller logic and return a typed error from the provider, instead of having the same logic in each provider? Something like what I suggested [here](https://github.com/kubernetes-sigs/ingress2gateway/pull/144#issuecomment-2027374613)

I dont think we should wrap an already `NotFound` error in another custom `NotFound` type. 

Also, I less like the idea of a list of providers that participated in the translation. I think providers can still participate in the translation with ingress annotations and without CRDs installed. 

It is not ideal to have the same logic - hence a helper might be the way but I _think_ we can not take it out to the caller, because we need to continue the execution of the provider even if the CRDs are not installed. 

cc: @levikobi

levikobi (commented):
This looks good to me also TBH

I would have preferred having the `NotFound` check once at the caller level, but the caller doesn't directly call `readTCP...` or `readUDP...` - it calls a method that aggregates the reading of specific CRDs. Since this mid level function that reads everything needs to keep going even if TCP for example isn't installed, we would have to make the check there...

In other words, even though I would have preferred it more, I don't see how we can move the check to the caller and preserve the required functionality

mlavacca (commented):
/approve
/unhold

PR #157, opened 2024-05-04
OpenAPI Provider

### LINKED ISSUE #147: OpenAPI to Gateway API support

Add the ability to convert OpenAPI (v2 and v3) input config into Gateway API config
e.g. https://petstore3.swagger.io

**Why this is needed**:
A lot of users seem to be hitting limits on the existing number of `rules`  ( max 8) 
https://github.com/kubernetes-sigs/gateway-api/issues/1485
Creating a translation by the gateway api community should provide the recommended way to convert user API intent into Gateway API config, reducing the friction users are feeling today due to the limit

guicassolato:
**What type of PR is this?**
/kind feature

Adds new `openapi` provider.

Closes #147

```release-note
- New OpenAPI Provider
```

**Try this PR out**:

```sh
make build
./ingress2gateway print --providers openapi3 --input-file=pkg/i2gw/providers/openapi3/fixtures/input/1-petstore3.yaml
```


guicassolato:
@arkodg @youngnick @LiorLieberman _et al_

Sharing this as a draft for some initial feedback.

Quite a few TODOs ahead still, including:
- [ ] backendRefs
- [ ] Gateways and parentRefs
- [ ] input with multiple OAS docs → Fix route names/namespaces

Apart from more testing of course.

Please check out the test cases already covered under `providers/openapi/fixtures`.

pkg/i2gw/providers/openapi/resource_reader.go
```diff
@@ -0,0 +1,54 @@
...
+	loader := openapi3.NewLoader()
+	spec, err := loader.LoadFromFile(filename)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load OpenAPI spec: %w", err)
+	}
+
+	if err := spec.Validate(ctx); err != nil {
+		return nil, fmt.Errorf("invalid OpenAPI 3.x spec: %w", err)
+	}
+
```
LiorLieberman, line 49: @levikobi @mlavacca  I think we should think about whether we want to preserve this all supported providers by default functionality. 

For example here, if we merge this PR, if someone provides `--input-file` but doesn't provide provider this will always fail (because a regular yaml/json input file would never be a valid openapi spec. 

@guicassolato, regardless to the above, I think running `openapi` provider should always be mutually exclusive to other providers.

pkg/i2gw/providers/openapi/storage.go
```diff
@@ -0,0 +1,47 @@
...
+	resources []*openapi3.T
+}
+
+func newResourceStorage() *storage {
+	return &storage{}
+}
+
+func (s *storage) addResource(resource *openapi3.T) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
LiorLieberman, line 37: why do we need a lock? couldn't find a place where you run `addResource` or `getResource` concurrently

pkg/i2gw/providers/openapi/utils.go
```diff
@@ -0,0 +1,37 @@
...
+	return arr
+}
+
+// Filter filters the input slice using the given predicate function and returns a new slice with the results.
+func Filter[T any](slice []T, f func(T) bool) []T {
+	arr := make([]T, 0)
+	for _, e := range slice {
+		if f(e) {
+			arr = append(arr, e)
+		}
```
LiorLieberman, line 34: Nice! would that be useful to have these utils shared outside of openapi package?

LiorLieberman (commented):
Thanks @guicassolato, that a very useful contribution! left some comments

pkg/i2gw/providers/openapi/README.md
```diff
@@ -0,0 +1,20 @@
+# OpenAPI Provider
+
+The provider translates OpenAPI Specification (OAS) 3.x documents to Kubernetes Gateway API resources Gateway and HTTPRoute.
+
+## Example
+
+```sh
+./ingress2gateway print --providers openapi --input_file=petstore3-openapi.json
```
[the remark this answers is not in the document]
guicassolato: I thought about that. I wasn't sure about maybe adding support to OAS 2.0 (swagger) too. (The lib we're using has auto-detection for it.) On the other hand, there are plenty of tools to convert OAS 2.0 → OAS 3.x IIRC.

Anyway, if anybody else thinks `openapi3` is good, I'm sold.

pkg/i2gw/providers/openapi/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+
+## Known limitations
+
+* Only offline translation supported – i.e. `--input_file` required
+* All API operation [paths](https://swagger.io/specification/v3/#paths-object) treated as `Exact` type – i.e. no support for [path templating](https://swagger.io/specification/v3/#path-templating), therefore no `PathPrefix`, nor `RegularExpression` path types output
```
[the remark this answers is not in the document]
guicassolato: I'm afraid there aren't many good options for that. At least, not that I could think of.

1. Delete the part in each path from the first variable placeholder, all the way to the end of the path, and use `PathPrefix` type → some nasty stuff could happen with the ordering of the route rules
2. Use `RegularExpression` type → not core; what flavour of regex would we use?

Path templates are not like server variables that in some case give us at least a closed set of possible values (enum.)

pkg/i2gw/providers/openapi/storage.go
guicassolato: It's a good practice in general. Maps are not thread-safe by default. Since the dependency is injected, the converter cannot prevent external usage of the storage.

In fact, I now realise my implementation is poor. A concurrent call to `ReadResourcesFromFile` would cause the entire storage to be invalidated while completely bypassing the lock, when arguably the right thing to do would be to use the context to infer whether reading a resource should be additive to the storage or clear it first. At the very least, do the latter, but not forcibly recreate the storage. I will fix that.

From the perspective of the `openapi` package, there is no way to tell what a client will do with those two competing processes.

pkg/i2gw/providers/openapi/utils.go
guicassolato: Happy to move them to the `common` package if you say it's worthwhile.

pkg/i2gw/providers/openapi/resource_reader.go
guicassolato: What's the use case for converting to all supported providers anyway?

pkg/i2gw/providers/openapi/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Example
+
+```sh
+./ingress2gateway print --providers openapi --input_file=petstore3-openapi.json
+```
+
+## Known limitations
+
+* Only offline translation supported – i.e. `--input_file` required
```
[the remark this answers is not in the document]
guicassolato: Done.

pkg/i2gw/providers/openapi/README.md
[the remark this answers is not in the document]
guicassolato: Renamed the package to `openapi3`.

pkg/i2gw/providers/openapi/openapi.go
```diff
@@ -0,0 +1,70 @@
...
+		storage:   newResourceStorage(),
+		reader:    newResourceReader(conf),
+		converter: newConverter(conf),
+	}
+}
+
+// ReadResourcesFromCluster reads OpenAPI specs stored in the Kubernetes cluster. UNIMPLEMENTED.
+func (p *Provider) ReadResourcesFromCluster(_ context.Context) error {
+	return fmt.Errorf("provider does not support reading resources from cluster")
+}
```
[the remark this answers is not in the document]
guicassolato: Done.

pkg/i2gw/providers/openapi/utils.go
guicassolato: Moved the funcs to `provider/common`.

guicassolato:
I want to mark this as ready for review, but not before coming to a decision about how to proceed for `backendRefs` and `gatewayClassName`.

It was [suggested](https://github.com/kubernetes-sigs/ingress2gateway/issues/147#issuecomment-2087804104) to use command-line args for those. I would like to know how people feel about that?

A few questions to drive the conversation next should we say this indeed is the desired approach:

- Should I just define those as flags of the `print` command?
- Since they don't have meaning for other providers converting from Ingress resources, will providers like istio, ingress-nginx, kong, etc ignore the values when supplied?
- Perhaps call those options `--default-backend-name` and `--default-gateway-controller-name`?
- Or they could be namespaced specifically for this provider – e.g. `--openapi3-bakcend-name` and  `--openapi3-gateway-controller-name` – tho I can see a possible mess in the future should other providers ask for other similar options. If it is provider-specific, should the options be _required_ when `--providers` includes `openapi3` and _forbidden_ otherwise?
- Provider-specific options could also be serialised into a single command-line flag (e.g. `--provider-options "backend-name=x,controller-name=y"`)
- Is it starting to get a little bit too much to pass in the call to [`ToGatewayAPIResources`](https://github.com/kubernetes-sigs/ingress2gateway/blob/11f5ee2f1edbd8acefd8372e146ac195e7917be6/cmd/print.go#L78)? Should we change the API to accept a more generic `PrintOptions` type?

cc @youngnick @arkodg @LiorLieberman

pkg/i2gw/providers/common/utils.go
```diff
@@ -200,3 +200,23 @@ func removeBackendRefsDuplicates(backendRefs []gatewayv1.HTTPBackendRef) []gatew
...
+// Filter filters the input slice using the given predicate function and returns a new slice with the results.
+func Filter[T any](slice []T, f func(T) bool) []T {
+	arr := make([]T, 0)
+	for _, e := range slice {
+		if f(e) {
+			arr = append(arr, e)
+		}
+	}
+	return arr
+}
```
mlavacca, line 222: Can we use https://github.com/samber/lo instead of maintaining our own version of these utils?

pkg/i2gw/providers/openapi3/openapi.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// ToGatewayAPI converts stored OpenAPI specs to Gateway API resources.
+func (p *Provider) ToGatewayAPI(_ i2gw.InputResources) (i2gw.GatewayResources, field.ErrorList) {
+	return p.converter.Convert(p.storage)
+}
+
+func init() {
+	i2gw.ProviderConstructorByName[ProviderName] = NewProvider
+}
```
mlavacca, line 75: Can we move this init to the top of the file, right before 
```
type Provider struct {
	storage   Storage
	converter Converter
}
```
?

pkg/i2gw/providers/openapi3/storage.go
```diff
@@ -0,0 +1,63 @@
...
+	Clear()
+}
+
+// NewResourceStorage returns a thread-safe storage for OpenAPI specs.
+func NewResourceStorage() Storage {
+	return &storage{}
+}
+
+type storage struct {
+	mu sync.RWMutex
```
mlavacca, line 37: Why do we need thread safety here? I haven't seen any concurrency in the code, did I miss something?

pkg/i2gw/providers/openapi3/converter.go
```diff
@@ -0,0 +1,465 @@
...
+var _ Converter = &converter{}
+
+func (c *converter) Convert(storage Storage) (i2gw.GatewayResources, field.ErrorList) {
+	gatewayResources := i2gw.GatewayResources{
+		Gateways:        make(map[types.NamespacedName]gatewayv1.Gateway),
+		HTTPRoutes:      make(map[types.NamespacedName]gatewayv1.HTTPRoute),
+		TLSRoutes:       make(map[types.NamespacedName]gatewayv1alpha2.TLSRoute),
+		TCPRoutes:       make(map[types.NamespacedName]gatewayv1alpha2.TCPRoute),
+		ReferenceGrants: make(map[types.NamespacedName]gatewayv1beta1.ReferenceGrant),
+	}
```
mlavacca, line 70: Do we really need to perform this initialization here?

mlavacca:
> I want to mark this as ready for review, but not before coming to a decision about how to proceed for `backendRefs` and `gatewayClassName`.
> 
> It was [suggested](https://github.com/kubernetes-sigs/ingress2gateway/issues/147#issuecomment-2087804104) to use command-line args for those. I would like to know how people feel about that?
> 
> A few questions to drive the conversation next should we say this indeed is the desired approach:
> 
> * Should I just define those as flags of the `print` command?
> * Since they don't have meaning for other providers converting from Ingress resources, will providers like istio, ingress-nginx, kong, etc ignore the values when supplied?
> * Perhaps call those options `--default-backend-name` and `--default-gateway-controller-name`?
> * Or they could be namespaced specifically for this provider – e.g. `--openapi3-bakcend-name` and  `--openapi3-gateway-controller-name` – tho I can see a possible mess in the future should other providers ask for other similar options. If it is provider-specific, should the options be _required_ when `--providers` includes `openapi3` and _forbidden_ otherwise?

+1 on this approach. having access to generic `--default-backend-name`, `--default-gateway-controller-name` flags might be confusing for users. I'd prefer having `openapi3-`-prefixed command line flags to avoid such a confusion. openapi is a special case, and I do cannot think of many other such special cases in the future

> * Provider-specific options could also be serialised into a single command-line flag (e.g. `--provider-options "backend-name=x,controller-name=y"`)
> * Is it starting to get a little bit too much to pass in the call to [`ToGatewayAPIResources`](https://github.com/kubernetes-sigs/ingress2gateway/blob/11f5ee2f1edbd8acefd8372e146ac195e7917be6/cmd/print.go#L78)? Should we change the API to accept a more generic `PrintOptions` type?

+1 on introducing a `PrintOptions` struct to pass all the needed parameters.

mlavacca (commented):
Thanks for this PR! I left some initial generic comments in the review

pkg/i2gw/providers/openapi3/storage.go
guicassolato: I tried to answer it here: https://github.com/kubernetes-sigs/ingress2gateway/pull/157#discussion_r1590105256

TL;DR - I think not having it is a road to problems in the future.

Not supposed to sound unnecessarily ominous here... I do believe things change, such as – I'm making stuff up here, ok? – start supporting multiple input files (or an input directory equivalently) and, if so, why not calling `ReadResourcesFromFile` concurrently?

It's a small price to pay IMO, but if more people believe we're better off _without_ the mutex, I can remove it – even if I don't understand why.

pkg/i2gw/providers/openapi3/converter.go
guicassolato: For `TLSRoutes`, `TCPRoutes` and `ReferenceGrants` probably not.

Nil maps are safe to read and it kinds of make sense to me that the caller will not try to write back on the `GatewayResources` struct without checking first. Well, if it does, it's the caller's problem anyway.

Besides, returning `nil` for `TLSRoutes`, `TCPRoutes` and `ReferenceGrants` seems semantically the right thing to do, I reckon. I'll fix it.

Thanks!

arkodg:
hey reg naming, i'd vote for
* `<openapi provider prefix>-backend` - openapi provider specific CLI arg
* `gatewayc-controller` - common to the entire CLI arg

guicassolato:
Where I [wrote](https://github.com/kubernetes-sigs/ingress2gateway/pull/157#issuecomment-2100109817) `controller-name` please read **`gateway-class-name`**, since obviously we're talking about generating Gateways, not GatewayClasses. Apologies for my distraction before.

guicassolato:
I was feeling somewhat uncomfortable with hard-coding provider-specific CLI flags when there's already a mechanism in place to register providers more dynamically. So please let me know what you think about https://github.com/kubernetes-sigs/ingress2gateway/pull/157/commits/12dd6a09cceb21f537a2c134e01d64f012665630.

I'm proposing a function to declare and a method to retrieve those provider-specific user input values, that keeps the common implementation clean from it. Explanation and example provided in `PROVIDER.md`.

I'm using the method to register `--openapi3-backend` and `--openapi3-gateway-class-name` both as specific to the `openapi3` provider. Example of how the flags will show in command help:

```
$ ./ingress2gateway print --help
Prints Gateway API objects generated from ingress and provider-specific resources.

Usage:
  ingress2gateway print [flags]

Flags:
  -A, --all-namespaces                       If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
                                             if specified with --namespace.
  -h, --help                                 help for print
      --input-file string                    Path to the manifest file. When set, the tool will read ingresses from the file instead of reading from the cluster. Supported files are yaml and json.
  -n, --namespace string                     If present, the namespace scope for this CLI request.
      --openapi3-backend string              Provider-specific: openapi3. The name of the backend service to use in the HTTPRoutes
      --openapi3-gateway-class-name string   Provider-specific: openapi3. The name of the gateway class to use in the Gateways
  -o, --output string                        Output format. One of: (json, yaml). (default "yaml")
      --providers strings                    If present, the tool will try to convert only resources related to the specified providers, supported values are [apisix ingress-nginx istio kong openapi3]. (default [apisix,ingress-nginx,istio,kong,openapi3])

Global Flags:
      --kubeconfig string   The kubeconfig file to use when talking to the cluster. If the flag is not set, a set of standard locations can be searched for an existing kubeconfig file.
```

@arkodg, if you think `gateway-class-name` is useful to all providers, I'm happy to hard-code this one as we normally would and remove it from `openapi3`-specific, even if for now only `openapi3` will be doing something with it.

arkodg:
@guicassolato no strong feelings, the project maintainers have a much better understand of naming and should decide

guicassolato:
> @guicassolato no strong feelings, the project maintainers have a much better understand of naming and should decide

cc @LiorLieberman @mlavacca

pkg/i2gw/providers/openapi/resource_reader.go
LiorLieberman: thats how we started, I think we need to change it, but will have a separate issue for that

LiorLieberman:
Thanks again for driving this Gui! 
Regarding `gateway-class-name` I hear you, I think what we ultimately want to do is to have IR (intermediate representation) and have `--input-provider` and `--output-provider`. Each provider would implement old config --> IR and IR --> Gateway configs (+gateway extensions). That at least what @arkodg and I discussed at KubeCon. 

If we do plan move on to this logic, I think `gateway-class-name` will be derived from the `--output-provider` flag. 

Since we dont support it now, an nor gateway extensions are supported, I am trying to think whether it will indeed be useful for other providers. The more I think about it - I get to the conclusion that we must move to input and output providers separately. **So I'd vote leaving `gateway-class-name`  prefixed**  


Regarding the backend flag - still not sure how you were going to represent it in a flag value? A formatted string with name,ns,port? (not loving this either)

About the general approach of registering provider specific flags dynamically. I like it but trying to think what are the use-cases for other providers.
We have to ask ourselves how different the `openapi` provider from all other providers PLUS how much of the logic is similar, maybe we want to have a subcommand for openapi which will make things simpler? not saying we should but just throwing it some thought 

Lastly I left a few random comments on https://github.com/kubernetes-sigs/ingress2gateway/commit/12dd6a09cceb21f537a2c134e01d64f012665630. 

Thats really awesome, thanks again!

guicassolato:
Hi @LiorLieberman. Thank you so much for taking the time to review this.

It turns out the biggest effort here has not been converting OAS to Gateway API _per se_, but the details... where, as we all know, lives the Devil. The PR has grown bigger and I still hope to be able to deliver valuable contribution with it.

A few comments in line...

<br/>

> Thanks again for driving this Gui! Regarding `gateway-class-name` I hear you, I think what we ultimately want to do is to have IR (intermediate representation) and have `--input-provider` and `--output-provider`. Each provider would implement old config --> IR and IR --> Gateway configs (+gateway extensions). That at least what @arkodg and I discussed at KubeCon.

I like the idea of IR. Of course, that only works if the abstraction in the middle is capable of absorbing all the complexity from all possible inputs/"old configs". In fact, the main reason why I didn't use Ingress as IR – and therefore do `OAS -> Ingress -> Gateway API`, relying on the standard ("common") conversion functions – was not being able to support capabilities that exist in the old config (OAS, in this case) but not in the IR, such as header and query matchers.

To be completely honest, I also wanted to avoid generating invalid HTTPRoutes with more rules or more matches per rule than it can take, without introducing breaking changes to existing providers. To give an example, even the [Swagger Petstore API](https://petstore3.swagger.io/) would activate this problem.

I believe a straight `OAS -> Gateway API` conversion for now allowed me solving both issues: capabilities and conformance.

<br/>

> If we do plan move on to this logic, I think `gateway-class-name` will be derived from the `--output-provider` flag.

Yes, I can see how that could work for `gateway-class-name`.

<br/>

> Since we dont support it now, an nor gateway extensions are supported, I am trying to think whether it will indeed be useful for other providers. The more I think about it - I get to the conclusion that we must move to input and output providers separately. **So I'd vote leaving `gateway-class-name` prefixed**

So, to be clear then, you're aligned with @mlavacca and with how I ended up implementing it, i.e. `openapi3-gateway-class-name`?

<br/>

> Regarding the backend flag - still not sure how you were going to represent it in a flag value? A formatted string with name,ns,port? (not loving this either)

Currently, it's `namespace/name`, no support for port number. We could instead establish a convention to use `$SERVICE.$NAMESPACE.svc.cluster.local:$PORT`, for example. Would that be preferable?

<br/>

> About the general approach of registering provider specific flags dynamically. I like it but trying to think what are the use-cases for other providers. We have to ask ourselves how different the `openapi` provider from all other providers PLUS how much of the logic is similar, maybe we want to have a subcommand for openapi which will make things simpler? not saying we should but just throwing it some thought

I was hoping the example provided in `PROVIDER.md` would convince you 🙂 It's a different use case than OpenAPI.

By that I don't mean to disagree that `openapi3` is a special case, because it definitely is. To begin with, it doesn't count with a few crucial pieces of information that Ingress does, such as names of the backends, names of the gateway TLS certificate secrets, which gateway controller ultimately will be responsible for the generated Gateway API objects... And I'm sure we'll find others. However, I believe OAS to Gateway API is not special in the sense that, sooner or later, all providers will need something that is specific to its case.

I can ramble for several pages about the truth behind this last statement, but instead I'll turn to my other point of motivation for having flags dynamically registered. I'm concerned about the leakage of a provider specific detail hard-coded at the level that is supposed to abstract that. It feels like violating a principle, apart from opening a dangerous precedent IMO.

If you are sure we won't regret it (with the information we know now), I will keep it simple and hard code `--openapi3-gateway-class-name`, `--openapi3-backend` and (_heads-up!_) `--openapi3-gateway-tls-secret` directly inside [`func newPrintCommand() {...}`](https://github.com/kubernetes-sigs/ingress2gateway/blob/825f587a0de41e52a0acfbe352008376d7f4577c/cmd/print.go#L223). Otherwise, the work is done and nothing to worry about.

<br/>

> Lastly I left a few random comments on [12dd6a0](https://github.com/kubernetes-sigs/ingress2gateway/commit/12dd6a09cceb21f537a2c134e01d64f012665630).

Thanks for those too! Both valid points that I will be addressing next.

LiorLieberman:
> I believe a straight OAS -> Gateway API conversion for now allowed me solving both issues: capabilities and conformance.

Right, to clarify, I did not mean that you should have done OAS-->Ingress-->Gateway, instead I am saying that in future extensions we should do OAS-->IR, Istio CRDs -->IR, Ingress-->IR, and have output providers from the other side do IR-->provider

> So, to be clear then, you're aligned with @mlavacca and with how I ended up implementing it, i.e. openapi3-gateway-class-name?

👍

> Currently, it's namespace/name, no support for port number. We could instead establish a convention to use $SERVICE.$NAMESPACE.svc.cluster.local:$PORT, for example. Would that be preferable?

hmm not sure, why should a user include `svc.cluster.local`? it seems like we only have to support namespace, name, and port, right? so I am good with whatever convention you decide here as long as you document it.

Regarding the last bullet, I think what you added is great, so definitely good with keeping it. 

Are there any more blockers/things we need to figure out before we can merge this ?

guicassolato:
>> Currently, it's namespace/name, no support for port number. We could instead establish a convention to use $SERVICE.$NAMESPACE.svc.cluster.local:$PORT, for example. Would that be preferable?
> hmm not sure, why should a user include `svc.cluster.local`? it seems like we only have to support namespace, name, and port, right? so I am good with whatever convention you decide here as long as you document it.

I guess I was just trying not to come up with yet another convention. But you're probably right. Why requiring users to know the cluster domain if we don't actually need it?

I'll include the option for the port number. Let's make it `[namespace/]name[:port]`. Examples of valid values:
- `--openapi3-backend=my-service`
- `--openapi3-backend=my-namespace/my-service`
- `--openapi3-backend=my-service:3000`
- `--openapi3-backend=my-namespace/my-service:3000`

> Are there any more blockers/things we need to figure out before we can merge this ?

After the above, I think we're good.

<br/>

Thanks @LiorLieberman!

LiorLieberman:
Awesome, before my final review. 

> Quite a few TODOs ahead still, including:
> 
> * [x]  Gateways and parentRefs
> * [ ]  backendRefs
> * [ ]  gatewayClass
> * [ ]  Name objects after the OAS title (?)
> * [ ]  Input with multiple OAS docs (?)
> 
> Apart from more testing of course.
> 
> Please check out the test cases already covered under `providers/openapi/fixtures`.


Is this up to date? I think you already took care of some unchecked stuff here (like gwc for example) 
Can you please update the comment whats already addressed and whats not, and then add TODOs (future/unsupported) things in provider.MD and ping me for final review?

LiorLieberman (commented):
Awesome, before my final review. 

> Quite a few TODOs ahead still, including:
> 
> * [x]  Gateways and parentRefs
> * [ ]  backendRefs
> * [ ]  gatewayClass
> * [ ]  Name objects after the OAS title (?)
> * [ ]  Input with multiple OAS docs (?)
> 
> Apart from more testing of course.
> 
> Please check out the test cases already covered under `providers/openapi/fixtures`.


Is this up to date? I think you already took care of some unchecked stuff here (like gwc for example) 
Can you please update the comment whats already addressed and whats not, and then add TODOs (future/unsupported) things in provider.MD and ping me for final review?

guicassolato:
> Awesome, before my final review.
> 
> > Quite a few TODOs ahead still, including:
> > 
> > * [x]   Gateways and parentRefs
> > * [ ]   backendRefs
> > * [ ]   gatewayClass
> > * [ ]   Name objects after the OAS title (?)
> > * [ ]   Input with multiple OAS docs (?)
> > 
> > Apart from more testing of course.
> > Please check out the test cases already covered under `providers/openapi/fixtures`.
> 
> Is this up to date? I think you already took care of some unchecked stuff here (like gwc for example) Can you please update the comment whats already addressed and whats not, and then add TODOs (future/unsupported) things in provider.MD and ping me for final review?

I've updated the comment, @LiorLieberman.

I think the only item not covered in this iteration is the one of input files declaring multiple OAS docs, which I prefer leaving for a future improvement, if ever requested.

The limitation to a maximum of 1 OAS per input file is already mentioned in provider.MD.

PROVIDER.md
```diff
@@ -152,17 +152,45 @@ import (
...
 ### Testing the feature parser
 There are 2 main things that needs to be tested when creating a feature parser:
 1. The conversion logic is actually correct.
 2. The new function doesn't override other functions modifications.
 For example, if one implemented the mirror backend feature and it deletes canary weight from `BackendRefs`, we have a
-problem.
\ No newline at end of file
+problem.
+
+## Provider-specific user input
```
LiorLieberman, line 170: maybe Provider-specific configurations? or Provider-specific flags?

cmd/print.go
```diff
@@ -59,6 +61,9 @@ type PrintRunner struct {
 
 	// providers indicates which providers are used to execute convert action.
 	providers []string
+
+	// Provider specific conf. Value assigned via provider specific flags --<provider>-<conf>.
+	providerSpecificConf map[string]*string
```
LiorLieberman, line 66: same comment, I think there is already `conf` field in the Provider struct, maybe its better to call it providerSpecificFlags ?

cmd/print.go
```diff
@@ -262,3 +275,19 @@ func getNamespaceInCurrentContext() (string, error) {
 
 	return currentNamespace, err
 }
+
+func (pr *PrintRunner) getProviderSpecificConf() map[string]map[string]string {
```
LiorLieberman, line 279: can you please add a function comment so its easier for other folks to read?

cmd/print.go
```diff
@@ -262,3 +275,19 @@ func getNamespaceInCurrentContext() (string, error) {
...
+	providerSpecificConf := make(map[string]map[string]string)
+	for flagName, value := range pr.providerSpecificConf {
+		provider, found := lo.Find(pr.providers, func(p string) bool { return strings.HasPrefix(flagName, p+"-") })
```
LiorLieberman, line 282: nit: can you use `fmt.Sprintf` instead of `+` ? ditto for line 286 ?

cmd/print.go
```diff
@@ -262,3 +275,19 @@ func getNamespaceInCurrentContext() (string, error) {
...
+		if !found {
+			continue
+		}
+		conf := strings.SplitN(flagName, provider+"-", 2)[1]
```
LiorLieberman, line 286: nit: can we change `conf` to something more meaningful? maybe `confName` or even more meaningful - `flagNameWithoutProvider`. These are not great either, so if you come up with something more creative thats better

Also, consider just trimming the prefix -  `strings.TrimPrefix(flagName, provider+"-")`

cmd/print_test.go
```diff
@@ -236,3 +236,60 @@ func Test_getNamespaceInCurrentContext(t *testing.T) {
...
+				"provider-b-conf2": &value2,
+			},
+			providers: []string{"provider-a", "provider-b", "provider-c"},
+			expected: map[string]map[string]string{
+				"provider-a": {"conf1": value1},
+				"provider-b": {"conf2": value2},
+			},
+		},
+		{
+			name:                 "Provider specific configuration not matching provider in the list",
```
LiorLieberman, line 276: I liked that you added this test. Should we `log.Warn` in if such case happen? (I mean in the library code, not in the test)

cmd/print_test.go
```diff
@@ -236,3 +236,60 @@ func Test_getNamespaceInCurrentContext(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			pr := PrintRunner{
+				providerSpecificConf: tc.providerSpecificConf,
+				providers:            tc.providers,
+			}
+			actual := pr.getProviderSpecificConf()
+			if !reflect.DeepEqual(actual, tc.expected) {
+				t.Errorf("getProviderSpecificConf() = %v, expected %v", actual, tc.expected)
```
LiorLieberman, line 291: nit: consider using cmp.Diff instead to indicate the diff nicely

pkg/i2gw/ingress2gateway.go
```diff
@@ -47,8 +47,9 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 	}
 
 	providerByName, err := constructProviders(&ProviderConf{
-		Client:    clusterClient,
-		Namespace: namespace,
+		Client:           clusterClient,
+		Namespace:        namespace,
+		ProviderSpecific: providerSpecificConf,
```
LiorLieberman, line 52: I'd already commented on this [here](https://github.com/kubernetes-sigs/ingress2gateway/commit/12dd6a09cceb21f537a2c134e01d64f012665630#r141876286)

I still think we need to change that, not sure I understood your point of view from the answer there. Maybe providerSpecificFlags? ExtraFlags?

pkg/i2gw/provider.go
```diff
@@ -105,3 +107,35 @@ type GatewayResources struct {
...
+
+type ProviderSpecificConf struct {
+	Name         string
+	Description  string
+	DefaultValue string
+}
+
+var (
+	providerSpecificConfs      = map[ProviderName]map[string]ProviderSpecificConf{}
+	providerSpecificConfsMutex = sync.RWMutex{}
```
LiorLieberman, line 119: Why mutex as a global var instead of wrapping `map[ProviderName]map[string]ProviderSpecificConf{}` as a new type and add the mutex within it?

pkg/i2gw/provider.go
```diff
@@ -105,3 +107,35 @@ type GatewayResources struct {
...
+)
+
+// RegisterProviderSpecificConf registers a provider-specific conf.
+// Each provider-specific conf is exposed to the user as a command-line flag, defined as optional.
+// If the flag is not provided, it is up for the provider to decide to use the default value or raise an error.
+// The provider can read the values of provider-specific confs input by the user from the ProviderConf,
+// prefixed by the provider name.
```
LiorLieberman, line 126: Thanks, can you add a line, explaining why you use the lock here. I want to save other folks time and searching the codebase for concurrency.

pkg/i2gw/providers/openapi3/converter.go
```diff
@@ -0,0 +1,619 @@
...
+	HostWildcard   = "*"
+	HostSeparator  = ","
+	ParamSeparator = ","
+
+	HTTPRouteRulesMax      = 16
+	HTTPRouteMatchesMax    = 8
+	HTTPRouteMatchesMaxMax = HTTPRouteRulesMax * HTTPRouteMatchesMax
+)
+
+var uriRegexp = regexp.MustCompile(`^((https?)://([^/]+))?(/.*)?$`) // [_][1] = scheme, [_][2] = host, [_][3] = path
```
LiorLieberman, line 51: can you explain the comment ?

pkg/i2gw/providers/openapi3/converter.go
```diff
@@ -0,0 +1,619 @@
...
+
+// NewConverter returns a converter of OpenAPI Specifications 3.x from a storage into Gateway API resources.
+func NewConverter(conf *i2gw.ProviderConf) Converter {
+	converter := &converter{
+		namespace:    conf.Namespace,
+		tlsSecretRef: types.NamespacedName{},
+		backendRef:   toBackendRef(""),
+	}
+
+	if ps := conf.ProviderSpecific[ProviderName]; ps != nil {
```
LiorLieberman, line 65: what ensures that all those flags are set if ps != nil?

pkg/i2gw/providers/openapi3/converter.go
```diff
@@ -0,0 +1,619 @@
...
+	resourcesNamePrefixes := make(map[string]int)
+
+	for _, spec := range storage.GetResources() {
+		resourcesNamePrefix := toResourcesNamePrefix(spec)
+		if _, exists := resourcesNamePrefixes[resourcesNamePrefix]; !exists {
+			resourcesNamePrefixes[resourcesNamePrefix] = 0
+		}
+		resourcesNamePrefixes[resourcesNamePrefix]++
+		if resourcesNamePrefixes[resourcesNamePrefix] > 1 {
+			resourcesNamePrefix = fmt.Sprintf("%s-%d", resourcesNamePrefix, resourcesNamePrefixes[resourcesNamePrefix]+1)
```
LiorLieberman, line 105: can we add logs and/or comments here to explain the logic?

It will also be useful for @Devaansh-Kumar project to build notifications system for conversion results.

LiorLieberman: similar comments for the the helper functions below, adding a line explaining when we you take a not obvious decisions will be very useful

pkg/i2gw/providers/openapi3/openapi.go
```diff
@@ -0,0 +1,111 @@
...
+func readSpecFromFile(ctx context.Context, filename string) (*openapi3.T, error) {
+	loader := openapi3.NewLoader()
+	spec, err := loader.LoadFromFile(filename)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load OpenAPI spec: %w", err)
+	}
+
+	if err := spec.Validate(ctx); err != nil {
+		log.Printf("%s provider: invalid OpenAPI 3.x spec: %v", ProviderName, err)
+		return nil, nil
```
LiorLieberman, line 107: why nil and not error?

pkg/i2gw/providers/openapi3/storage.go
```diff
@@ -0,0 +1,63 @@
...
+
+	s.resources = append(s.resources, resource)
+}
+
+func (s *storage) GetResources() []*openapi3.T {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.resources
+}
```
LiorLieberman, line 56: Had a similar comment above, can you please add a line why you lock it, so people wont search the codebase for concurrency ?

LiorLieberman (commented):
Thanks @guicassolato ! 
added my comments, mostly nits.

pkg/i2gw/providers/openapi3/converter.go
guicassolato: They don't have to. `toNamespacedName` and `toBackendRef` funcs will fallback to the defaults, which are analogous in effect to an empty `gatewayClassName`.

IOW, we need this nil check not to panic, but we don't need to check each individual flag.

pkg/i2gw/providers/openapi3/converter.go
guicassolato: Sure. I can add the comment.

Out of curiosity though, what would a more obvious decision here be in your opinion?

cmd/print_test.go
guicassolato: For now and following the common implementation, I've added a simple `log.Printf("Warning: …")`, from the std lib. But I've seen providers (Istio, Kong) relying on [k8s.io/klog/v2](https://pkg.go.dev/k8s.io/klog/v2@v2.110.1). Do we want to make a call on this?

LiorLieberman: > Do we want to make a call on this?
Sounds good to me

pkg/i2gw/providers/openapi3/converter.go
guicassolato: I've added comments. Please let me know if we want logs as well, in which case I'd probably also expect some definition regarding https://github.com/kubernetes-sigs/ingress2gateway/pull/157#discussion_r1608602572, to avoid multiple formats of log messages.

LiorLieberman: sorry, I did not meant that your choice was not obvious as "I would choose something else". But when a user is invoking the tool, I am not sure it will be straight forward to them. 
@Devaansh-Kumar is working on a notification system, so he might consider this as an INFO notification,  and comments in the code will help him.
Hopefully this makes sense. 

No need for logs.

pkg/i2gw/providers/openapi3/openapi.go
LiorLieberman: What I meant in https://github.com/kubernetes-sigs/ingress2gateway/pull/157#discussion_r1590001944 is the latter. And yes, for that comment it was important that `ReadResourcesFromFile` would not return error, otherwise the tool will fail for all `providers`. 

However, I opened #158 to change the default mode to choose no providers. Will soon get to do this. 

So I think for now you should make `readSpecFromFile` return an error, log this error in `ReadResourcesFromFile` and return nil. Once I make a PR for #158 I will change that to return the error.

LiorLieberman (commented):
I think I answered all the open comments. 

This looks good to me, going to approve it and I asked @mlavacca to review and leave final lgtm. 

Thanks Gui! 
/approve

pkg/i2gw/providers/openapi3/converter.go
guicassolato: No problem, @LiorLieberman! You're absolutely right about making things clear. What's obvious to some people may not be to others indeed. I'm glad you brought it up. Thanks!

mlavacca (approved):
Looks great to me, thanks @guicassolato!

/ok-to-test
/approve
/lgtm

pkg/i2gw/providers/openapi3/openapi.go
LiorLieberman: Did you implemented this?

guicassolato: I guess I missed this one. Doing it now...

guicassolato: Done

guicassolato: Oops. Forgot to fix the tests. Doing it now.

guicassolato: Now done.

mlavacca (approved):
/lgtm

I'll let @LiorLieberman take a final look and remove the unhold the PR

LiorLieberman:
@guicassolato can you change the release notes to be slightly more detailed? It will make my life easier when releasing a new version

PR #159, opened 2024-06-05
default no providers, mark providers required

### LINKED ISSUE #158: Changing default behavior to no providers instead of all-providers
Default behavior needs to change to not include any provider by default

**Why this is needed**:
Including all provider by default causes us many problems. Two example are https://github.com/kubernetes-sigs/ingress2gateway/issues/127 and https://github.com/kubernetes-sigs/ingress2gateway/issues/138. 

Also, most people would only have have one provider, and in case of more providers installed, they can run the tool multiple times.

/help
/triage accepted

LiorLieberman:
**What type of PR is this?**

/kind feature

Fixes #158 

```release-note
providers flag is now required
```

/cc @mlavacca


cmd/print.go
```diff
@@ -284,7 +284,7 @@ func (pr *PrintRunner) getProviderSpecificFlags() map[string]map[string]string {
 	for flagName, value := range pr.providerSpecificFlags {
 		provider, found := lo.Find(pr.providers, func(p string) bool { return strings.HasPrefix(flagName, fmt.Sprintf("%s-", p)) })
 		if !found {
-			log.Printf("Warning: Ignoring flag %s as it does not match any of the providers", flagName)
+			// log.Printf("Warning: Ignoring flag %s as it does not match any of the providers", flagName)
```
LiorLieberman, line 287: thats noisy as there are some default flags that are configured. I'd like to have something like that for flags that the user explicitly provided.

@guicassolato what do you think?

guicassolato: That makes total sense, @LiorLieberman.

The warning was added in the best intention of letting the user know a provider-specific flag was supplied out of context, but if it has the potential to create more confusion than actually help, then we should probably skip it.

cmd/print.go
```diff
@@ -263,6 +262,7 @@ if specified with --namespace.`)
 		}
 	}
 
+	_ = cmd.MarkFlagRequired("providers")
```
mlavacca, line 265: I think we can add validation that ensures openapi is the only provider when set.

LiorLieberman: done, PTAL!

cmd/print.go
mlavacca: can we remove this line altogether, instead of commenting it?

mlavacca (commented):
Apart from a minor nit, lgtm

LiorLieberman:
mind leaving lgtm?

Issue #131, opened 2024-02-08
Build a notification system/process to report conversion results


LiorLieberman:
**What would you like to be added**:
A new notification system/mechanism/process to notify users conversion results, warnings and crucial messages regarding their conversion.

Ideally, this should be a generic notification mechanism that is able to support different types of notifications/messages and present them nicely. 

**Why this is needed**:
As we onboard more implementations and add more logic, the logs are not sufficient to report such a thing. And in order to improve the usability users should know exactly the original resource/s -> new resource/s mapping plus what the tool couldn't convert so they could make the final touches before applying the new resources to the cluster.

Among the things it can include are:
- Overview of original resources and the equivalent resources they have been converted to
- Warnings on partial conversions (partial fields within a resource)
- Warnings in case a new resource does not(or can not) retain the exact same semantics. For example a different default redirect code, semantics around `*` and empty host fields, etc.
- Warnings about fields/logics not supported in Gateway API
- Warnings about fields/logics not supported by the tool BUT are supported by Gateway API

Ishad-M-I-M:
Hello @LiorLieberman @mlavacca, I'm intereseted on working for this Idea for GSoC'24. I started learning golang and somewhat familiar with kubernetes. I appriciate if you can provide further resources and instructions on how to get started with this project.
Thank you in advance!.

Devaansh-Kumar:
Hello everyone, I am Devaansh Kumar and I also interested in applying for this project under GSoC'24. I have a basic knowledge of golang and would love to discuss the requirements of the project.

YIHONG-JIN:
Hi there, I am Yihong Jin, and interested in working on this issue as well. I have experience in Golang and have used ingress2gateway at work and research. Is this project supposed to be an incremental patch to ingress2gateway which enhance its usability? 

Regarding the notification mechanism, are we trying to build an active mechanism (forced manual check) or a passive mechanism (aggregate the warnings and generate a report)?

LiorLieberman:
Thank you all for expressing your interest. 
Brief requirements are more or less what appear in the description. 

Some useful resources you could start with are:
1. [Familiarize yourself with Gateway API ](https://gateway-api.sigs.k8s.io/)
2. If you are not familiar with ingress, read about this it [here](https://kubernetes.io/docs/concepts/services-networking/ingress/)
[...]

PR #160, opened 2024-06-06
Add basic support for notifications package

Devaansh-Kumar:
**What type of PR is this?**

/kind feature
/kind documentation

This PR provides a skeleton structure for the notifications package

This pull request addresses part of the issue described in #131 

```release-note
Created a new notifications package to provide mechanism for providers to display useful information generated during conversion process
```
/cc @LiorLieberman @mlavacca


pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,74 @@
...
+	Provider string
+}
+
+type NotificationAggregator struct {
+	Notifications []Notification
+}
+
+var CommonNotification NotificationAggregator
+
+func (na *NotificationAggregator) DispatchNotication(notification Notification) {
```
mlavacca, line 49:
```suggestion
func (na *NotificationAggregator) DispatchNotification(notification Notification) {
```

All the callers need to be changed accordingly

pkg/i2gw/notifications/notifications.go
mlavacca, line 1: All the public symbols declared here need a proper comment

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,74 @@
...
+	// Segregate messages into
+	for _, msg := range na.Notifications {
+		providerNotifications[msg.Provider] = append(providerNotifications[msg.Provider], msg)
+	}
+
+	for provider, msgs := range providerNotifications {
+		t := newTableConfig()
+
+		t.SetTitle(fmt.Sprintf("Notifications from %v", provider))
+		t.AppendHeader(table.Row{"Provider", "Message Type", "Notification"})
```
mlavacca, line 66: I'd stick with either the "message" or the "notification" naming, which means "Notification type | notification" or "message type | message".

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (na *NotificationAggregator) DispatchNotication(notification Notification) {
+	na.Notifications = append(na.Notifications, notification)
+}
+
+func (na *NotificationAggregator) ProcessNotifications() {
+	// Create a mapping of provider and their messages
+	providerNotifications := make(map[string][]Notification)
+
+	// Segregate messages into
```
mlavacca, line 57: I feel this comment is not much explanatory; could you either remove or make it clearer?

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,74 @@
...
+type MessageType string
+
+type Notification struct {
+	Type     MessageType
+	Message  string
+	Provider string
+}
+
+type NotificationAggregator struct {
+	Notifications []Notification
```
mlavacca, line 44: Since we already have the provider name, why not use a `map[string]Notification`, with the providers as the key? This way we'd save the aggregation step we perform in the `ProcessNotifications` function

pkg/i2gw/notifications/config.go
```diff
@@ -0,0 +1,63 @@
...
+	errorColor   = text.FgRed
+	defaultColor = text.FgWhite
+)
+
+func newTableConfig() table.Writer {
+	t := table.NewWriter()
+
+	t.SetOutputMirror(os.Stdout)
+	t.SetRowPainter(func(row table.Row) text.Colors {
+		switch row[1] {
```
mlavacca, line 40:
```suggestion
		switch messageType := row[1]; messageType {
```

I know this is 100% the same, but I feel it'd increase code readability

mlavacca: Also, I think we should perform some checks on the row length. To make it bulletproof, we need to be sure that `len(row) > 1`

pkg/i2gw/notifications/config.go
```diff
@@ -0,0 +1,63 @@
...
+			return text.Colors{bgColor, warningColor}
+		case ErrorNotification:
+			return text.Colors{bgColor, errorColor}
+		default:
+			return text.Colors{bgColor, defaultColor}
+		}
+	})
+
+	t.SetColumnConfigs([]table.ColumnConfig{
+		{Number: 3, WidthMax: 100},
```
mlavacca, line 53: Can we extract these parameters into proper consts?

pkg/i2gw/notifications/config.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+
+	style := table.StyleRounded
+	style.Options.SeparateRows = true
+	style.Title.Align = text.AlignCenter
```
mlavacca, line 58: The same for these parameters

mlavacca (commented):
Thanks for this first PR, @Devaansh-Kumar!

/ok-to-test

pkg/i2gw/notifications/notifications.go
Devaansh-Kumar: So do you mean something like: 
```go
type NotificationAggregator struct {
	Notifications map[string][]Notification
}
var CommonNotifications NotificationAggregator

CommonNotification.Notifications[string(ProviderName)] = append(CommonNotification.Notifications[string(ProviderName)], newNotification)
```

pkg/i2gw/notifications/notifications.go
Devaansh-Kumar: I agree with changing the Heading "Message Type" to "Notification Type", but withing the `Notification` struct I cannot change `Message` field to `Notification` as there would be a name clash and `Notification` struct consists of both a Type and a message that we want to send.
Do you thing I should change `MessageType` to `NotificationType` as well?

pkg/i2gw/notifications/config.go
Devaansh-Kumar: That is automatically handled by the `table` package. If there are no rows, then it wont render anything. So the check is unnecessary

Devaansh-Kumar:
@mlavacca I have made changes according to your comments and left some comments of my own and am ready for next round of review

PROVIDER.md
```diff
@@ -137,7 +137,18 @@ func init() {
 	i2gw.ProviderConstructorByName[Name] = NewProvider
 }
 ```
-6. Import the new package at `cmd/print`.
+6. In order to use notification mechanism, create a `notify` function in a file named `notification.go`. This method is used to reduce the function signature for creating notifications during the conversion process.
```
LiorLieberman, line 140:
```suggestion
6. [optional] In order to use notification mechanism, create a `notify` function in a file named `notification.go`. This method is used to reduce the function signature for creating notifications during the conversion process.
```

pkg/i2gw/notifications/config.go
```diff
@@ -0,0 +1,72 @@
...
+limitations under the License.
+*/
+
+package notifications
+
+import (
+	"os"
+
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/jedib0t/go-pretty/v6/text"
```
LiorLieberman, line 23: We should be careful importing these non standard libraries, are there more common alternatives ?

pkg/i2gw/notifications/config.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+const (
+	bgColor = text.BgBlack
+
+	infoColor    = text.FgHiWhite
+	warningColor = text.FgHiMagenta
+	errorColor   = text.FgRed
+	defaultColor = text.FgWhite
```
LiorLieberman, line 32: consider calling the colors by the names, not where we expect them to be used.
errorColor --> red
infoColor --> white

LiorLieberman (commented):
I am uncomfortable to approve this CL because I want to see the whole picture of how the notifications will be added, then I can have an opinion on the structure introduced here as well.

Can you add the other part of the logic here ? 
To make review easy, you can logically split it to commits instead of separate PRs.

Devaansh-Kumar:
> I am uncomfortable to approve this CL because I want to see the whole picture of how the notifications will be added, then I can have an opinion on the structure introduced here as well.
> 
> Can you add the other part of the logic here ? To make review easy, you can logically split it to commits instead of separate PRs.

By other part you mean the part involving having a calling object as a column right?

pkg/i2gw/notifications/config.go
Devaansh-Kumar: This is a well used library, so it should not be a concern. One alternative would be https://github.com/olekukonko/tablewriter but i feel the package we are using offers more flexibility and options.

pkg/i2gw/notifications/config.go
Devaansh-Kumar: i dont think i understood this comment. Do you want me to remove these and directly put the colors directly in the switch case statements?

LiorLieberman:
No, I am fine if you add that column later (in a different commit or PR).
I mean the part where providers use this package. Currently no one is
adding or removing notifications IIRC.

pkg/i2gw/ingress2gateway.go
```diff
@@ -74,6 +75,7 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 		errs = append(errs, conversionErrs...)
 		gatewayResources = append(gatewayResources, providerGatewayResources)
 	}
+	notifications.CommonNotification.ProcessNotifications()
```
LiorLieberman, line 78: maybe change ProcessNotifications to something like `createNotificationTables`

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,84 @@
...
+	CommonNotification = NotificationAggregator{map[string][]Notification{}}
+}
+
+const (
+	InfoNotification    MessageType = "INFO"
+	WarningNotification MessageType = "WARNING"
+	ErrorNotification   MessageType = "ERROR"
+)
+
+type MessageType string
```
LiorLieberman, line 37:
```suggestion
type Severity string
```

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,84 @@
...
+	Type           MessageType
+	Message        string
+	CallingObjects []client.Object
+}
+
+type NotificationAggregator struct {
+	Notifications map[string][]Notification
+}
+
+var CommonNotification NotificationAggregator
```
LiorLieberman, line 49: notificationAggr

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,84 @@
...
+
+type MessageType string
+
+type Notification struct {
+	Type           MessageType
+	Message        string
+	CallingObjects []client.Object
+}
+
+type NotificationAggregator struct {
```
LiorLieberman, line 45: lets add a lock here, and then in dispatch we should acquire the lock before we add something. Also when we read the notifications we should acquire the locks first.

We have nothing concurrent yet but this sets up a good infrastructure for the future

pkg/i2gw/providers/istio/converter.go
```diff
@@ -295,18 +307,20 @@ var hostnameRegexp = regexp.MustCompile(`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\
...
-func convertHostnames(hosts []string) []gatewayv1.Hostname {
+func convertHostnames(ctx context.Context, hosts []string, fieldPath *field.Path) []gatewayv1.Hostname {
 	var resHostnames []gatewayv1.Hostname
-	for _, host := range hosts {
+	vs := ctx.Value(virtualServiceKey).(*istioclientv1beta1.VirtualService)
+	for i, host := range hosts {
 		// '*' is valid in istio, but not in HTTPRoute
+		hostsFieldPath := fieldPath.Child("Hosts").Key(fmt.Sprintf("%v", i))
 		if !hostnameRegexp.MatchString(host) {
-			klog.Warningf("ignoring host %s, which is not allowed in Gateway API HTTPRoute", host)
```
LiorLieberman, line 303: can you keep the logs instead of replacing them?

pkg/i2gw/notifications/notifications.go
Devaansh-Kumar: I think we decided to let this be as `MessageType`

pkg/i2gw/notifications/config.go
```diff
@@ -0,0 +1,71 @@
...
+
+	tableStyle    table.Style = table.StyleRounded
+	rowSeperation             = true
+	textAlignment text.Align  = text.AlignCenter
+)
+
+func newTableConfig() table.Writer {
+	t := table.NewWriter()
+
+	t.SetOutputMirror(os.Stdout)
```
LiorLieberman, line 46: consider removing this. Then when you call `CreateNotificationTables` assign it to a `table` variable and return the table from `ToGatewayApiResources` 

Then I'd use it somewhere [here](https://github.com/kubernetes-sigs/ingress2gateway/blob/2333bec43b76da886ec521b3a9aea5bf28652a22/cmd/print.go#L92) to print it. I think thats a bit cleaner.

pkg/i2gw/notifications/notifications.go
```diff
@@ -0,0 +1,88 @@
...
+var NotificationAggr NotificationAggregator
+
+// DispatchNotification is used to send a notification to the NotificationAggregator
+func (na *NotificationAggregator) DispatchNotification(notification Notification, ProviderName string) {
+	na.mutex.Lock()
+	na.Notifications[ProviderName] = append(na.Notifications[ProviderName], notification)
+	na.mutex.Unlock()
+}
+
+// CreateNotificationTables takes all generated notifications and displays it in a tabular format based on provider
```
LiorLieberman, line 60: if you decide to apply my comment about returning a a string from here (i.e return `t.Render()`) then dont forget to update this comment to "returns it in a tabular format"

pkg/i2gw/providers/istio/converter.go
```diff
@@ -79,6 +89,8 @@ func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.Erro
 			Name:      vs.Name,
 		}.String())
 
+		c.ctx = context.WithValue(c.ctx, virtualServiceKey, vs)
```
LiorLieberman, line 92: add a comment what we store in the context and why

LiorLieberman (commented):
This is great, thank you Devaansh!

Mostly LGTM, I left a few last comments, plus can you add a test to notifications.go ? to see that we get the tables we expect?

If its too hard for this PR, I am happy to wait for a followup PR but if its straightforward i'd prefer to include it here.

Devaansh-Kumar:
> This is great, thank you Devaansh!
> 
> Mostly LGTM, I left a few last comments, plus can you add a test to notifications.go ? to see that we get the tables we expect?
> 
> If its too hard for this PR, I am happy to wait for a followup PR but if its straightforward i'd prefer to include it here.

Yeah I would prefer to do this in a seperate PR.

pkg/i2gw/notifications/config.go
LiorLieberman: ok about this, if it was an APACHE licenses it was all good. 
I checked and it looks like CNCF has a [process](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md#approved-licenses-for-allowlist) for allowlisting MIT licenses. https://github.com/olekukonko/tablewriter is allowed, you can find it [here](https://github.com/cncf/foundation/blob/main/license-exceptions/CNCF-licensing-exceptions.csv). 

The approval processes would probably take a bit of time, and has some requirements, [here](https://github.com/cncf/foundation/issues/809) is an example of exception request.

Up to you if you want to go through this process, or move to use `github.com/olekukonko/tablewriter` and remove the colors. I'd go with the latter.

pkg/i2gw/notifications/notifications.go
```diff
@@ -69,8 +71,10 @@ func (na *NotificationAggregator) CreateNotificationTables() {
 			t.AppendRow(table.Row{n.Type, n.Message, convertObjectsToStr(n.CallingObjects)})
 		}
 
-		t.Render()
+		tables = append(tables, t)
```
LiorLieberman, line 74: would it be simpler to have tables as list of strings. 

and then you do `tables = append(tabels,t.render())` ?

Then you would not need to render them outside this function
see the related comment on print.go

cmd/print.go
```diff
@@ -85,19 +86,24 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
 	}
 
-	pr.outputResult(gatewayResources)
+	pr.outputResult(gatewayResources, tables)
 
 	return nil
 }
 
-func (pr *PrintRunner) outputResult(gatewayResources []i2gw.GatewayResources) {
+func (pr *PrintRunner) outputResult(gatewayResources []i2gw.GatewayResources, tables []table.Writer) {
```
LiorLieberman, line 99: I dont think tables are part of the result. The result are only the resulted resources.

So why not simply print it above line 94?
if you apply the comment I left below so you `tables` will be a list of already print-ready tables and you can just print it there (above line 94)

pkg/i2gw/notifications/notifications.go
Devaansh-Kumar: But when we apply t.Render() it automatically outputs it on the terminal. It returns a string as well. So there is no point of having just t.Render()

cmd/print.go
Devaansh-Kumar: Ok i will change this

pkg/i2gw/notifications/notifications.go
```diff
@@ -73,9 +73,10 @@ func (na *NotificationAggregator) CreateNotificationTables() []*tablewriter.Tabl
 			t.Append(row)
 		}
 
-		tables = append(tables, t)
+		tableString.WriteString(fmt.Sprintf("Notifications from %v\n", provider))
+		t.Render()
 	}
-	return tables
+	return tableString
```
LiorLieberman, line 79: correct me if I am wrong, but it looks like you hold all the tables in one string `tableString`. Is there a reason you dont do as we said to have a list of `tableStrings` ?

pkg/i2gw/notifications/notifications.go
```diff
@@ -73,9 +73,10 @@ func (na *NotificationAggregator) CreateNotificationTables() []*tablewriter.Tabl
 			t.Append(row)
 		}
 
-		tables = append(tables, t)
+		tableString.WriteString(fmt.Sprintf("Notifications from %v\n", provider))
```
LiorLieberman, line 76: why this is needed?

Devaansh-Kumar: Earlier i could add the name of provider as a seperate header in the table above the headings of columns, but in tablewriter we cant do that, so to specify the provider for which the table is created i added this line

pkg/i2gw/notifications/notifications.go
Devaansh-Kumar: Oh i think i missed that, I ll make the change. Do you think it would be better to have a map of provider to string instead of a list.

LiorLieberman: I dont think thats necessary now. If you can think of any use of it *now* - do it.

pkg/i2gw/notifications/notifications.go
LiorLieberman: Oh ok. So it looks like just the title of the table? Maybe wrap the provider with quotes or make it capital letters.

pkg/i2gw/notifications/notifications.go
Devaansh-Kumar: It will be easier to parse and use in my opinion thats all. Plus we get a reference to the provider generating the table this way. We can easy change to a list implementation later is we feel it is unnecessary

cmd/print.go
```diff
@@ -82,11 +85,15 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
-	gatewayResources, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.getProviderSpecificFlags())
+	gatewayResources, tableString, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.getProviderSpecificFlags())
 	if err != nil {
 		return err
 	}
 
+	for _, table := range tableString {
+		fmt.Println(table)
+	}
+
```
LiorLieberman, line 96:
```suggestion
	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.getProviderSpecificFlags())
	if err != nil {
		return err
	}

	for _, table := range notificationTablesMap {
		fmt.Println(table)
	}

```

pkg/i2gw/ingress2gateway.go
```diff
@@ -74,11 +75,12 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 		errs = append(errs, conversionErrs...)
 		gatewayResources = append(gatewayResources, providerGatewayResources)
 	}
+	tableString := notifications.NotificationAggr.CreateNotificationTables()
```
LiorLieberman, line 78:
```suggestion
	notificationTablesMap := notifications.NotificationAggr.CreateNotificationTables()
```

pkg/i2gw/ingress2gateway.go
```diff
@@ -74,11 +75,12 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
...
 	if len(errs) > 0 {
-		return nil, aggregatedErrs(errs)
+		return nil, tableString, aggregatedErrs(errs)
```
LiorLieberman, line 80:
```suggestion
		return nil, notificationTablesMap, aggregatedErrs(errs)
```

pkg/i2gw/ingress2gateway.go
```diff
@@ -74,11 +75,12 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
...
 	}
 
-	return gatewayResources, nil
+	return gatewayResources, tableString, nil
```
LiorLieberman, line 83:
```suggestion
	return gatewayResources, notificationTablesMap, nil
```

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/getkin/kin-openapi v0.124.0
 	github.com/google/go-cmp v0.6.0
 	github.com/kong/kubernetes-ingress-controller/v2 v2.12.3
+	github.com/olekukonko/tablewriter v0.0.5
```
LiorLieberman, line 9: @robscott I found that this is allowlisted, can you double check I am right?

robscott: Yep, looks like this is [very widely used in k8s repos](https://cs.k8s.io/?q=github.com%2Folekukonko%2Ftablewriter&i=nope&files=&excludeFiles=&repos=), thanks for checking!

LiorLieberman:
@Devaansh-Kumar can you rebase?

PR #167, opened 2024-07-12
Add ingress class

### LINKED ISSUE #87: Add support for GCE ingress class/provider
**What would you like to be added**: support for the GCE ingress provider

**Why this is needed**: to convert existing GCE ingress to the new GW API

Error while executing 
```sh
ingress2gateway print
I1026 10:58:24.119436   48466 request.go:682] Waited for 1.049636692s due to client-side throttling, not priority and fairness, request: GET:https://XX.XX.XX.XX/apis/elasticsearch.k8s.elastic.co/v1beta1?timeout=32s
Error:


# Encountered 2 errors # spec.rules[0].http.paths[0].pathType: Invalid value: "ImplementationSpecific": unsupported path match type: ImplementationSpecific # spec.rules[0].http.paths[0].pathType: Invalid value: "ImplementationSpecific": unsupported path match type: ImplementationSpecific
Usage:
  ingress2gateway print [flags]

Flags:
  -A, --all-namespaces      If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
                            if specified with --namespace.
  -h, --help                help for print
      --input_file string   Path to the manifest file. When set, the tool will read ingresses from the file instead of reading from the cluster. Supported files are yaml and json
  -n, --namespace string    If present, the namespace scope for this CLI request
  -o, --output string       Output format. One of: (json, yaml) (default "yaml")
      --providers strings   If present, the tool will try to convert only resources related to the specified providers, supported values are [ingress-nginx kong] (default [ingress-nginx,kong])
```

the current ingress has the following YAML (Redacted info)

```yaml
kind: Ingress
metadata:
  annotations:
    ingress.gcp.kubernetes.io/pre-shared-cert: mcrt-XXX,mcrt-YYY,mcrt-ZZZ
    ingress.kubernetes.io/backends: '{"k8s1-XXX":"HEALTHY"}'
    ingress.kubernetes.io/forwarding-rule: k8s2-fr-XXX
    ingress.kubernetes.io/https-forwarding-rule: k8s2-fs-XXX
    ingress.kubernetes.io/https-target-proxy: k8s2-ts-XXX
    ingress.kubernetes.io/redirect-url-map: k8s2-rm-XXX
    ingress.kubernetes.io/ssl-cert: mcrt-XXX,mcrt-YYY,mcrt-ZZZ
    ingress.kubernetes.io/target-proxy: k8s2-tp-XXX
    ingress.kubernetes.io/url-map: k8s2-um-XXX
    kubernetes.io/ingress.allow-http: "true"
    kubernetes.io/ingress.class: gce
    kubernetes.io/ingress.global-static-ip-name: ingress-gw-public-XXX
    networking.gke.io/managed-certificates: XXX,YYY,ZZZ
    networking.gke.io/v1beta1.FrontendConfig: default-frontend-config-XXX
  labels:
    app.kubernetes.io/managed-by: pulumi
  name: gke-ingress-XXX
spec:
  defaultBackend:
    service:
      name: istio-XXX
      port:
        number: 443
  rules:
  - host: '*.example.com'
    http:
      paths:
      - backend:
          service:
            name: istio-XXX
            port:
              number: 443
        path: /*
        pathType: ImplementationSpecific
  tls:
  - hosts:
    - XXX.example.com
    secretName: secret-XXX
  - hosts:
    - YYY.example.com
    secretName: secret-YYY
  - hosts:
    - ZZZ.example.com
    secretName: secret-ZZZ
status:
  loadBalancer:
    ingress:
    - ip: XX.XX.XX.XX
```

sawsa307:
**What type of PR is this?**
/kind bug

Make sure Gateway name is not empty when the GKE ingress class name is not specified.

Fixes #87

```release-note
Fixed an issue that when the ingress class annotation is not specified on a GKE Ingress, the translation would result in a Gateway without name.
```

### DIFF
--- a/pkg/i2gw/providers/gce/converter.go
+++ b/pkg/i2gw/providers/gce/converter.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/common"
 	networkingv1 "k8s.io/api/networking/v1"
+	networkingv1beta1 "k8s.io/api/networking/v1beta1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
@@ -45,6 +46,12 @@ func newConverter(conf *i2gw.ProviderConf) converter {
 func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.ErrorList) {
 	ingressList := []networkingv1.Ingress{}
 	for _, ing := range storage.Ingresses {
+		if ing != nil && common.GetIngressClass(*ing) == "" {
+			if ing.Annotations == nil {
+				ing.Annotations = make(map[string]string)
+			}
+			ing.Annotations[networkingv1beta1.AnnotationIngressClass] = gceIngressClass
+		}
 		ingressList = append(ingressList, *ing)
 	}
 
--- a/pkg/i2gw/providers/gce/converter_test.go
+++ b/pkg/i2gw/providers/gce/converter_test.go
@@ -224,13 +224,12 @@ func Test_ToGateway(t *testing.T) {
 			expectedErrors: field.ErrorList{},
 		},
 		{
-			name: "no ingress class, default to gce ingress class behavior",
+			name: "empty ingress class, default to gce ingress class",
 			ingresses: map[types.NamespacedName]*networkingv1.Ingress{
 				{Namespace: testNamespace, Name: noIngClassIngressName}: {
 					ObjectMeta: metav1.ObjectMeta{
-						Name:        noIngClassIngressName,
-						Namespace:   testNamespace,
-						Annotations: map[string]string{networkingv1beta1.AnnotationIngressClass: gceIngressClass},
+						Name:      noIngClassIngressName,
+						Namespace: testNamespace,
 					},
 					Spec: networkingv1.IngressSpec{
 						Rules: []networkingv1.IngressRule{{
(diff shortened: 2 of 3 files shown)


LiorLieberman:
Thanks David, mind adding a test for it?

sawsa307:
> Thanks David, mind adding a test for it?

Just added one. Please take a look!

pkg/i2gw/providers/gce/converter_test.go
```diff
@@ -224,13 +224,12 @@ func Test_ToGateway(t *testing.T) {
...
-			name: "no ingress class, default to gce ingress class behavior",
+			name: "empty ingress class, default to gce ingress class",
 			ingresses: map[types.NamespacedName]*networkingv1.Ingress{
 				{Namespace: testNamespace, Name: noIngClassIngressName}: {
 					ObjectMeta: metav1.ObjectMeta{
-						Name:        noIngClassIngressName,
-						Namespace:   testNamespace,
-						Annotations: map[string]string{networkingv1beta1.AnnotationIngressClass: gceIngressClass},
+						Name:      noIngClassIngressName,
+						Namespace: testNamespace,
```
LiorLieberman, line 232: shouldn't we expect in this test to have an ingressclass annotation? 

Also, why not a ingressClass field?

pkg/i2gw/providers/gce/converter.go
```diff
@@ -45,6 +46,12 @@ func newConverter(conf *i2gw.ProviderConf) converter {
 func (c *converter) convert(storage *storage) (i2gw.GatewayResources, field.ErrorList) {
 	ingressList := []networkingv1.Ingress{}
 	for _, ing := range storage.Ingresses {
+		if common.GetIngressClass(*ing) == "" {
```
levikobi, line 49: Consider adding a check for `ing != nil` to avoid panicking when the ingress is nil

sawsa307: Sounds good, added. Thanks!

pkg/i2gw/providers/gce/converter_test.go
sawsa307: 1.  GCE/GKE does not use ingressClass field. The annotation `kubernetes.io/ingress.class` is used to specify the ingress class.
2. This test case is regarding the scenario that neither annotation or ingressClass field is set. In such case, GKE default it to be an external ingress.

Both information can be found here https://cloud.google.com/kubernetes-engine/docs/how-to/load-balance-ingress#create-ingress

PR #175, opened 2024-07-29
ingress-nginx: apply canary to httproute only when enabled

### LINKED ISSUE #174: Convention of Ingress with multiple rules generates incorrect results
**What happened**:
I try to convert the ingress resource from single file using Ingress2Gateway v0.3.0-rc1, but the generated HTTPRoute is incorrect. 
```
kind: HTTPRoute
metadata:
  creationTimestamp: null
  name: example-ingress-foo-example-com
spec:
  hostnames:
  - foo.example.com
  parentRefs:
  - name: nginx
  rules:
  - backendRefs:
    - name: foo-app
      port: 80
    - name: foo-orders-app
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /
  - backendRefs:
    - name: foo-orders-app
      port: 80
    - name: foo-app
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /orders
status:
  parents: []
```

Please see `How to reproduce it` below

The correct result should be:

```
kind: HTTPRoute
metadata:
  name: foo
spec:
  parentRefs:
  - name: example-gateway
    sectionName: https
  hostnames:
  - foo.example.com
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - name: foo-app
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: /orders
    backendRefs:
    - name: foo-orders-app
      port: 80
```

Here's my ingress resource, it's just an example from Gateway API doc [https://gateway-api.sigs.k8s.io/guides/migrating-from-ingress/#example](url)
In order to use Ingress2Gateway, I have to change the value of ingressClassName to nginx and delete annotion `some-ingress-controller.example.org/tls-redirect`.
After convention, The HTTPRoute `example-ingress-foo-example-com` has two backendRefs

The command is
`ingress2gateway print --input-file ingress.yaml -A --providers ingress-nginx`

```apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: example-ingress
spec:
  ingressClassName: nginx
  tls:
  - hosts:
    - foo.example.com
    - bar.example.com
    secretName: example-com
  rules:
  - host: foo.example.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: foo-app
            port:
              number: 80
      - path: /orders
        pathType: Prefix
        backend:
          service:
            name: foo-orders-app
            port:
              number: 80
  - host: bar.example.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: bar-app
            port:
              number: 80
```

LiorLieberman:
**What type of PR is this?**
/kind bug

This is a fix for #174, we should not try to parse canary if its not enabled with the corresponding ingress annotations.

Fixes #174 

```release-note
ingress-nginx: fix backendRefs duplication due to wrong canary annotations logic
```

/cc @levikobi

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/canary.go
+++ b/pkg/i2gw/providers/ingressnginx/canary.go
@@ -73,12 +73,14 @@ func getPathsByMatchGroups(rg common.IngressRuleGroup) (map[pathMatchKey][]ingre
 			return nil, errs
 		}
 
-		extraFeatures := extra{canary: &annotations}
+		if annotations.enable {
+			extraFeatures := extra{canary: &annotations}
 
-		for _, path := range ir.IngressRule.HTTP.Paths {
-			ip := ingressPath{ingress: ingress, ruleType: "http", path: path, extra: &extraFeatures}
-			pmKey := getPathMatchKey(ip)
-			ingressPathsByMatchKey[pmKey] = append(ingressPathsByMatchKey[pmKey], ip)
+			for _, path := range ir.IngressRule.HTTP.Paths {
+				ip := ingressPath{ingress: ingress, ruleType: "http", path: path, extra: &extraFeatures}
+				pmKey := getPathMatchKey(ip)
+				ingressPathsByMatchKey[pmKey] = append(ingressPathsByMatchKey[pmKey], ip)
+			}
 		}
 	}
 


LiorLieberman:
This is actually not the solution. We need to fix https://github.com/kubernetes-sigs/ingress2gateway/blob/beb19dc871bce8c47a114563ce72f0ea69e869d2/pkg/i2gw/providers/ingressnginx/canary.go#L216

LiorLieberman:
close in favor of https://github.com/kubernetes-sigs/ingress2gateway/pull/182

PR #180, opened 2024-08-15
Added notifications in the common package

Devaansh-Kumar:
**What type of PR is this?**

/kind feature

This PR adds notifications from the common package

This pull request addresses part of the issue described in #131 

```release-note
Added notifications from the common package
```

/cc @mlavacca @LiorLieberman

### DIFF
--- a/pkg/i2gw/providers/common/converter.go
+++ b/pkg/i2gw/providers/common/converter.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/intermediate"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -34,8 +35,20 @@ import (
 
 // ToIR converts the received ingresses to intermediate.IR without taking into
 // consideration any provider specific logic.
-func ToIR(ingresses []networkingv1.Ingress, options i2gw.ProviderImplementationSpecificOptions) (intermediate.IR, field.ErrorList) {
+//
+// If a provider wishes to recieve notifications from the common package,
+// it can pass a notifications.NotificationCallback function which can be used
+// to send notifications on behalf of the provider while keeping the logic of
+// ToGateway seperated from the provider.
+func ToIR(ingresses []networkingv1.Ingress, options i2gw.ProviderImplementationSpecificOptions, notifyOpts ...notifications.NotificationCallback) (intermediate.IR, field.ErrorList) {
 	aggregator := ingressAggregator{ruleGroups: map[ruleGroupKey]*ingressRuleGroup{}}
+
+	var notify notifications.NotificationCallback
+	if len(notifyOpts) > 0 {
+		notify = notifyOpts[0]
+	} else {
+		notify = noNotifications
+	}
 
 	var errs field.ErrorList
 	for _, ingress := range ingresses {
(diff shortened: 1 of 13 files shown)


pkg/i2gw/providers/common/utils.go
```diff
@@ -200,3 +203,30 @@ func removeBackendRefsDuplicates(backendRefs []gatewayv1.HTTPBackendRef) []gatew
 	}
 	return uniqueBackendRefs
 }
+
+func noNotifications(mType notifications.MessageType, message string, callingObject ...client.Object) {
+}
+
+func mockIngressFromRuleGroup(rg *ingressRuleGroup) client.Object {
```
mlavacca, line 210: It took me a bit to figure out why we are introducing testing helpers in this package, and then I realized this is not a testing helper, but it's a function to build an Ingress out of a ruleGroup. I suggest renaming it to something like `buildIngressFromRuleGroup` or similar.

pkg/i2gw/providers/common/utils.go
```diff
@@ -200,3 +203,30 @@ func removeBackendRefsDuplicates(backendRefs []gatewayv1.HTTPBackendRef) []gatew
...
+			Kind: "Ingress",
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Name:      rg.name,
+			Namespace: rg.namespace,
+		},
+	}
+}
+
+func mockIngressFromDefaultBackend(db ingressDefaultBackend) client.Object {
```
mlavacca, line 222: The same here

pkg/i2gw/providers/common/utils.go
```diff
@@ -200,3 +203,30 @@ func removeBackendRefsDuplicates(backendRefs []gatewayv1.HTTPBackendRef) []gatew
 	}
 	return uniqueBackendRefs
 }
+
+func noNotifications(mType notifications.MessageType, message string, callingObject ...client.Object) {
+}
```
mlavacca, line 208:
```suggestion
func noNotifications(mType notifications.MessageType, message string, callingObject ...client.Object) {}
```

mlavacca: Also, I think this function could benefit from a comment.

pkg/i2gw/providers/common/converter.go
```diff
@@ -31,9 +32,21 @@ import (
...
+// ToGateway seperated from the provider.
+func ToGateway(ingresses []networkingv1.Ingress, options i2gw.ProviderImplementationSpecificOptions, notifyOpts ...notifications.NotificationCallback) (i2gw.GatewayResources, field.ErrorList) {
 	aggregator := ingressAggregator{ruleGroups: map[ruleGroupKey]*ingressRuleGroup{}}
 
+	var notify notifications.NotificationCallback
+	if len(notifyOpts) > 0 {
+		notify = notifyOpts[0]
+	} else {
+		notify = noNotifications
+	}
```
mlavacca, line 48:
```suggestion
	switch len(notifyOpts) {
	case 0:
		notify = noNotifications
	case 1:
		notify = notifyOpts[0]
	default:
		return fmt.Errorf("unsupported amount of callbacks provided. Only 0 or 1 callbacks are currently supported")
	}
```
I think we should clearly state that only 0 - 1 callbacks are allowed here. This error can be triggered only by a wrong provider implementation, so we should avoid it to happen. Feel free to rephrase the error message as you prefer and/or re-structure the switch.

pkg/i2gw/providers/gce/notification.go
```diff
@@ -22,6 +22,6 @@ import (
 )
 
 func notify(mType notifications.MessageType, message string, callingObject ...client.Object) {
-	newNotification := notifications.Notification{Type: mType, Message: message, CallingObjects: callingObject}
+	newNotification := notifications.NewNotification(mType, message, callingObject...)
```
LiorLieberman, line 25: In go, variadic arguments should be plural, so CallingObject should be callingObjects

LiorLieberman: Also, notify is a func, should we call it noifyFunc?

Devaansh-Kumar: I ll change the CallingObject to Calling Objects.

I dont think it makes a difference whether we use notify or notifyFunc. Both can convey the same message. We dont put Func after every function i suppose.

LiorLieberman: Fair. But most of the functions are not passed as params to different functions. Totally fine to keep it like this as well though

LiorLieberman (commented):
We need tests for it

PR #185, opened 2024-08-21
Create IR struct to hold specs for extension features.

### LINKED ISSUE #125: Support Gateway API extensions (output providers)

Add the ability to translate `provider` specific APIs into implementation specific Gateway API extensions (output providers) for fields not currently supported in the Gateway API

**Why this is needed**:
Adding ability to translate all input provider specific intent without losing any functionality
For e.g. the current [ingress-nginx](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/ingressnginx/README.md#ingress-nginx-provider) translation only translates a subset of the API (annotations)  https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#annotations 
Without this feature, its unlikely that the user will migrate to Gateway API until all the input provider features are supported in the Gateway API APIs
If this tool supports the notion of output providers, the input API could be translated to various implementation specific Gateway API extensions

sawsa307:
**What type of PR is this?**
/kind feature

Add structures needed to support translation of extension features.

Fixes #125

```release-note
None
```


pkg/i2gw/provider.go
```diff
@@ -99,8 +101,90 @@ type GatewayResources struct {
...
+	ReferenceGrants map[types.NamespacedName]gatewayv1beta1.ReferenceGrant
 }
 
+// GatewayContext contains the Gateway-API Gateway object, along with a
+// dedicated field for each provider to specify their extension features on
+// Gateway.
+type GatewayContext struct {
+	gatewayv1.Gateway
+
+	ApisixGatewayIR
```
arkodg, line 131: we probably need to not embed the extension fields and instead use
```
Apisix *ApisixGatewayIR
```

because 
* they are optional 
* the same fields may exist in both extensions

pkg/i2gw/provider.go
```diff
@@ -99,8 +101,90 @@ type GatewayResources struct {
 	UDPRoutes  map[types.NamespacedName]gatewayv1alpha2.UDPRoute
 
 	ReferenceGrants map[types.NamespacedName]gatewayv1beta1.ReferenceGrant
+
+	GatewayExtensions map[types.NamespacedName]unstructured.Unstructured
+}
+
+// IR contains all Gateway-API objects, and Intermediate Representation(IR) to
```
arkodg, line 108: curious what's going to happen to the `GatewayResources` struct

pkg/i2gw/provider.go
```diff
@@ -99,8 +101,90 @@ type GatewayResources struct {
 	UDPRoutes  map[types.NamespacedName]gatewayv1alpha2.UDPRoute
 
 	ReferenceGrants map[types.NamespacedName]gatewayv1beta1.ReferenceGrant
+
+	GatewayExtensions map[types.NamespacedName]unstructured.Unstructured
```
arkodg, line 105: why is this field needed here

pkg/i2gw/provider.go
arkodg: shouldnt the `IR` replace it ?

pkg/i2gw/provider.go
robscott: This was an alternative that I'd discussed earlier with @sawsa307 - instead of modifying the existing path at all, we'd just add `GatewayExtensions` type that output providers could optionally read from. I slightly prefer that over the `GatewayContext` struct as it's less disruptive and easy to ignore for implementations that don't support extensions.

Imagine that instead of `unstructured.Unstructured` here we had a `GatewayExtension` struct that had all the content of `GatewayContext` except for the actual Gateway. I can be ok with the `GatewayContext` approach as well, I just generally preferred the optionality of this kind of approach.

pkg/i2gw/provider.go
sawsa307: IR and Gateway Resources are for different translation purpuse. What I have in mind is:

A pool of resources
------- > Ingress + Ingress extension (By input provider reader)
------- > IR, or Gateway/HTTPRoute context(By input provider converter)
------->  Gateway + Gateway extension(By output provider extension converter)

pkg/i2gw/provider.go
sawsa307: Sounds good, updated

pkg/i2gw/provider.go
sawsa307: Let's say we are doing translation of enableHTTPRedirect feature, we will have:

type GceGatewayIR struct {
   EnableHTTPRedirect bool
}
type GceHTTPRouteIR struct {
   EnableHTTPRedirect bool
}

And during translation, what would happen is:
Ingress+Ingress extension
--->IR.GceGatewayIR.EnableHTTPRedirect=true, IR.GceHTTPRouteIR.EnableHTTPRedirect=true
--->Gateway with additional HTTPRoute

Gateway extensions will be translated in a similar manner.

pkg/i2gw/provider.go
sawsa307: Yes, as Rob mentioned, the advantage of doing this as a map is, we can determine if we need to print `GatewayExtensions` based on if the map is empty.

The reason I used `unstructured.Unstructured` here is so we can hold Gateway Extensions for all providers. If we want to have a strongly typed field, it would have to be a field for each provider, and iterate over the supported Gateway extensions, something like:

```
type GatewayResources struct {
   ...
   
   Gce GceGatewayExtensions
}

type GceGatewayExtensions struct {
   GCPBackendPolicies map[types.NamespacedName]gkegatewayv1.GCPBackendPolicies
   GCPGatewayPolicies map[types.NamespacedName]gkegatewayv1.GCPGatewayPolicy
   HealthCheckPolicies map[types.NamespacedName]gkegatewayv1.HealthCheckPolicy
}
```

When printing results, we will also need to iterate over the list of extensions. This is my interpretation as an alternative for strongly typed, and I think it could be a valid option as well.

pkg/i2gw/provider.go
```diff
@@ -99,8 +101,94 @@ type GatewayResources struct {
...
+}
+
+// IR contains all Gateway-API objects, and Intermediate Representation(IR) to
+// hold specicifications on Gateway Objects for supporting Ingress extension
+// features. An feature on IR can be mapped to changes on Gateway-API objects,
+// or provider-specific Gateway extensions.
```
LiorLieberman, line 111:
```suggestion
// IR holds specifications of Gateway Objects for supporting Ingress extensions, annotations, and proprietary api
// features not supported as Gateway core features. An IR field can be mapped to core Gateway-API fields,
// or provider-specific Gateway extensions.
```

LiorLieberman (commented):
Can we see _one_ feature that will work with the IR, ideally in this PR as a different commit, so we can look at these independently (and we wont squash the PR when we merge, to keep the separation)?

pkg/i2gw/provider.go
sawsa307: Sure, just created an example for session affinity feature translation, PTAL!

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,225 @@
...
+		if bePolicy == nil {
+			continue
+		}
+		obj, err := i2gw.CastToUnstructured(bePolicy)
+		if err != nil {
+			notify(notifications.ErrorNotification, "Failed to cast GCPBackendPolicy to unstructured", bePolicy)
+			continue
+		}
+		key := types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()}
+		gatewayResources.GatewayExtensions[key] = *obj
```
arkodg, line 192: so you're treating a `GCPBackendPolicy` as a Gateway Extension here ? what happens when there are multiple kinds with the same namespaced name :) ?
if the goal is to use `GatewayExtensions` purely for marshaling reasons, then a list is better than a map

pkg/i2gw/provider.go
```diff
@@ -68,13 +70,19 @@ type CustomResourceReader interface {
...
+	// ToIR converts stored API entities associated with the Provider into IR.
+	ToIR() (IR, field.ErrorList)
+}
 
-	// ToGatewayAPIResources converts stored API entities associated
-	// with the Provider into GatewayResources.
-	ToGatewayAPI() (GatewayResources, field.ErrorList)
+// The ResourceConverter interface specifies conversion functions from IR
+// into Gateway and Gateway extensions.
+type ExtensionConverter interface {
```
LiorLieberman, line 82: Am I right that the purpose of this interface is to implement transition from the IR to Gateway API resources and extensions? If yes, `ExtensionConverter` name is a bit confusing. Maybe `IRToGatewayAPIConverter`?

I'd like to hear more thoughts. 

cc: @arkodg @robscott

pkg/i2gw/provider.go
```diff
@@ -68,13 +70,19 @@ type CustomResourceReader interface {
...
+	// ToGatewayAPIAndExtensions converts stored IR with the Provider into
+	// Gateway API resources and extensions
+	ToGatewayAPIAndExtensions(IR) (GatewayResources, field.ErrorList)
```
LiorLieberman, line 85: Do we really need to indicate extensions here? Also, how about two different functions on this interface, `ToGatewayAPI` and `ToGatewayExtensions` ? can also be `BuildGatewayAPIResources` & `BuildGatewayExtensions`

pkg/i2gw/provider.go
LiorLieberman: similarly, `ResourceConverter` can be changed to `ResourcesToIR` or something similar

pkg/i2gw/provider.go
```diff
@@ -99,14 +108,106 @@ type GatewayResources struct {
...
+// extensions, but not the extensions themselves.
+type GatewayContext struct {
+	gatewayv1.Gateway
+
+	Apisix       *ApisixGatewayIR
+	GCE          *GceGatewayIR
+	IngressNginx *IngressNginxGatewayIR
+	Istio        *IstioGatewayIR
+	Kong         *KongGatewayIR
+	Openapi3     *Openapi3GatewayIR
```
LiorLieberman, line 145: Thinking out loud - can we somehow make each provider register itself and his IR type with the GatewayContext? similarly to how we did with ProviderConstructorByName [1](https://github.com/kubernetes-sigs/ingress2gateway/blob/c626d145e4eae55a6ced3fd204afee9f676eace0/pkg/i2gw/provider.go#L35C5-L35C30 ) & [2](https://github.com/kubernetes-sigs/ingress2gateway/blob/c626d145e4eae55a6ced3fd204afee9f676eace0/pkg/i2gw/providers/apisix/apisix.go#L31)?

If possible, it is going to be easier to maintain and a bit cleaner

pkg/i2gw/providers/ingressnginx/ingressnginx.go
```diff
@@ -45,15 +46,21 @@ func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
...
 // including the ingress-nginx specific features.
-func (p *Provider) ToGatewayAPI() (i2gw.GatewayResources, field.ErrorList) {
+func (p *Provider) ToIR() (i2gw.IR, field.ErrorList) {
 	return p.converter.convert(p.storage)
 }
 
+func (p *Provider) ToGatewayAPIAndExtensions(ir i2gw.IR) (i2gw.GatewayResources, field.ErrorList) {
+	return p.extConverter.irToGateway(ir)
+
+}
```
LiorLieberman, line 62: we may probably change it based on my previous comments on the interface, but why not simply this? 

```go
func (p *Provider) ToGatewayAPIAndExtensions(ir i2gw.IR) (i2gw.GatewayResources, field.ErrorList) {
	return common.ToGatewayResources(ir)
}
```

pkg/i2gw/providers/ingressnginx/ingressnginx.go
```diff
@@ -37,6 +37,7 @@ type Provider struct {
 	storage        *storage
 	resourceReader *resourceReader
 	converter      *converter
+	extConverter   *extensionConverter
```
LiorLieberman, line 40: I just picked on ingress-nginx to leave my comments, but most of them are applicable for all providers. 

These names probably deserves better naming

pkg/i2gw/providers/gce/gce.go
```diff
@@ -32,16 +35,26 @@ func init() {
...
-	reader    reader
-	converter converter
+	storage      *storage
+	reader       reader
+	converter    converter
+	extConverter extensionConverter
 }
 
 func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
+	if conf.Client != nil {
```
LiorLieberman, line 45: I believe this check is in place to differentiate for when we are reading from a file, mind adding a short comment?

pkg/i2gw/providers/gce/resource_reader.go
LiorLieberman, line 1: Not necessarily now, but lots of logic here, I'd like to see tests for all the extraction of resources, especially from a file

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	if len(configs.Ports) != 0 {
+		notify(notifications.ErrorNotification, "Only config with default is supported since HealthCheckPolicy is attached on the whole service", service)
+		return "", false
+	}
+	return configs.Default, true
+}
+
+func ensureGceServiceIR(ctx context.Context, storage *storage, ir *i2gw.IR) {
```
LiorLieberman, line 108: this func could benefit from a better name and a comment

LiorLieberman: ditto for `ensureGceServiceExtensions`

LiorLieberman (commented):
Thank you David, this is great! Left some comments

pkg/i2gw/providers/gce/gce_extensions.go
LiorLieberman: +1

sawsa307: Yes that is a valid concern. Changing it to list.

pkg/i2gw/provider.go
sawsa307: Done.

pkg/i2gw/provider.go
sawsa307: I would say we we do `ToGatewayResources` here, but still having one struct and one function we writing all results to `GatewayResouces` struct.

pkg/i2gw/providers/gce/gce_extensions.go
arkodg: yeah i prefer `build` or `translate` as prefixes for the methods e.g. `buildGceServiceIR`

pkg/i2gw/providers/ingressnginx/ingressnginx.go
sawsa307: Sounds good, I'll update this for other providers as well. They can use the common function until they want to create a dedicate object to include translation logics for extensions.

pkg/i2gw/providers/gce/gce.go
sawsa307: Sounds good! Added.

pkg/i2gw/providers/gce/resource_reader.go
sawsa307: Sure, I'll add tests for reading from files as a follow up.

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: Done. Updated to `buildGceServiceIR` and `buildGceServiceExtensions`.

pkg/i2gw/providers/ingressnginx/ingressnginx.go
sawsa307: Updated. Now it is called `irToGatewayResourcesConverter`.

pkg/i2gw/provider.go
sawsa307: Yes that's a good suggestion! I think we can make them into a map of string to interface{}, so each provider can implement the IR however they wanted.

sawsa307: Updated, please take a look!

pkg/i2gw/provider.go
LiorLieberman: Why one function and not two? not sure I am following

sawsa307: Discussed offline, we will still use one function for conversion so we output one GatewayResources directly.

pkg/i2gw/ingress2gateway.go
```diff
@@ -136,53 +140,53 @@ func GetSupportedProviders() []string {
...
 //     ingresses, as they have a their GatewayClass' name as name. For this reason,
 //     if there are mutiple gateways named the same, their listeners are merged into
 //     a unique Gateway.
 //
 // This behavior is likely to change after https://github.com/kubernetes-sigs/gateway-api/pull/1863 takes place.
-func MergeGatewayResources(gatewayResources ...GatewayResources) (GatewayResources, field.ErrorList) {
-	mergedGatewayResources := GatewayResources{
-		Gateways:        make(map[types.NamespacedName]gatewayv1.Gateway),
+func MergeIRs(irs ...IR) (IR, field.ErrorList) {
+	mergedIRs := IR{
```
LiorLieberman, line 153: I cant see any referecnce for merging Service IRs

pkg/i2gw/provider.go
```diff
@@ -99,14 +108,106 @@ type GatewayResources struct {
...
+	HTTPRoutes map[types.NamespacedName]HTTPRouteContext
+	Services   map[types.NamespacedName]*ServiceIR
+
+	GatewayClasses map[types.NamespacedName]gatewayv1.GatewayClass
+	TLSRoutes      map[types.NamespacedName]gatewayv1alpha2.TLSRoute
+	TCPRoutes      map[types.NamespacedName]gatewayv1alpha2.TCPRoute
+	UDPRoutes      map[types.NamespacedName]gatewayv1alpha2.UDPRoute
+
+	ReferenceGrants map[types.NamespacedName]gatewayv1beta1.ReferenceGrant
+}
```
LiorLieberman, line 130: maybe `type IR` should live in its own package? Assuming you are going to change it to the IR folder like you did in https://github.com/sawsa307/ingress2gateway/tree/ir-dir

pkg/i2gw/provider.go
```diff
@@ -99,14 +108,106 @@ type GatewayResources struct {
...
+// The IR will contain necessary information to construct the Gateway
+// extensions, but not the extensions themselves.
+type GatewayContext struct {
+	gatewayv1.Gateway
+
+	Apisix       *ApisixGatewayIR
+	GCE          *GceGatewayIR
+	IngressNginx *IngressNginxGatewayIR
+	Istio        *IstioGatewayIR
+	Kong         *KongGatewayIR
```
LiorLieberman, line 144: I wonder if these types should live under a a type of `GatewayIRs` or something

pkg/i2gw/provider.go
sawsa307: Updated.

pkg/i2gw/provider.go
sawsa307: Updated.

pkg/i2gw/ingress2gateway.go
```diff
@@ -136,53 +140,53 @@ func GetSupportedProviders() []string {
 	return supportedProviders
 }
 
-// MergeGatewayResources accept multiple GatewayResources and create a unique Resource struct
-// built as follows:
+// MergeGatewayIRs accepts multiple IRs and creatse a unique IR struct built
+// as follows:
 //   - GatewayClasses, *Routes, and ReferenceGrants are grouped into the same maps
```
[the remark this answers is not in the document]
sawsa307: Updated

pkg/i2gw/ingress2gateway.go
sawsa307: Updated.

pkg/i2gw/ingress2gateway.go
```diff
@@ -136,53 +140,53 @@ func GetSupportedProviders() []string {
...
+		maps.Copy(mergedIRs.UDPRoutes, gr.UDPRoutes)
+		maps.Copy(mergedIRs.ReferenceGrants, gr.ReferenceGrants)
 	}
-	return mergedGatewayResources, errs
+	return mergedIRs, errs
 }
 
-func mergeGateways(gatewaResources []GatewayResources) (map[types.NamespacedName]gatewayv1.Gateway, field.ErrorList) {
-	newGateways := map[types.NamespacedName]gatewayv1.Gateway{}
+func mergeGateways(irs []IR) (map[types.NamespacedName]GatewayContext, field.ErrorList) {
```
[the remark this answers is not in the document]
sawsa307: Updated for GatewayContext. We don't currently perform any operations on HTTPRouteContext other than putting them together. There is no merge on HTTPRoutes.

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,188 @@
...
+		obj, err := i2gw.CastToUnstructured(bePolicy)
+		if err != nil {
+			notify(notifications.ErrorNotification, "Failed to cast GCPBackendPolicy to unstructured", bePolicy)
+			continue
+		}
+		gatewayResources.GatewayExtensions = append(gatewayResources.GatewayExtensions, *obj)
+	}
+}
+
+func toGCPBackendPolicy(serviceNamespacedName types.NamespacedName, serviceIR *i2gw.ServiceIR) *gkegatewayv1.GCPBackendPolicy {
```
robscott, line 159:
```suggestion
func addBackendPolicyIfConfigured(serviceNamespacedName types.NamespacedName, serviceIR *i2gw.ServiceIR) *gkegatewayv1.GCPBackendPolicy {
```

Tiny nit: When I saw the `return nil` a couple lines below this if inputs were empty, I assumed that was an error case that we should be logging. Renaming the function to make it clear that this should only add a backendPolicy if it's specified, could help avoid confusion.

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,188 @@
...
+
+	gkegatewayv1 "github.com/GoogleCloudPlatform/gke-gateway-api/apis/networking/v1"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/ir"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/ingress-gce/pkg/annotations"
+	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
```
robscott, line 31: All these additional imports may be unavoidable, but I'm worried that it could balloon the overall size of gwctl, especially if/when more organizations follow this example and we end up depending on every Ingress + Gateway controller. Do you have an idea of how much work it would be to avoid these imports? Maybe we could just use unstructured types to begin with? Alternatively, if you can show that this PR is not making a meaningful difference to the size of the gwctl binary, it may be a non-issue.

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,188 @@
...
+	"encoding/json"
+
+	gkegatewayv1 "github.com/GoogleCloudPlatform/gke-gateway-api/apis/networking/v1"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/ir"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/ingress-gce/pkg/annotations"
```
robscott, line 30: Very tiny nit: The `annotations` name keeps throwing me off because I think I'm reading the contents of an annotations field, and not some kind of constants, maybe a different name would help:
```suggestion
	gceAnnotations "k8s.io/ingress-gce/pkg/annotations"
```

pkg/i2gw/providers/gce/gce_extensions_test.go
```diff
@@ -0,0 +1,166 @@
...
+						annotations.BetaBackendConfigKey: `{"default":"backendconfig-1"}`,
+					},
+				},
+			},
+			beConfigKey:    annotations.BetaBackendConfigKey,
+			expectedName:   testBeConfigName,
+			expectedExists: true,
+		},
+		{
+			desc: "Service with v1 BackendConfig, using Port Config, not supported",
```
robscott, line 138: Not really any great solution here until we can support ports in targetRefs, but it may be slightly less bad to just pick the first one and translate that, combined with a warning describing what happened.

pkg/i2gw/providers/gce/storage.go
```diff
@@ -17,16 +17,30 @@ limitations under the License.
...
+	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
+	backendconfigv1beta1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1beta1"
 )
 
 type storage struct {
 	Ingresses map[types.NamespacedName]*networkingv1.Ingress
+
+	// BackendConfig is a GKE Ingress extension, and it is associated to an GKE
+	// Ingress through specifying `cloud.google.com/backend-config` or
+	// `beta.cloud.google.com/backend-config` annotation on its Backend Services.
```
robscott, line 32: I thought these annotations were actually equivalent, just that one happened to be newer and without `beta` in the name. I'm not sure which one has precedence, but that should be fairly easy to find in ingress-gce.

robscott (commented):
Thanks for all the great work on this @sawsa307! Mostly focused on GCP section with this review

pkg/i2gw/providers/gce/storage.go
sawsa307: They are both used to point to backend config, but we do need to do query on both annotations because this annotation is added by user in the yaml so we don't have control over which annotation they choose.
As a reference in ingress controller, here is how we are querying with both annotations https://github.com/kubernetes/ingress-gce/blob/master/pkg/annotations/service.go#L443

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: Updated, thanks!

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: Actually I think we can skip this import, and just add imported variable as constance because we are only using `backendConfigKey`, `betaBackendConfigKey`, and the following struct
```
type backendConfigs struct {
	Default string
	Ports   map[string]string
}
```

I think it would be better if we just maintain them ourselves. Updated to include as constant in types.go and gce_extensions.go

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: I think we have to import backendconfigv1, otherwise we won't be able to do read backendConfig from clusters(we need `backendconfigv1.AddToScheme`). However, the other import `k8s.io/ingress-gce/pkg/annotations` we can get rid of that. Described in details in https://github.com/kubernetes-sigs/ingress2gateway/pull/185#discussion_r1744703985.

pkg/i2gw/providers/gce/gce_extensions_test.go
sawsa307: Sure just updated to use the one associated with alphabetically lowest port so we get a consistent behavior.

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,188 @@
...
+
+func toGCPBackendPolicy(serviceNamespacedName types.NamespacedName, serviceIR *i2gw.ServiceIR) *gkegatewayv1.GCPBackendPolicy {
+	if serviceIR.Gce == nil || serviceIR.Gce.SessionAffinity == nil {
+		return nil
+	}
+	affinityType := serviceIR.Gce.SessionAffinity.AffinityType
+	backendPolicy := gkegatewayv1.GCPBackendPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: serviceNamespacedName.Namespace,
+			Name:      serviceNamespacedName.Name + "-GCPBackendPolicy",
```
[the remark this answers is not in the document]
sawsa307: I see we are manually generating names for Gateway and HTTPRoute based on the ingress name and port so I went ahead and generate a name by appending a suffix, given we can only attach one GCPBackendPolicy to a k8s Service. Definitely would like more suggestions here on the naming schema.

pkg/i2gw/providers/gce/gce_extensions.go
LiorLieberman: just to make sure, you meant ingress2gateway binary and not gwctl, right?

pkg/i2gw/ir/provider_istio.go
```diff
@@ -0,0 +1,33 @@
...
+	var mergedGatewayIR *IstioGatewayIR
+	// Populate the value of each fields once IstioGatewayIR is not empty.
+	if existing != nil {
+		mergedGatewayIR = existing
+	}
+	if current != nil {
+		mergedGatewayIR = current
+	}
+	return mergedGatewayIR
+}
```
LiorLieberman, line 33: what is the logic here? cant see how both are merged..

pkg/i2gw/provider.go
LiorLieberman: `type IR`  is still here, cant see the update

pkg/i2gw/ingress2gateway.go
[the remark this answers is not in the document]
LiorLieberman: before, we had `MergeGatewayResources`, right?
This would have merged anything under GatewayResources, including Gateways, GatewayRoutes, HTTPRoutes, etc. (we can see this here on the left panel)

My question is how `maps.Copy(mergedIRs.HTTPRoutes, gr.HTTPRoutes)` would work given that HTTPRouteContext has nested HTTPRouteIR field with its own nested fields. 
Maybe maps.copy is enough for us, idk.. but something to check, I am not sure it will work out of the box

pkg/i2gw/ingress2gateway.go
```diff
@@ -196,5 +203,24 @@ func mergeGateways(gatewaResources []GatewayResources) (map[types.NamespacedName
...
+}
+
+func mergeGatewayIR(current, existing *GatewayContext) {
+	current.Apisix = ir.MergeApisixGatewayIR(current.Apisix, existing.Apisix)
+	current.Gce = ir.MergeGceGatewayIR(current.Gce, existing.Gce)
+	current.IngressNginx = ir.MergeIngressNginxGatewayIR(current.IngressNginx, existing.IngressNginx)
+	current.Istio = ir.MergeIstioGatewayIR(current.Istio, existing.Istio)
+	current.Kong = ir.MergeKongGatewayIR(current.Kong, existing.Kong)
+	current.Openapi3 = ir.MergeOpenapi3GatewayIR(current.Openapi3, existing.Openapi3)
+}
```
LiorLieberman, line 216: I am not sure I understand how all these sub-functions work, it looks like it is either returning existing, current or a null struct.

pkg/i2gw/ingress2gateway.go
```diff
@@ -71,7 +74,9 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 		errs             field.ErrorList
 	)
 	for _, provider := range providerByName {
-		providerGatewayResources, conversionErrs := provider.ToGatewayAPI()
+		ir, conversionErrs := provider.ToIR()
+		errs = append(errs, conversionErrs...)
+		providerGatewayResources, conversionErrs := provider.ToGatewayResources(ir)
```
LiorLieberman, line 79: anchor

pkg/i2gw/ingress2gateway.go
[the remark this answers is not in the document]
sawsa307: They should work. I've tried to run tests on functions using this helper, and they passed with expected merged results.

sawsa307:
Discussed with Lior, since this PR is getting too big, will separate the logic into individual PRs. #188 will be the new PR to include definitions of IRs.

pkg/i2gw/providers/gce/gce_extensions.go
robscott: > just to make sure, you meant ingress2gateway binary and not gwctl, right?

🤦 haha yes, whoops.

> we won't be able to do read backendConfig from clusters(we need backendconfigv1.AddToScheme)

I thought we were already converting to unstructured here. If we use unstructured client to list these, we wouldn't need to call AddToScheme. (That does get annoying to work with, so if the binary isn't massively bigger, maybe we should just leave this as is, just trying to understand the order of magnitude here).

pkg/i2gw/ingress2gateway.go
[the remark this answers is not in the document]
LiorLieberman: Well thats because you have nothing on the http IR yet. similarly to the logic you employ for the GatewayIR merging in https://github.com/kubernetes-sigs/ingress2gateway/pull/188/commits/db033246488ac4489fe484046baf3c7f6ecb7ace you would have to add logic to merge the HTTPIRs and ServiceIR, maps.copy() would not cover that.

pkg/i2gw/providers/gce/gce_extensions.go
robscott: SGTM, I like that approach

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: The issue with unstructured client is, we need to provide namespace when we do the list, like
```
client.Resource(...).Namespace(namespace).List(context.TODO(), metav1.ListOptions{})
```

Though theoretically, the backend/frontend config should live in the same namespace as the services so this information isn't impossible to infer, this could make read gets very messy. So my suggestion is we leave it like this.  

I can look into the size difference in the binary with import, but I don't think it would make a big difference since there isn't much content in the v1 backendconfig package.

sawsa307: I tried to compare the size of the verison built of main and this PR with `ls -l`. The size of binary from main branch version is 72.78MB, while the size of binary from this branch is 75.02MB, which is around 3% increase in size.

pkg/i2gw/providers/gce/gce_extensions.go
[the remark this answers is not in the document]
robscott: That's fair, maybe just remove the `-GCPBackendPolicy` suffix as that's already implied by the type of resource it is.

pkg/i2gw/providers/gce/gce.go
```diff
@@ -32,16 +34,25 @@ func init() {
...
-	reader    reader
-	converter converter
+	storage                       *storage
+	reader                        reader
+	resourcesToIRConverter        resourcesToIRConverter
+	irToGatewayResourcesConverter irToGatewayResourcesConverter
 }
 
 func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
+	// Add BackendConfig and FrontendConfig to Schema when reading in-cluster
```
robscott, line 44: Nit: Comment includes "FrontendConfig" but it looks like you're only adding "BackendConfig" below.

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,201 @@
...
+
+		// Read BackendConfig based on v1 BackendConfigKey.
+		beConfigName, exists := getBackendConfigName(ctx, service, backendConfigKey)
+		if exists {
+			beConfigKey := types.NamespacedName{Namespace: service.Namespace, Name: beConfigName}
+			beConfigToSvcs[beConfigKey] = append(beConfigToSvcs[beConfigKey], svc)
+			continue
+		}
+
+		// Read BackendConfig based on v1beta1 BackendConfigKey.
```
robscott, line 52: Pretty sure you've already covered this, but just making sure - does ingress-gce also give precedence to the v1 key over the v1beta1 one?

pkg/i2gw/providers/gce/gce_extensions.go
```diff
@@ -0,0 +1,201 @@
...
+	return "", false
+}
+
+type backendConfigs struct {
+	Default string            `json:"default,omitempty"`
+	Ports   map[string]string `json:"ports,omitempty"`
+}
+
+// Parse the name of the BackendConfig based on the annotation only if the same
+// BackendConfig is used for all Service ports.
```
robscott, line 89: Nit: This comment is out of date now that you're picking a BackendConfig even if multiple ports are configured.

pkg/i2gw/providers/gce/gce_extensions_test.go
```diff
@@ -0,0 +1,192 @@
...
+		},
+		{
+			desc: "Specify BackendConfig with beta.cloud.google.com/backend-config annotation, using the same BackendConfig for all ports",
+			services: map[types.NamespacedName]*apiv1.Service{
+				{Namespace: testNamespace, Name: testServiceName}: {
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      testServiceName,
+						Namespace: testNamespace,
+						Annotations: map[string]string{
+							betaBackendConfigKey: `{"default":"backendconfig-1"}`,
```
robscott, line 74: Nit: Should have a test with both beta and v1 annotations specified

pkg/i2gw/providers/gce/gce_extensions_test.go
```diff
@@ -0,0 +1,192 @@
...
+			expectedExists: true,
+		},
+		{
+			desc: "Service using beta.cloud.google.com/backend-config annotation, using default Config over all ports",
+			service: &apiv1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      testServiceName,
+					Namespace: testNamespace,
+					Annotations: map[string]string{
+						betaBackendConfigKey: `{"default":"backendconfig-1"}`,
```
robscott, line 155: Same comment as above - should have both new and old annotation

pkg/i2gw/providers/gce/storage.go
```diff
@@ -17,16 +17,26 @@ limitations under the License.
...
 	"k8s.io/apimachinery/pkg/types"
+	backendconfigv1 "k8s.io/ingress-gce/pkg/apis/backendconfig/v1"
 )
 
 type storage struct {
 	Ingresses map[types.NamespacedName]*networkingv1.Ingress
+
+	// BackendConfig is a GKE Ingress extension, and it is associated to an GKE
+	// Ingress through specifying `cloud.google.com/backend-config` or
+	// `beta.cloud.google.com/backend-config` annotation on its Backend Services.
```
robscott, line 31: A few nits here:
1. This comment feels like it belongs directly above BackendConfigs line, not Services.
2. Recommend mentioning that BackendConfig map is keyed by _Service_ namespace/name, not BackendConfig namespace/name. Or I'm wrong here, either way, would be good to clarify.
3. Recommend just saying "Services" instead of "Backend Services".

robscott (approved):
Thanks @sawsa307! This all LGTM, left a few nits but nothing blocking.

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: Sorry I think I missed the previous comment. v1 key has precedence over v1beta1(See [here](https://github.com/kubernetes/ingress-gce/blob/2c4530cdd9e9597e0044800c2c95039cbc0f2e73/pkg/annotations/service.go#L443)), and we are doing the same thing here: If we found with v1 key on this service, we stop looking for v1beta1 key and move on to the next one, which is consistent.

pkg/i2gw/providers/gce/storage.go
sawsa307: Updated, thanks for the suggestion!

pkg/i2gw/providers/gce/gce_extensions_test.go
sawsa307: Added the test cases, thanks!

pkg/i2gw/providers/gce/gce_extensions_test.go
sawsa307: Added the test cases, thanks!

pkg/i2gw/providers/gce/gce_extensions.go
sawsa307: Updated comment, thanks!

pkg/i2gw/providers/gce/gce.go
sawsa307: Thanks for catching this, updating the comment.

sawsa307:
Closing in favor of #188.

PR #187, opened 2024-08-26
Create Ingress2gateway version annotation and attach it to Gateway.

### LINKED ISSUE #179: Find a way to attribute Gateways converted via Ingress2Gateway.
Find a way to attribute the number of existing deployed Gateways that are converted via Ingress2Gateway.
One possible solution is to attach a custom label/annotation in the translated Gateway. The name of the tool and the version would be useful information to be included.

**Why this is needed**:
Keep track of resources that we're generated with ingress2gateway.

sawsa307:
* Ingress2gateway annotation will track the current version of the ingress2gateway tool. The version is currently a string that needs to be updated manually before release.
* This annotation will be attached to the resulting Gateway during print since this is an output-only translation, and we can avoid interfering with existing Gateway translation tests.

**What type of PR is this?**
/kind feature

Create a dedicate annotation for Gateways translated via ingress2gateway tool, so each provider can perform a census on the number of Gateways translated via ingress2gateway.

Fixes #179

```release-note
Gateways translated via ingress2gateway will be attached with a new annotation `kubernetes.io/ingress2gateway.version` to track the current version of the tool.
```


sawsa307:
Would like some suggestion on the preferred format of this kind of annotation. I created the current one based on how ingressnginx specify proxy http version https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#proxy-http-version.
Two alternatives I can think of are:
1. `gateway.kubernetes.io/ingress2gateway.version`
2. `ingress2gateway.gateway.kubernetes.io/version`

pkg/i2gw/ingress2gateway.go
```diff
@@ -31,6 +31,10 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
+const Ingress2GatewayVersionAnnotationKey = "kubernetes.io/ingress2gateway.version"
```
robscott, line 34:
```suggestion
const Ingress2GatewayGeneratorAnnotationKey = "gateway.networking.k8s.io/generator"
```

robscott (commented):
Thanks @sawsa307!

pkg/i2gw/ingress2gateway.go
sawsa307: Updated.

cmd/print.go
```diff
@@ -117,6 +117,10 @@ func (pr *PrintRunner) outputResult(gatewayResources []i2gw.GatewayResources) {
 		resourceCount += len(r.Gateways)
 		for _, gateway := range r.Gateways {
 			gateway := gateway
+			if gateway.Annotations == nil {
+				gateway.Annotations = make(map[string]string)
+			}
+			gateway.Annotations[i2gw.Ingress2GatewayVersionAnnotationKey] = i2gw.CurrentVersion
```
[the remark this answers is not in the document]
sawsa307: Updated.

levikobi (commented):
Hi @sawsa307!
Thank you for your contribution!
I have one small comment in case we are going for this change. Since this feature requires manual version modification upon a new release, please update the `RELEASE.md` file with the appropriate steps.

sawsa307:
> Hi @sawsa307! Thank you for your contribution! I have one small comment in case we are going for this change. Since this feature requires manual version modification upon a new release, please update the `RELEASE.md` file with the appropriate steps.

Thanks for the comment Kobi! I've just added a section in RELEASE.md. Let me know if you have any suggestions on it, thank you!

LiorLieberman:
Is there a reason we do it only for gateway? 
Whats the rationale behind it?

sawsa307:
> Is there a reason we do it only for gateway? Whats the rationale behind it?

We could attach this in other resources like HTTPRoute, but I don't think we can gain any information with it since we can find all translated HTTPRoute based on the Gateways link to them.

robscott:
> Is there a reason we do it only for gateway? Whats the rationale behind it?

Actually I think this is a good point - might as well add this annotation to all generated resources so it's both more consistent and easier to keep track of where and GW API resource came from.

sawsa307:
> > Is there a reason we do it only for gateway? Whats the rationale behind it?
> 
> Actually I think this is a good point - might as well add this annotation to all generated resources so it's both more consistent and easier to keep track of where and GW API resource came from.

Sure, just attaches this annotation on all translated Gateway resources.

LiorLieberman:
> @sawsa307 The release note mention `kubernetes.io/ingress2gateway.version` and the code `"gateway.networking.k8s.io/generator"` can you align them please?

I updated the release note myself to unblock.

/lgtm Thanks David!

PR #193, opened 2024-09-16
Refactor GCE test code

sawsa307:
**What type of PR is this?**
/kind cleanup

Refactor GCE test codes.

None

```release-note
None
```


LiorLieberman (commented):
ping me to lgtm after you rebase

pkg/i2gw/providers/gce/ir_converter_test.go
```diff
@@ -1084,3 +959,252 @@ func Test_convertToIR(t *testing.T) {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      testServiceName,
+						Namespace: testNamespace,
+						Annotations: map[string]string{
+							backendConfigKey: `{"default":"backendconfig-1"}`,
+						},
+					},
+				},
+			},
+			expectedServices: expectedServices,
```
robscott, line 1035: Why do all test cases have the same expected Services? What is this test actually testing?

pkg/i2gw/providers/gce/ir_converter_test.go
```diff
@@ -1084,3 +959,252 @@ func Test_convertToIR(t *testing.T) {
...
+					Name:      testServiceName,
+					Namespace: testNamespace,
+					Annotations: map[string]string{
+						backendConfigKey: `{"default":"backendconfig-1"}`,
+					},
+				},
+			},
+			beConfigKey:    backendConfigKey,
+			expectedName:   testBeConfigName,
+			expectedExists: true,
```
robscott, line 1164: Seems like we'd need at least one test case where this is false

pkg/i2gw/providers/gce/ir_converter_test.go
```diff
@@ -36,25 +38,33 @@ import (
...
+	testBackendConfigName = "test-backendconfig"
 
-	gPathPrefix := gatewayv1.PathMatchPathPrefix
-	gExact := gatewayv1.PathMatchExact
+	saTypeClientIP = "CLIENT_IP"
+	saTypeCookie   = "GENERATED_COOKIE"
+)
 
+var (
+	// These variables are referenced by pointers.
```
robscott, line 52: Can simplify a lot of this with a func like https://github.com/kubernetes-sigs/gateway-api/blob/b5e115d06513555c7bc81bf7535f896105aac450/pkg/test/cel/main_test.go#L60-L62.

robscott (commented):
Thanks @sawsa307! Just a few questions, otherwise LGTM

pkg/i2gw/providers/gce/ir_converter_test.go
sawsa307: It is testing: the different ways of using BackendConfig annotation on a Service, so the service doesn't need to be changed.

robscott: Generally with test cases like this I'd expect an input and expected output var. If this expected output var is not useful, what would be?

pkg/i2gw/providers/gce/ir_converter_test.go
sawsa307: Added.

pkg/i2gw/providers/gce/ir_converter_test.go
sawsa307: Yes that makes sense, just updated the test case.

pkg/i2gw/providers/gce/ir_converter_test.go
sawsa307: Updated, thanks!

robscott (approved):
Thanks @sawsa307!

/approve

PR #216, opened 2025-04-18
Add version command

### LINKED ISSUE #213: Print version number of ingress2gateway

When running the ingress2gateway binary, it would be nice for it to print the version number

**Why this is needed**:

I installed ingress2gateway via brew, and I wasn't sure which version I was using. I had a look around and struggled to find it.
Being able to print the version number would be useful.

Additionally, if this tool got more popular and more support requests are created, it may be useful for maintainers to know which version of the tool someone needs support for.

spencerhance:
**What type of PR is this?**
/kind feature

Adds a `version` command that can be used to display the current binary version.

This is useful for users to be able to tell what version they are running, especially if installing via a package manager like homebrew.

Example:

    ./ingress2gateway version
    ingress2gateway version: v0.4.0
    Built with Go version: go1.24.2

* This is implemented via a combination of ldflags (to propogate the git tag info) and debug.BuildInfo (to get the golang version).
* This does not modify the CurrentVersion var in the i2gw package, but plan is to migrate that in a follow-up.

Fixes #213

```release-note
The version of the binary can now be printed with the `version` command (e.g. ingress2gateway version)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,19 @@
 # Enable Go modules.
 export GO111MODULE=on
 
+# CMDPKG is the package path for cmd.  This allows us to propogate the git info
+# to the binary via LDFLAGS.
+CMDPKG := $(shell go list .)/cmd
+
+# Get the version string from git describe.
+# --tags: Use annotated tags.
+# --always: Fallback to commit hash if no tag is found.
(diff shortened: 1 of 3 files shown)


robscott:
This is great, thanks @spencerhance!

/lgtm
/approve

PR #217, opened 2025-04-18
Refactor CurrentVersion to use the git version string

spencerhance:
**What type of PR is this?**
/kind cleanup

* Previously the git version string was plumbed to cmd.Version via the linker and i2gw had a hardcoded version
* Now the linker propogates the version string to i2gw and the cmd package imports it from there
* Renames `CurrentVersion` to just `Version`
* This has a few effects:
  1. You no longer need to manually update i2gw.CurrentVersion whenever there is a release.  Simply tagging the release will work.
  2. The generator annotations (gateway.networking.k8s.io/generator) will now use the git version string instead instead of just the hardcoded value.  This means that during local development you will see something like `"ingress2gateway-v0.4.0-5-gabcdef-dirty"` in your generated annotations, letting you know exactly what local commit you were using when you generated that Gateway.  However, end users who download a release or install via a package manager should **not** see any difference in behavior.

This PR is a follow-up to #216

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,9 @@
 # Enable Go modules.
 export GO111MODULE=on
 
-# CMDPKG is the package path for cmd.  This allows us to propogate the git info
+# I2GWPKG is the package path for i2gw.  This allows us to propogate the git info
 # to the binary via LDFLAGS.
-CMDPKG := $(shell go list .)/cmd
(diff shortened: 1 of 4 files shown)


robscott:
Thanks @spencerhance!

/lgtm
/approve

PR #218, opened 2025-04-18
Fix print command instructions in README

spencerhance:
**What type of PR is this?**
/kind documentation

Fixes the example `print` command in the README, which currently will print this:

```
./ingress2gateway print
Error: required flag(s) "providers" not set
Usage:
  ingress2gateway print [flags]
...
```

* v0.3.0 made the `--providers` print command required, however the README wasn't updated
* Updates the example print command to use --providers=ingress-nginx since at least one provider *must* be specified
* Removes the default option of providers in the table in the README since it no longer exists.

```release-note
NONE
```


robscott:
Thanks @spencerhance!

/lgtm
/approve

Issue #162, opened 2024-06-13
Automatically generate the list of flags in the documentation


mlavacca:
**What would you like to be added**:

We have a list of flags in our README that needs to be manually updated every time a flag gets added, removed, or changed. We may want to automatically generate such a list of flags to avoid misalignment between the actual flags and our docs

/kind documentation

LiorLieberman:
SGTM!
/triage accepted

levikobi:
Sounds great👍🏻

spencerhance:
Took a look at this and it's actually quite complicated because Cobra doesn't allow you to do custom formatting when generating docs - I have a working prototype though and can send a PR to get feedback.

icohangar-ops:
/assign

I'm implementing this by auto-generating the flags tables in the README from the cobra command definitions, wired into the build via `make gen-flags` and verified in CI via a `hack/verify-flags.sh` check. PR incoming shortly.

PR #219, opened 2025-04-20
Sort the names of the names of the providers in the print command

spencerhance:
**What type of PR is this?**

Add one of the following kinds:
/kind cleanup

* Previously the help string for `print --providers` had a non-deterministic ordering since it is generated from map keys.
* This will be helpful in tackling #162

*Before (point in time example)*
`
      --providers strings                    If present, the tool will try to convert only resources related to the specified providers, supported values are [kong openapi3 apisix cilium gce ingress-nginx istio].
`

*After*
`
      --providers strings                    If present, the tool will try to convert only resources related to the specified providers, supported values are [apisix cilium gce ingress-nginx istio kong openapi3].
`

```release-note
NONE
```

### DIFF
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -19,6 +19,7 @@ package i2gw
 import (
 	"context"
 	"fmt"
+	"sort"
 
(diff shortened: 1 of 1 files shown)


LiorLieberman:
Looks good to me. 

/lgtm
/approve

PR #220, opened 2025-04-20
Automatically generate flag docs in README

### LINKED ISSUE #162: Automatically generate the list of flags in the documentation

spencerhance:
**What type of PR is this?**
/kind documentation
/kind feature

Automatically generate the docs for the entire CLI surface and inject it into the README.

This PR is somewhat large so i've separated it into 3 commits so it's easier to review.  It also involved a fair amount of *vibe coding* so i'm not particularly attached to the implementation 😄 

There are a few new things being introduced here:

1. A _hidden_ `generate-docs` command which does the heavy lifting of parsing the Cobra command tree and generating the relevant sections and tables.  Cobra actually has some built in logic by way of `GenMarkdownTree()`, but after messing with it a bunch I don't think it's a good fit here.  Mainly it generates one file per command+subcommand, and it doesn't allow you to do any custom formatting (like a table).

2. A `hack/generate-cli-docs.sh` script which runs `generate-docs` and then injects it into the README using some begin/end markers specified via HTML comments.

3. a `hack/verify-readme.sh` script which verifies that the README is up-to-date when contributors submit a PR.

NOTE: This PR also depends on #219 being merged (but currently that diff is included here).

Fixes #162 

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ brew install ingress2gateway
    make build
    ```
 
-## Usage
+## Basic Usage
 
 Ingress2gateway reads Ingress resources and/or provider-specifc CRDs from a Kubernetes
 cluster or a file. It will output the equivalent Gateway API resources in a YAML/JSON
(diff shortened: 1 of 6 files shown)


spencerhance:
cc: @mlavacca @robscott

spencerhance:
Looks like I may need to change some CI stuff since `verify-readme.sh` requires the ingress2gateway binary...

PR #222, opened 2025-05-05
Add support for named ports

### LINKED ISSUE #209: Error for named ports
**What happened**:

I executed `./ingress2gateway print --providers ingress-nginx` and got an error:

```
# Encountered 3 errors # paths.backends[0].service.port: Invalid value: "name": named ports not supported: http # paths.backends[0].service.port: Invalid value: "name": named ports not supported: http # paths.backends[0].service.port: Invalid value: "name": named ports not supported: http
```

I expected a successful conversion from Ingress to Gateway.

Deploy an nginx and an application with a named port in your k8s cluster.

gavinkflam:
**What type of PR is this?**

/kind feature

Support converting ingress rules with named ports.

Fixes #209

The proposed change includes the following steps:

1. Read service definitions at the same time as we read ingresses. (resource readers)
2. Construct a mapping from port names to port numbers for each service.
3. Use this mapping to resolve port name references into port numbers.

```release-note
Add support for named ports
```

### DIFF
--- a/pkg/i2gw/providers/common/utils.go
+++ b/pkg/i2gw/providers/common/utils.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"regexp"
 
+	apiv1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	networkingv1beta1 "k8s.io/api/networking/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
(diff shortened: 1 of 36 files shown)


pkg/i2gw/providers/common/utils.go
```diff
@@ -105,16 +107,27 @@ func RouteName(ingressName, host string) string {
...
 			fieldPath := path.Child("service", "port")
-			return nil, field.Invalid(fieldPath, "name", fmt.Sprintf("named ports not supported: %s", ib.Service.Port.Name))
+			return nil, field.Invalid(fieldPath, "name", fmt.Sprintf("cannot find port with name %s in service %s", ib.Service.Port.Name, ib.Service.Name))
 		}
+
 		return &gatewayv1.BackendRef{
 			BackendObjectReference: gatewayv1.BackendObjectReference{
 				Name: gatewayv1.ObjectName(ib.Service.Name),
-				Port: (*gatewayv1.PortNumber)(&ib.Service.Port.Number),
+				Port: (*gatewayv1.PortNumber)(&portNumber),
```
LiorLieberman, line 130: To make sure I understand correctly, you are reading ingress configs and creating a map between service port names and port numbers? then you retrieve it here?

LiorLieberman (commented):
Thanks for picking this up @gavinkflam !

Mind updating the PR description with the logic you added?

Also, cross ref https://github.com/kubernetes-sigs/gateway-api/issues/2420 for why named ports are not supported.

pkg/i2gw/providers/common/utils.go
gavinkflam: The proposed change includes the following steps:

1. Read service definitions at the same time as we read ingresses. (resource readers)
2. Construct a mapping from port names to port numbers for each service.
3. Use this mapping to resolve port name references into port numbers here.

gavinkflam:
Thanks @LiorLieberman! I've updated the description to include an explanation of the implementation.

So upstream choose not to support named ports consciously. If that’s the case, it becomes even more important for us to support automatic migration via ingress2gateway.

LiorLieberman (commented):
/lgtm
/approve

/hold in case @robscott have other opinions

robscott:
Thanks @gavinkflam!

/hold cancel

PR #227, opened 2025-07-17
Add server-alias support for ingress-nginx provider

adrianmoisey:
**What type of PR is this?**
/kind feature

Adds support for the `nginx.ingress.kubernetes.io/server-alias` ingress-nginx annotation 

Fixes #

```release-note
Add support for ingress-nginx server-alias annotation
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/README.md
+++ b/pkg/i2gw/providers/ingressnginx/README.md
@@ -11,6 +11,6 @@ Current supported annotations:
 - `nginx.ingress.kubernetes.io/canary-by-header-pattern`: If specified, this is the pattern to match against for the HTTPHeaderMatch, which will be of type HeaderMatchRegularExpression.
 - `nginx.ingress.kubernetes.io/canary-weight`: If specified and non-zero, this value will be applied as the weight of the backends for the routes generated from this Ingress resource.
 - `nginx.ingress.kubernetes.io/canary-weight-total`: If specified, this value will be used as the total weight for canary deployments.
-- `nginx.ingress.kubernetes.io/server-alias`: If specified, this value will be added as additional hostnames to the Gateway listeners and HTTPRoutes. Supports comma-separated values for multiple aliases.
+- `nginx.ingress.kubernetes.io/server-alias`: If specified, this value will be added as additional hostnames to the HTTPRoutes. Supports comma-separated values for multiple aliases.
 
 If you are reliant on any annotations not listed above, please open an issue. In the meantime you'll need to manually find a Gateway API equivalent.
--- a/pkg/i2gw/providers/ingressnginx/README.md
+++ b/pkg/i2gw/providers/ingressnginx/README.md
@@ -11,6 +11,6 @@ Current supported annotations:
 - `nginx.ingress.kubernetes.io/canary-by-header-pattern`: If specified, this is the pattern to match against for the HTTPHeaderMatch, which will be of type HeaderMatchRegularExpression.
 - `nginx.ingress.kubernetes.io/canary-weight`: If specified and non-zero, this value will be applied as the weight of the backends for the routes generated from this Ingress resource.
 - `nginx.ingress.kubernetes.io/canary-weight-total`: If specified, this value will be used as the total weight for canary deployments.
-- `nginx.ingress.kubernetes.io/server-alias`: If specified, this value will be added as additional hostnames to the HTTPRoutes. Supports comma-separated values for multiple aliases.
+- `nginx.ingress.kubernetes.io/server-alias`: If specified, this value will be added as additional hostnames to the HTTPRoutes. Supports comma-separated values for multiple aliases. Note: ingress2gateway will create an empty Gateway Listener, and will add all the values in server-alias annotation to the generated HTTPRoute.spec.hostnames
 
 If you are reliant on any annotations not listed above, please open an issue. In the meantime you'll need to manually find a Gateway API equivalent.
--- a/pkg/i2gw/providers/ingressnginx/README.md
+++ b/pkg/i2gw/providers/ingressnginx/README.md
@@ -10,6 +10,7 @@ Current supported annotations:
 - `nginx.ingress.kubernetes.io/canary-by-header-value`: If specified, the value of this annotation is the header value to perform an HeaderMatchExact match on in the generated HTTPHeaderMatch.
 - `nginx.ingress.kubernetes.io/canary-by-header-pattern`: If specified, this is the pattern to match against for the HTTPHeaderMatch, which will be of type HeaderMatchRegularExpression.
 - `nginx.ingress.kubernetes.io/canary-weight`: If specified and non-zero, this value will be applied as the weight of the backends for the routes generated from this Ingress resource.
-`nginx.ingress.kubernetes.io/canary-weight-total`
+- `nginx.ingress.kubernetes.io/canary-weight-total`: If specified, this value will be used as the total weight for canary deployments.
+- `nginx.ingress.kubernetes.io/server-alias`: If specified, this value will be added as additional hostnames to the Gateway listeners and HTTPRoutes. Supports comma-separated values for multiple aliases.
 
 If you are reliant on any annotations not listed above, please open an issue. In the meantime you'll need to manually find a Gateway API equivalent.
--- a/pkg/i2gw/providers/ingressnginx/notification.go
+++ b/pkg/i2gw/providers/ingressnginx/notification.go
@@ -21,7 +21,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func notify(mType notifications.MessageType, message string, callingObject ...client.Object) { //nolint:unparam
-	newNotification := notifications.NewNotification(mType, message, callingObject...)
+func notify(message string, callingObject ...client.Object) {
+	newNotification := notifications.NewNotification(notifications.InfoNotification, message, callingObject...)
 	notifications.NotificationAggr.DispatchNotification(newNotification, string(Name))
 }
--- a/pkg/i2gw/providers/ingressnginx/notification.go
+++ b/pkg/i2gw/providers/ingressnginx/notification.go
@@ -21,7 +21,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func notify(mType notifications.MessageType, message string, callingObject ...client.Object) {
+func notify(mType notifications.MessageType, message string, callingObject ...client.Object) { //nolint:unparam
 	newNotification := notifications.NewNotification(mType, message, callingObject...)
 	notifications.NotificationAggr.DispatchNotification(newNotification, string(Name))
 }
--- a/pkg/i2gw/providers/ingressnginx/server_alias.go
+++ b/pkg/i2gw/providers/ingressnginx/server_alias.go
@@ -22,7 +22,6 @@ import (
 	"strings"
 
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/intermediate"
-	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/common"
 
 	networkingv1 "k8s.io/api/networking/v1"
@@ -63,8 +62,9 @@ func serverAliasFeature(ingresses []networkingv1.Ingress, _ map[types.Namespaced
 				}
 			}
 
-			notify(notifications.InfoNotification,
-				fmt.Sprintf("parsed server-alias annotation with %d aliases: %s", len(aliases), strings.Join(aliases, ", ")), &ingress)
+			notify(
+				fmt.Sprintf("parsed server-alias annotation with %d aliases: %s", len(aliases), strings.Join(aliases, ", ")), &ingress,
+			)
 		}
 	}
 
@@ -81,9 +81,10 @@ func serverAliasFeature(ingresses []networkingv1.Ingress, _ map[types.Namespaced
 
 					if !exists {
 						routeContext.Spec.Hostnames = append(routeContext.Spec.Hostnames, hostname)
-						notify(notifications.InfoNotification,
+						notify(
 							fmt.Sprintf("added server alias hostname \"%s\" to HTTPRoute \"%s\" for Gateway \"%s\"", alias, routeContext.Name, gatewayKey.Name),
-							&routeContext.HTTPRoute)
+							&routeContext.HTTPRoute,
+						)
 
 						// Update the route context in the IR
 						ir.HTTPRoutes[routeKey] = routeContext
--- a/pkg/i2gw/providers/ingressnginx/server_alias.go
+++ b/pkg/i2gw/providers/ingressnginx/server_alias.go
@@ -1,0 +1,136 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ingressnginx
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/intermediate"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/common"
+
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// serverAliasFeature processes the server-alias annotation for ingress-nginx.
+// https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-alias
+func serverAliasFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, ir *intermediate.IR) field.ErrorList {
+	var errs field.ErrorList
+
+	// Keep track of server aliases per gateway
+	serverAliasesByGateway := make(map[types.NamespacedName][]string)
+
+	for _, ingress := range ingresses {
+		if serverAlias, ok := ingress.Annotations["nginx.ingress.kubernetes.io/server-alias"]; ok && serverAlias != "" {
+			aliases := strings.Split(serverAlias, ",")
+
+			// Determine which gateway this ingress belongs to
+			// For ingress-nginx, the gateway name is typically the ingress class name
+			ingressClass := common.GetIngressClass(ingress)
+			if ingressClass == "" {
+				ingressClass = "nginx" // default for ingress-nginx
+			}
+
+			gatewayKey := types.NamespacedName{Namespace: ingress.Namespace, Name: ingressClass}
+
+			// Collect all unique aliases for this gateway
+			existingAliases := serverAliasesByGateway[gatewayKey]
+			for _, alias := range aliases {
+				alias = strings.TrimSpace(alias)
+				// Check if alias already exists for this gateway
+				exists := slices.Contains(existingAliases, alias)
+				if !exists {
+					serverAliasesByGateway[gatewayKey] = append(serverAliasesByGateway[gatewayKey], alias)
+				}
+			}
+
+			notify(notifications.InfoNotification,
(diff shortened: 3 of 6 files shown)


adrianmoisey:
/hold

I can't get golangci-lint working locally. Let me get this all sorted first.

pkg/i2gw/providers/ingressnginx/notification.go
```diff
@@ -21,7 +21,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func notify(mType notifications.MessageType, message string, callingObject ...client.Object) {
+func notify(mType notifications.MessageType, message string, callingObject ...client.Object) { //nolint:unparam
```
LiorLieberman: why is this needed?

pkg/i2gw/providers/ingressnginx/server_alias_test.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+
+			// Compare HTTPRoutes
+			if diff := cmp.Diff(tc.expectedIR.HTTPRoutes, ir.HTTPRoutes); diff != "" {
+				t.Fatalf("HTTPRoutes mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+func TestGetIngressClass(t *testing.T) {
```
LiorLieberman: why is this in server_alias_test.go? was it supposed to be on #231 ?

pkg/i2gw/providers/ingressnginx/README.md
```diff
@@ -10,6 +10,7 @@ Current supported annotations:
 - `nginx.ingress.kubernetes.io/canary-by-header-value`: If specified, the value of this annotation is the header value to perform an HeaderMatchExact match on in the generated HTTPHeaderMatch.
 - `nginx.ingress.kubernetes.io/canary-by-header-pattern`: If specified, this is the pattern to match against for the HTTPHeaderMatch, which will be of type HeaderMatchRegularExpression.
 - `nginx.ingress.kubernetes.io/canary-weight`: If specified and non-zero, this value will be applied as the weight of the backends for the routes generated from this Ingress resource.
-`nginx.ingress.kubernetes.io/canary-weight-total`
+- `nginx.ingress.kubernetes.io/canary-weight-total`: If specified, this value will be used as the total weight for canary deployments.
+- `nginx.ingress.kubernetes.io/server-alias`: If specified, this value will be added as additional hostnames to the Gateway listeners and HTTPRoutes. Supports comma-separated values for multiple aliases.
```
LiorLieberman: I am wondering if we have to generate 3 distinct listeners. What is the behavior of `nginx.ingress.kubernetes.io/server-alias` in ingress-nginx - does it create separate listeners?

/cc @rikatz

rikatz: no, it adds the server aliases to the same listener: https://github.com/kubernetes/ingress-nginx/blob/main/rootfs/etc/nginx/template/nginx.tmpl#L639

rikatz: (btw I will try to take a look on this on Monday!)

pkg/i2gw/providers/ingressnginx/notification.go
adrianmoisey: Prior to this change, the `ingressnginx.notify` function wasn't used, so the linting was ignoring it.
Then I made this PR to start using it, and this lint starts happening:

```
pkg/i2gw/providers/ingressnginx/notification.go:24:13: `notify` - `mType` always receives `notifications.InfoNotification` (`"INFO"`) (unparam)
func notify(mType notifications.MessageType, message string, callingObject ...client.Object) {
            ^
```

I figured it made sense to keep this function as-is, as it may be used in other ways in the future.

ref: https://github.com/kubernetes-sigs/ingress2gateway/pull/227/commits/48d501fc0864b8229d78b580438bd953c07ba417

pkg/i2gw/providers/ingressnginx/server_alias_test.go
adrianmoisey: It's to test https://github.com/kubernetes-sigs/ingress2gateway/pull/227/files#diff-390c4deeef1d9cf7dca46e3444ce4565ca9057fbf900ce1c4740cfdfcffa5ab9R137-R146

pkg/i2gw/providers/ingressnginx/README.md
adrianmoisey: I'm not too sure on the right way to handle these in Gateway API, so some guidance will be appreciated.

rikatz: I see where you are coming from @adrianmoisey 

@LiorLieberman yesterday I missed some clarification before answering, sorry. When you asked about "listener" I assumed it was an nginx listener, and not a gateway listener entry on a gatewayapi resource. 

While the answer for the first is right (nginx creates aliases on a single listener entry), I was thinking that today there is no way, on gw api, to say "I want this .spec.listener[] to have aliases", so I guess the only right way of doing it is appending listeners to the list with the same configuration and different hosts, and leaving for the backend to consolidate it. 

Which probably falls back into the need of ListenerSet and maybe some better allowance of the app developer to define a single entry and aliases to it. 

I am planning on spending more time next week on improving ListenerSet gep, and this is for sure a case we need to think of

pkg/i2gw/providers/ingressnginx/server_alias.go
```diff
@@ -0,0 +1,164 @@
...
+				}
+			}
+		}
+	}
+
+	return errs
+}
+
+// getIngressClass returns the ingress class for the ingress
+func getIngressClass(ingress networkingv1.Ingress) string {
```
rikatz: You can probably use https://github.com/kubernetes-sigs/ingress2gateway/blob/312bd894ba8553c566eecf2fecbbb34451ae6b29/pkg/i2gw/providers/common/utils.go#L31 instead of duplicating this function

pkg/i2gw/providers/ingressnginx/server_alias.go
```diff
@@ -0,0 +1,164 @@
...
+	if ingress.Spec.IngressClassName != nil && *ingress.Spec.IngressClassName != "" {
+		return *ingress.Spec.IngressClassName
+	}
+	if class, ok := ingress.Annotations["kubernetes.io/ingress.class"]; ok {
+		return class
+	}
+	return ""
+}
+
+// parseServerAliases parses the server-alias annotation value which can contain
```
rikatz: but the use of this function is different right? IIUC what it does is just to trim spaces from each item on the array, and create a new array with it.

As you call it again on https://github.com/kubernetes-sigs/ingress2gateway/pull/227/files#diff-390c4deeef1d9cf7dca46e3444ce4565ca9057fbf900ce1c4740cfdfcffa5ab9R56 only, I am wondering if you really need a whole helper function for it, or just doing a strings.TrimSpace on the line getting all of the items from the array isn't more clear.

Just my 20c :)

pkg/i2gw/providers/ingressnginx/server_alias.go
```diff
@@ -0,0 +1,164 @@
...
+// serverAliasFeature processes the server-alias annotation for ingress-nginx.
+// https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-alias
+func serverAliasFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, ir *intermediate.IR) field.ErrorList {
+	var errs field.ErrorList
+
+	// Keep track of server aliases per gateway
+	serverAliasesByGateway := make(map[types.NamespacedName][]string)
+
+	for _, ingress := range ingresses {
+		if serverAlias := ingress.Annotations["nginx.ingress.kubernetes.io/server-alias"]; serverAlias != "" {
```
rikatz: a nit:

```suggestion
		if serverAlias, ok := ingress.Annotations["nginx.ingress.kubernetes.io/server-alias"]; ok && serverAlias != "" {
```

The reason is so you check first if the key on the map exists, and then you check its value

adrianmoisey: Thanks, I've added that

pkg/i2gw/providers/ingressnginx/server_alias.go
adrianmoisey: 🤦 
Yup, and I've found other useful functions in common too, thanks for that.

pkg/i2gw/providers/ingressnginx/server_alias.go
adrianmoisey: Yeah, that simplifies it, thanks

pkg/i2gw/providers/ingressnginx/README.md
adrianmoisey: What's the path forward here?
Should we merge this as is, or wait for ListenerSet? I assume adding aliases in a neat way is something most Gateway API users would want.

LiorLieberman: /cc @robscott as well for opinions

robscott: Couldn't we just leave `hostname` empty on the Listener and instead specify `hostnames` on the HTTPRoute to mirror Ingress behavior?

adrianmoisey: I've pushed a change that does that. 

Given this input:
```
$ cat example.yml
kind: Ingress
metadata:
  name: testing
  namespace: default
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
    nginx.ingress.kubernetes.io/server-alias: adrian.com,api.adrian.com,cdn.adrian.com
spec:
  ingressClassName: nginx
  rules:
  - http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: test
            port:
              number: 80
```

I get this output:
```
$ ./ingress2gateway print --input-file example.yml --providers=ingress-nginx
Notifications from INGRESS-NGINX:
+--------------+-------------------------------------------------------------------------------------------+--------------------------------------+
| MESSAGE TYPE |                                       NOTIFICATION                                        |            CALLING OBJECT            |
+--------------+-------------------------------------------------------------------------------------------+--------------------------------------+
| INFO         | parsed server-alias annotation with 3 aliases: adrian.com, api.adrian.com, cdn.adrian.com | Ingress: default/testing             |
+--------------+-------------------------------------------------------------------------------------------+--------------------------------------+
| INFO         | added server alias hostname to HTTPRoute: adrian.com                                      | HTTPRoute: default/testing-all-hosts |
+--------------+-------------------------------------------------------------------------------------------+--------------------------------------+
| INFO         | added server alias hostname to HTTPRoute: api.adrian.com                                  | HTTPRoute: default/testing-all-hosts |
+--------------+-------------------------------------------------------------------------------------------+--------------------------------------+
| INFO         | added server alias hostname to HTTPRoute: cdn.adrian.com                                  | HTTPRoute: default/testing-all-hosts |
+--------------+-------------------------------------------------------------------------------------------+--------------------------------------+

kind: Gateway
metadata:
  annotations:
  creationTimestamp: null
  name: nginx
  namespace: default
spec:
  gatewayClassName: nginx
  listeners:
  - name: http
    port: 80
    protocol: HTTP
status: {}
---
kind: HTTPRoute
metadata:
  annotations:
  creationTimestamp: null
  name: testing-all-hosts
  namespace: default
spec:
  hostnames:
  - adrian.com
  - api.adrian.com
  - cdn.adrian.com
  parentRefs:
  - name: nginx
  rules:
  - backendRefs:
    - name: test
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /
status:
  parents: []
```

That look good to you?

LiorLieberman: it is better to me, the question is whether it is sufficiently mimic the behavior of ingress-nginx. Like of course there is a difference if the traffic is denied on the listener level or the route level. 

I am good to merge this with potentially a caveat on this logic we are taking. 
@rikatz - any thoughts/concerns with the above?

pkg/i2gw/providers/ingressnginx/notification.go
rikatz: As this is a private function, I think it makes sense to avoid linter problems and change its signature for now. But this is just my feeling, I prefer not having any "nolint" and we forgetting it, instead of having it and forgetting in the future.

pkg/i2gw/providers/ingressnginx/server_alias.go
```diff
@@ -0,0 +1,96 @@
...
+				alias = strings.TrimSpace(alias)
+				// Check if alias already exists for this gateway
+				exists := slices.Contains(existingAliases, alias)
+				if !exists {
+					serverAliasesByGateway[gatewayKey] = append(serverAliasesByGateway[gatewayKey], alias)
+				}
+			}
+
+			notify(notifications.InfoNotification,
+				fmt.Sprintf("parsed server-alias annotation with %d aliases: %s", len(aliases), strings.Join(aliases, ", ")), &ingress)
```
rikatz: This is sending the log to stdout, so when I do something like "ingress2gateway" > file.yaml it gets added to the final manifest

rikatz: ok fixed on https://github.com/kubernetes-sigs/ingress2gateway/pull/233

pkg/i2gw/providers/ingressnginx/README.md
rikatz: I was testing it locally here, but it is not tbh. If you don't add the hostname to Gateway listeners, it will not work.

The .spec.hostname will tell you what hostnames that HTTPRoute answers to, but if the same hostnames are not on Gateway it will drop the request.

I have tested it on Envoy Gateway

adrianmoisey: > I was testing it locally here, but it is not tbh. If you don't add the hostname to Gateway listeners, it will not work.

I've tested with kgateway, and without the hostname on the Gateway listener, it does work.

What is the path forward here? Does Gateway API need to have a strong opinion on this?

adrianmoisey: > I was testing it locally here, but it is not tbh. If you don't add the hostname to Gateway listeners, it will not work.

I tested with Envoy Gateway too, and it did work without the hostname set in the Gateway listener.

There is also an example listed on this page: https://gateway.envoyproxy.io/docs/install/install-helm/ (the https://github.com/envoyproxy/gateway/releases/download/v1.5.0/quickstart.yaml part) that has a Gateway without the hostname, and it also works.

rikatz: hum seriously? I may have messed with something then, sorry. I was testing with envoy gateway, maybe something was wrong.

I will take a further look tomorrow again, as today have some couple of things to run :)

adrianmoisey: > I may have messed with something then, sorry.

It's entirely possible that I messed up too 😆 
Given that their example doesn't have the hostname in the Gateway, I'm assuming I didn't... but I think it's worth double checking, and also makes me think that getting https://github.com/kubernetes-sigs/ingress2gateway/issues/234 in place is going to be important

mikemorris: HTTPRoutes specifying hostnames should not require those hostnames to be specified on the Gateway listener as well (omitting listener hostnames is should be acceptable).

It's a secondary filter though, so if hostnames _are_ specified on the listener to which a route is attached, only traffic which passes the listener filter will reach the route.

I'm not _quite_ clear on if omitting hostnames on a Gateway listener is sufficient for this purpose, or if we should be building a single listener with the union of all hostnames specified in aliases (and separate HTTPRoutes for each hostname)? Would it be permissible for additional manually-created HTTPRoutes later to match other hostnames, or is this functionality in ingress-nginx intended to constrain which hostnames can be matched?

adrianmoisey: > is this functionality in ingress-nginx intended to constrain which hostnames can be matched

I'm not too sure if I'd use the word "constrain", may be "route" is better.
My understanding is that the normal flow is that traffic ends up in the default backend, unless it matches a hostname or server-alias.
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-alias

adrianmoisey: I found this guide: https://gateway-api.sigs.k8s.io/guides/http-routing/
It seems to imply that we only need the hostnames on the HTTPRoute

LiorLieberman: The important thing is to understand how ingress-nginx actually handles it. For gateway api, IIUC, **there is** a difference between specifying hostnames on the listener vs httproute, (also wrt to certs). 

Ingress2gateway does not have to, and will never be able to do provide feature-complete conversion, so it _may_ be fine just moving forward with this hostnames logic, but its important to call this caveat (both in docs and in a notification to the user)

I am not sure the difference it would make if we compare a ingress-nginx with server-alias config vs gateway-api, empty hostname in listener, and a list of hostnames in httproute. Routing-wise, it will do the same, but whether it would have security implications -- that I dont know.

rikatz: IMO, if we do have a way to set hostnames on HTTPRoute without needing to change Gateway listeners, I think this is close to what is expected from Ingress (with server-alias)

This means that each hostname in a ingress server-alias will become a .spec.hostname on an HTTPRoute.

mikemorris: That behavior/structure feels like it would make sense to me @rikatz

LiorLieberman: Sounds good. @adrianmoisey can you please add a note (both in docs and in a notification to the user)

docs will be enough to state it in `pkg/i2gw/providers/ingressnginx/README.md` where you already added the description for `nginx.ingress.kubernetes.io/server-alias`, just add there a note about the assumption we are making.

And a notification from the conversion (I saw you've already used notifications in this PR so no need for me to explain). I'd expect notification at the point where you take the server aliases and add them to HTTPRoute, to say something like "nginx.ingress.kubernetes.io/server-alias values have been added to httproute.spec.hostnames sharing one Gateway listener.  (you can probably come up with a better phrasing than what I wrote here)

adrianmoisey: > docs will be enough to state it in pkg/i2gw/providers/ingressnginx/README.md where you already added the description for nginx.ingress.kubernetes.io/server-alias, just add there a note about the assumption we are making.

Can someone help me wordsmith this? I don't really know Gateway API, so I'm unsure exactly what assumptions we're making here other than what I've already added here: https://github.com/kubernetes-sigs/ingress2gateway/pull/227/files#diff-50a468f56707e31ebd87bc0151a62a8cd1a6b7d5859e178cc9b792a697377327R14

> And a notification from the conversion (I saw you've already used notifications in this PR so no need for me to explain). I'd expect notification at the point where you take the server aliases and add them to HTTPRoute, to say something like "nginx.ingress.kubernetes.io/server-alias values have been added to httproute.spec.hostnames sharing one Gateway listener. (you can probably come up with a better phrasing than what I wrote here)

Added in 496f208

pkg/i2gw/providers/ingressnginx/notification.go
adrianmoisey: Fixed in 60d8f88

pkg/i2gw/providers/ingressnginx/README.md
LiorLieberman: Just something like

Note: ingress2gateway will create an empty **Gateway** Listener, and will add all the values in server-alias annotation to the generated `HTTPRoute.spec.hostnames`

adrianmoisey: I've added that in 06a44ea

pkg/i2gw/providers/ingressnginx/server_alias.go
```diff
@@ -0,0 +1,97 @@
...
+				// Add server aliases as additional hostnames to the HTTPRoute
+				for _, alias := range aliases {
+					hostname := gatewayv1.Hostname(alias)
+					// Check if this hostname already exists
+					exists := slices.Contains(routeContext.Spec.Hostnames, hostname)
+
+					if !exists {
+						routeContext.Spec.Hostnames = append(routeContext.Spec.Hostnames, hostname)
+						notify(
+							fmt.Sprintf("added server alias hostname \"%s\" to HTTPRoute \"%s\" for Gateway \"%s\"", alias, routeContext.Name, gatewayKey.Name),
```
LiorLieberman, line 85: what do you mean by saying "for Gateway"?

adrianmoisey: The output of a single Ingress looks like this:
```
kind: Gateway
metadata:
  annotations:
  creationTimestamp: null
  name: nginx
  namespace: default
spec:
  gatewayClassName: nginx
  listeners:
  - name: http
    port: 80
    protocol: HTTP
status: {}
---
kind: HTTPRoute
metadata:
  annotations:
  creationTimestamp: null
  name: httpbin-all-hosts
  namespace: default
spec:
  hostnames:
  - adrian.com
  - api.adrian.com
  - cdn.adrian.com
  parentRefs:
  - name: nginx
  rules:
  - backendRefs:
    - name: httpbin
      port: 80
    matches:
    - path:
        type: PathPrefix
        value: /
status:
  parents: []
```

The HTTP route has a parentRefs, which is a Gateway.
So this host is added to the HTTPRoute, which is linked to a Gateway. In this case that Gateway's name is 'nginx'

adrianmoisey:
I'm deprioritising my work on ingress2gateway, so I won't have capacity to work on this
Someone else is welcome to reuse this if they want

PR #230, opened 2025-07-29
Idea/concept: Add source ingress annotation to httproute

### LINKED ISSUE #229: HTTPRouteRules lose source ingress
**What happened**:

When multiple Ingress resources in the same namespace and class share the same host, they are merged into a single `HTTPRoute`. However, the source Ingress name is not preserved per `HTTPRouteRule`, which causes all annotations that are specific to an individual Ingress to be incorrectly applied across all rules in the merged HTTPRoute.

This is an issue for Nginx Ingress Controller annotations, where annotations are scoped to an individual Ingress resource.

Each `HTTPRouteRule` should retain the name (or a reference to) the Ingress that contributed it. This enables annotations—especially those interpreted by providers—to be scoped and applied correctly per original Ingress.

1. Create two Ingresses with the same `host`, namespace, and class.
2. Attach a provider-specific annotation (e.g., for header manipulation, security, timeouts, etc.) only to one of them.
3. Observe that the annotation’s effects are incorrectly applied to rules from both Ingresses in the merged `HTTPRoute`.

### Root Cause

The merge key used to group Ingresses is:

```go
fmt.Sprintf("%s/%s/%s", namespace, ingressClass, rule.Host)
````

This groups all Ingresses with the same host into one HTTPRoute. However, the code drops the source Ingress context:

```go
rg.rules = append(rg.rules, ingressRule{rule: rule}) // no Ingress name tracked
```

So there's no way to later associate a rule with its original Ingress.

### Why it matters

Provider-specific features (e.g., header filtering, authentication, rate limiting, etc.) are often configured using annotations on individual Ingresses. Without knowing which rule came from which Ingress, those features cannot be scoped or applied accurately.

### Possible Solutions

1. **Track source Ingress name per rule internally.**
   Surface it in metadata or annotations (e.g., `i2gw.ingressName`) in the generated HTTPRoute.

2. **Generate one HTTPRoute per Ingress.**
   Include Ingress name in the merge key to avoid combining Ingresses under the same host. Less desirable due to output changes.

3. **Introduce provider hooks or callbacks.**
   Allow providers to access original rule context during translation. More complex.

The preferred solution is (1) — retain the source Ingress per rule without changing the output structure or merging behavior.

I am recreating this issue because I accidentally closed the previous one and cannot reopen

sarthyparty:
**What type of PR is this?**
/kind bug

This PR adds an annotation to each httproute, allowing Ingress2Gateway to track which Ingress resource contributed each HTTPRouteRule. This is necessary because when multiple Ingresses are merged into a single HTTPRoute (sharing host, namespace, and class), the origin of individual rules is currently lost.

Tracking the source Ingress enables providers to correctly scope annotations and features that are defined per-Ingress, rather than applying them incorrectly to all rules in a merged HTTPRoute.

This change does not affect merging behavior or output structure, but adds metadata that downstream provider implementations can optionally use.

Fixes #229 

```release-note
None
````

### DIFF
--- a/pkg/i2gw/providers/common/converter.go
+++ b/pkg/i2gw/providers/common/converter.go
@@ -123,7 +123,8 @@ type ingressRuleGroup struct {
 }
 
 type ingressRule struct {
-	rule networkingv1.IngressRule
+	rule          networkingv1.IngressRule
+	sourceIngress types.NamespacedName
 }
 
 type ingressDefaultBackend struct {
@@ -170,7 +171,7 @@ func (a *ingressAggregator) addIngressRule(namespace, name, ingressClass string,
 	if len(iSpec.TLS) > 0 {
 		rg.tls = append(rg.tls, iSpec.TLS...)
 	}
-	rg.rules = append(rg.rules, ingressRule{rule: rule})
+	rg.rules = append(rg.rules, ingressRule{rule: rule, sourceIngress: types.NamespacedName{Namespace: namespace, Name: name}})
 }
 
 func (a *ingressAggregator) toHTTPRoutesAndGateways(options i2gw.ProviderImplementationSpecificOptions) ([]gatewayv1.HTTPRoute, []gatewayv1.Gateway, field.ErrorList) {
--- a/pkg/i2gw/providers/common/converter.go
+++ b/pkg/i2gw/providers/common/converter.go
@@ -341,11 +341,6 @@ func (rg *ingressRuleGroup) toHTTPRoute(servicePorts map[types.NamespacedName]ma
 		errors = append(errors, errs...)
 		hrRule.BackendRefs = backendRefs
 
-		// Add source ingress information for this specific rule
-		if hrRule.Filters == nil {
-			hrRule.Filters = []gatewayv1.HTTPRouteFilter{}
-		}
-
 		for _, p := range paths {
 			ruleSourceIngresses[p.sourceIngress] = append(ruleSourceIngresses[p.sourceIngress], len(httpRoute.Spec.Rules))
 		}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/i2gw/providers/common/converter.go [hunk 1]
```diff
@@ -337,9 +341,35 @@ func (rg *ingressRuleGroup) toHTTPRoute(servicePorts map[types.NamespacedName]ma
...
 		httpRoute.Spec.Rules = append(httpRoute.Spec.Rules, hrRule)
 	}
 
+	// Initialize annotations map if needed
+	if httpRoute.Annotations == nil {
+		httpRoute.Annotations = make(map[string]string)
+	}
+
+	// Build annotation mapping source ingress to rule indices
+	var annotationParts []string
```
sjberman: When initializing a slice, it's best to set its size if you know it beforehand. Saves on having to reallocate whenever it grows for each append.

`annotationParts := make([]string, 0, len(ruleSourceIngresses))`

pkg/i2gw/providers/common/converter.go
```diff
@@ -337,9 +341,35 @@ func (rg *ingressRuleGroup) toHTTPRoute(servicePorts map[types.NamespacedName]ma
...
+	for sourceIngress, ruleIndices := range ruleSourceIngresses {
+		ingressKey := sourceIngress.Namespace + "/" + sourceIngress.Name
+		var indicesStr []string
```
sjberman:
```suggestion
		indicesStr := make([]string, 0, len(ruleIndices))
```

pkg/i2gw/providers/common/converter.go
```diff
@@ -337,9 +341,35 @@ func (rg *ingressRuleGroup) toHTTPRoute(servicePorts map[types.NamespacedName]ma
 		errors = append(errors, errs...)
 		hrRule.BackendRefs = backendRefs
 
+		// Add source ingress information for this specific rule
+		if hrRule.Filters == nil {
+			hrRule.Filters = []gatewayv1.HTTPRouteFilter{}
+		}
```
sjberman: Why was this needed?

pkg/i2gw/providers/common/converter_test.go
```diff
@@ -100,7 +100,13 @@ func Test_ToIR(t *testing.T) {
 				HTTPRoutes: map[types.NamespacedName]intermediate.HTTPRouteContext{
 					{Namespace: "test", Name: "simple-example-com"}: {
 						HTTPRoute: gatewayv1.HTTPRoute{
-							ObjectMeta: metav1.ObjectMeta{Name: "simple-example-com", Namespace: "test"},
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      "simple-example-com",
+								Namespace: "test",
+								Annotations: map[string]string{
+									"ingress2gateway.io/source-ingress-rules": "test/simple:0",
```
sjberman, line 107: Why does the name get cut off? Is that happening in a different part of the code?`simple-example-com` -> `simple`

pkg/i2gw/providers/common/converter.go
sarthyparty: I think copilot added it and I missed it

pkg/i2gw/providers/common/converter_test.go
sarthyparty: simple is the name of the ingress, while example.com is the host name, and then the httproute name became simple-example-com, but the original source ingress name is still simple

sjberman: Oh gotcha, I misread.

PR #233, opened 2025-08-18
Do not print the table on stdout

rikatz:
**What type of PR is this?**
/kind bug

Stop printing the info table as part of the manifest. When we print to stdout the manifest becomes invalid and doing things like "ingress2gateway | kubectl apply -f -" won't work

Fixes #

```release-note
Print notification table on stderr
```


robscott:
Thanks @rikatz!

/approve

Issue #234, opened 2025-08-18
Implement e2e test for ingress2gateway


rikatz:
**What would you like to be added**:
Implement e2e test for ingress2gateway

We don't have as of today any sort of e2e that can get some Ingress resource and turn into a GatewayAPI resource, to later apply on a cluster and test its behavior.

This can cause situations where an implementation is not correct, or doesn't reflect the real user experience

As a proposal, we can implement some e2e tests that:
* Bootstrap a KinD cluster
* Install some GatewayAPI controller - Suggestion is to use envoy gateway
* Has a set of tests for the desired providers that represent an ingress resource. As an example, we can have yaml manifests that represents Ingress NGINX with its features (server-alias, canary, etc). 
* Run ingress2gateway on these files, convert to GatewayAPI manifests and apply to cluster
* Run a simple "curl" to guarantee that the expected behavior is working

As this is a cli, we want to reproduce the user behavior, so my suggestion is to use [bats](https://github.com/bats-core/bats-core)  for these tests.

carmal891:
@rikatz Hi,
I’ve gone through the initial e2e structure you’ve set up. I noticed that the mentioned KinD cluster bootstrap and the pre_check() function still need some work. I'd be happy to collaborate and help complete these pending parts. My plan is to use your [PR](https://github.com/kubernetes-sigs/ingress2gateway/pull/240) as a base and contribute additional changes in a separate commit.

Please let me know if this is fine by you and if there’s anything else you’d like me to focus on regarding this ticket.

rikatz:
@carmal891 totally fine, go for it. I am not sure we do have an agreement on my structure, but I think that using bats reflects the best user experience for e2e tests (we are fully reproducing what users do on kubectl, not what a program would do using go libraries)

One note: I wanted to discuss with @LiorLieberman first (and you are welcome to contribute!) about https://github.com/kubernetes-sigs/ingress2gateway/pull/244 

I think part of the experience that needs to be improved on i2g is the fact that we assume that a Gateway will need to be created, and that we want to attach routes to it, while we may have already a Gateway or other parent (like a ListenerSet) and just want to create routes that attaches to it.

But for now, as we don't have any e2e test I would love to see something, so feel free to contribute over my PR, opening yours :)

carmal891:
@rikatz Apologies for the delay from my side. I got caught up with other project priorities. I ran the initial e2e structure that you set up using Bats but I encountered an issue

[...]

PR #240, opened 2025-08-29
Initial e2e structure

rikatz:
**What type of PR is this?**
/kind test

This is the initial implementation of e2e test. No Github Actions/CI is set yet, nor provider specifics.

Part of #234 
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@
 # Enable Go modules.
 export GO111MODULE=on
 
+# INGRESS2GATEWAY defines the ingress2gateway location
+# It will be used by some makefile targets to avoid rebuilding it
+INGRESS2GATEWAY ?= $(shell pwd)/ingress2gateway
+
 # I2GWPKG is the package path for i2gw.  This allows us to propogate the git info
 # to the binary via LDFLAGS.
 I2GWPKG := $(shell go list .)/pkg/i2gw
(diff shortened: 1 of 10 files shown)


rikatz:
/hold

couple of things not working:
* ingress2gateway is failing to use local files. While we may want to make simple tests against a namespace AND local files, I would say the majority of tests could be just an assessment of local files and this way we need to make the input flag work
* The provider flag says it is optional, and yet it is required when running the program

hack/verify-e2e.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+
+# Copyright 2014 The Kubernetes Authors.
```
adrianmoisey, line 3: :D

rikatz: dammit copy/pasta

Makefile
```diff
@@ -57,10 +61,21 @@ test: vet;$(info $(M)...Begin to run tests.)  @ ## Run tests.
...
 
 # Run static analysis.
 .PHONY: verify
 verify:
 	hack/verify-all.sh -v
+
+# Run e2e tests
+.PHONY: e2e
+e2e: $(INGRESS2GATEWAY)
+	I2G=$(INGRESS2GATEWAY) hack/verify-e2e.sh
```
LiorLieberman, line 81: nit: Can we do I2GW?

PR #241, opened 2025-09-01
Do not require a namespace when input-file is used

rikatz:
**What type of PR is this?**
/kind cleanup

This PR changes the behavior of i2g when using an input-file. In this case, as we are not querying a live Kubernetes cluster it may be expected that the namespace selector is not used unless it is explicitly required.

This way, when using --input-file without a namespace, all the namespaces are selected

```release-note
Do not require namespace when using input-file flag, default to all-namespaces
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -278,7 +278,6 @@ func (pr *PrintRunner) initializeResourcePrinter() error {
 // 2. If an input-file is used and no namespace is added then all namespaces are selected.
 // 3. If namespace is specified, it filters resources based on that namespace.
 // 4. If no namespace is specified and reading from the cluster, it attempts to get the namespace from the cluster; if unsuccessful, initialization fails.
-// 5. If no namespace is specified and reading from a file, it attempts to get the namespace from the cluster; if unsuccessful, it reads all resources.
 func (pr *PrintRunner) initializeNamespaceFilter() error {
 	// When we should use all namespaces, empty string is used as the filter.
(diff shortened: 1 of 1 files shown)


cmd/print.go
```diff
@@ -275,12 +275,13 @@ func (pr *PrintRunner) initializeResourcePrinter() error {
 
 // initializeNamespaceFilter initializes the correct namespace filter for resource processing with these scenarios:
 // 1. If the --all-namespaces flag is used, it processes all resources, regardless of whether they are from the cluster or file.
-// 2. If namespace is specified, it filters resources based on that namespace.
-// 3. If no namespace is specified and reading from the cluster, it attempts to get the namespace from the cluster; if unsuccessful, initialization fails.
-// 4. If no namespace is specified and reading from a file, it attempts to get the namespace from the cluster; if unsuccessful, it reads all resources.
+// 2. If an input-file is used and no namespace is added then all namespaces are selected.
+// 3. If namespace is specified, it filters resources based on that namespace.
+// 4. If no namespace is specified and reading from the cluster, it attempts to get the namespace from the cluster; if unsuccessful, initialization fails.
+// 5. If no namespace is specified and reading from a file, it attempts to get the namespace from the cluster; if unsuccessful, it reads all resources.
```
sarthyparty: 5 should be updated or removed?

rikatz: fixed

PR #242, opened 2025-09-01
Support kyaml output

rikatz:
**What type of PR is this?**
/kind feature

This PR adds support for the new supported kyaml format. kyaml can be used for more explicit outputs, where disambiguation is required or during format conversions.

```release-note
Add support for kyaml printer
```


rikatz:
/hold

To be discussed

.golangci.yml
```diff
@@ -2,7 +2,7 @@ run:
   timeout: 10m
   issues-exit-code: 1
   tests: true
-  skip-dirs-use-default: true
+  exclude-dirs-use-default: true
```
rikatz, line 5: older parameter was deprecated

.golangci.yml
```diff
@@ -32,6 +31,10 @@ linters-settings:
     shadow: true
   misspell:
     locale: US
+  # (rkatz) - Added as we have a couple of APIs conversion (eg Istio, Kong) that convert uint->int and there's no solution
```
rikatz, line 34: this check was bumped with golangci-lint 1.64 (1.59 wasn't working properly) and is raising problems on providers (istio, kong) that use their own API.

rikatz:
> Set where? in the terminal? Is it really necessary?

`KUBECTL_KYAML=true ./ingress2gateway print -o yaml`

It is, basically because of this:

https://github.com/kubernetes-sigs/ingress2gateway/blob/dd0bf1c9f741729261e65874a1c2de5e907bad93/cmd/print.go#L309

That calls this:

https://github.com/kubernetes/cli-runtime/blob/master/pkg/genericclioptions/json_yaml_flags.go#L36

so if you don't set the envvar the output will never be accepted :)

LiorLieberman:
then what happens if I provide --output=kyaml but does not have this envar set. 

I dont think I fail validation with your changes. The block you references is only used for the flag description [here](https://github.com/kubernetes-sigs/ingress2gateway/blob/dd0bf1c9f741729261e65874a1c2de5e907bad93/cmd/print.go#L327)

rikatz:
true :man_facepalming: 

I gotta go for the day, but can test tomorrow, probably we can just drop the printer format if we do a further validation and it is just being used to say "you can use these formats" and rely on the fact the we control the printers anyway

rikatz:
@LiorLieberman good catch, fixed here, as we kind of don't care much about the flag printers (and we do our own printer instantiation) I have removed the printerflags from the file as well.

LiorLieberman:
~Thanks, I may be missing something though~

~>then what happens if I provide --output=kyaml but does not have this envar set.~

~This will still pass validation, right? but it wont work.~

nvm, I see that now we are not calling the func that checks the existence of the env var. 

Thanks! 
/lgtm
/approve

rikatz:
right, and kyaml is a client side implementation, not a server side, so as we are bumping the libraries everything should work fine and we will have @thockin happy that people are now using kyaml :P

rikatz:
@LiorLieberman btw as you have edit your comment, you need to do a lgtm/approve again on a separate comment :)

thockin:
Just be aware that KYAML could still be deemed A Bad Idea and removed. :)

rikatz:
then once we bump the library again, it will fail to compile and we need to remove 4 lines.

Also i2gw is pre GA, so people should be aware before using it and bumping its version :)

Issue #243, opened 2025-09-01
Improve usage with different personas (use ListenerSet or Gateway when required)


rikatz:
**What would you like to be added**:
 
* As a developer, I want to use ListenerSet instead of `Gateway` and be able to attach my ListenerSet to an existing Gateway
* As a developer, I may want to reuse an existing `Gateway`, without explicitly requiring the creation of a new Gateway

**Why this is needed**:

The conversion of resources on Ingress2Gateway today assumes the need of creation of a new Gateway Resource. This may not be always true. As an example, imagine the following scenario:
* The admin has already created a Gateway for me
* I just want to point to that Gateway (scenario 1 - reuse an existing Gateway)
* Then, I realize that I have my own certificate, and that I want to add it to the Listener. In that case, I need to create a ListenerSet and attach my HTTPRoute to it (scenario 2)
* I can also require that my new HTTPRoutes are attached to an existing ListenerSet and not to a Gateway (scenario 3)

This way, we may need new flags that:
* Allow a user to override the `parentRef` of the route to an existing resource
* Allow a user to disable the Gateway creation. In case the user set the `parentRef` can we skip the gateway creation?
* Allow a user to specify that "instead of creating a Gateway, create a ListenerSet and attach to a Gateway" 
* Allow a user to specify two `parentRefs`, one for Gateway and one for `ListenerSet` (pending the discussion of ListenerSet GEP)

PR #244, opened 2025-09-01
[WIP] Add initial support for parentRefs

rikatz:
**What type of PR is this?**
/kind feature

This change adds a feature to support adding parentRefs on Route resources. It is a new slice flag that can be used to establish parentRefs that should be added to converted resources.

This is still a WIP

Part of #243 

```release-note
Add support for parentRefs on route conversion
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"slices"
+	"strconv"
 	"strings"
 
 	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
@@ -30,6 +31,8 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/utils/ptr"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	// Call init function for the providers
 	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/apisix"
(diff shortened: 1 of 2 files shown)


rikatz:
For now this is a WIP, I have added the flags only.

If we agree with the syntax, I will apply the parentRefs on the *Route generation. 

Later we need to discuss about "not creating a Gateway" if a parentRef is specified (we can add a flag to disable it, so it wont be a breaking change)

cmd/print.go
```diff
@@ -73,6 +76,19 @@ type PrintRunner struct {
 
 	// Provider specific flags --<provider>-<flag>.
 	providerSpecificFlags map[string]*string
+
+	// parentRefs indicates that the created *Route should contain a specific set
+	// of parentRefs. Its format should be as: --parentRefs=ns/name:sectionname:port=group/kind
```
LiorLieberman, line 81: not sure I understand the format.what is the double "=" for?

LiorLieberman: like why we use both ":" and "=" as dividers? I understand the code and what it does but probably missing something on why we choose it.

also this would highly benefit from being a type by itself, I think.

rikatz: It would, do we support types on flags? 

So, from the structure of the flag itself, I wanted to follow something that was "as easy as set just the name of the ref" (--parentRef=xpto) but as complex as "I need all of the other fields".

To be really honest, I don't like flags that start to have its own complexity to cover all of the edge cases (eg.: do we really want to add port? sectionName? etc) but also, how can we make it easier once we support ListenerSet to provide the proper conversion? I am open to ideas here

One last comment: this format of the flag was highly inspired on "kubectl create ingress" command, I just did inverted some parameters to make it easier to have something like `--parentRefs=Ricardo=XListenerSet`

PR #246, opened 2025-09-03
Read NGINX CRDs and build gateway

sarthyparty:
**What type of PR is this?**
/kind feature

Adds NGINX CRD conversion infra

```release-note

```

### DIFF
--- a/pkg/i2gw/providers/nginx/nginx.go
+++ b/pkg/i2gw/providers/nginx/nginx.go
@@ -36,7 +36,7 @@ func init() {
 
 	i2gw.RegisterProviderSpecificFlag(Name, i2gw.ProviderSpecificFlag{
 		Name:        GlobalConfigurationFlag,
-		Description: "Name of NIC GlobalConfiguration resource.",
+		Description: "Namespace and Name of NIC GlobalConfiguration resource.",
 	})
 }
 


go.mod
bjee19, line 1: why are there so many go.mod and go.sum updates? Especially the go version, i feel like this pr isn't the place for them.

pkg/i2gw/providers/nginx/annotations/path_rewrite.go
```diff
@@ -87,34 +80,25 @@ func RewriteTargetFeature(ingresses []networkingv1.Ingress, _ map[types.Namespac
...
 		}
-
 		serviceName := strings.TrimSpace(part[serviceIdx+1 : rewriteIdx])
 		rewritePath := strings.TrimSpace(part[rewriteIdx+9:])
-
 		if serviceName != "" && rewritePath != "" {
 			rules[serviceName] = rewritePath
 		}
 	}
-
```
bjee19, line 118: Why are all these spaces being removed?

pkg/i2gw/providers/nginx/annotations/header_manipulation.go
```diff
@@ -153,12 +119,7 @@ func parseSetHeaders(setHeaders string) map[string]string {
...
 		}
-		// Note: Headers without explicit values (format "$Variable-Name") are skipped
-		// as Gateway API cannot use NGINX variables like $http_* and headers need values
 	}
 
 	return headers
 }
-
-// Note: The patchHTTPRouteHeaderMatching function has been removed as it was incomplete.
-// Header matching should be implemented separately if needed for specific NGINX features.
```
bjee19, line 164: Are these comments supposed to be deleted?

pkg/i2gw/providers/nginx/annotations/header_manipulation.go
```diff
@@ -143,7 +110,6 @@ func parseSetHeaders(setHeaders string) map[string]string {
 
 	for _, part := range parts {
 		if strings.Contains(part, ":") {
-			// Format: "Header-Name: value"
```
bjee19, line 146: Same with this comment, is it supposed to be deleted?

pkg/i2gw/providers/nginx/crd_reader.go
```diff
@@ -0,0 +1,83 @@
...
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(gvk)
+
+	if err := c.List(ctx, list); err != nil {
+		return nil, fmt.Errorf("failed to list %s: %w", gvk.GroupKind().String(), err)
+	}
+
+	var items []T
+	for _, u := range list.Items {
+		if namespace != "" && u.GetNamespace() != namespace {
```
bjee19, line 44: I feel like this check `namespace != ""` is unnecessary, unless you expect it to be "", but the default should be "default". OR this check should be at the top of the function and should probably throw an error.

pkg/i2gw/providers/nginx/crd_reader.go
```diff
@@ -0,0 +1,83 @@
...
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), obj); err != nil {
+			return nil, fmt.Errorf("failed to parse %s object %s/%s: %w", gvk.Kind, u.GetNamespace(), u.GetName(), err)
+		}
+		items = append(items, *obj)
+	}
+	return items, nil
+}
+
+// genericReadFromFile reads CRDs of type T from a YAML file using the given GVK.
+func genericReadFromFile[T any](filename string, namespace string, gvk schema.GroupVersionKind, newObj func() *T) ([]T, error) {
```
bjee19, line 57: What does namespace do here? Doesn't make sense when reading from a file.

pkg/i2gw/providers/nginx/crd_reader.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		if !u.GroupVersionKind().Empty() && u.GroupVersionKind() == gvk {
+			obj := newObj()
+			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), obj); err != nil {
+				return nil, fmt.Errorf("failed to parse %s object %s/%s: %w", gvk.Kind, u.GetNamespace(), u.GetName(), err)
+			}
+			items = append(items, *obj)
+		}
+	}
+	return items, nil
```
bjee19, line 82:
```suggestion
	for _, u := range objs {
		if namespace != "" && u.GetNamespace() != namespace {
			continue
		}
		
		if !u.GroupVersionKind().Empty() && u.GroupVersionKind() == gvk {
			obj := newObj()
			if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), obj); err != nil {
				return nil, fmt.Errorf("failed to parse %s object %s/%s: %w", gvk.Kind, u.GetNamespace(), u.GetName(), err)
			}
			
			items = append(items, *obj)
		}
	}
	
	return items, nil
```

pkg/i2gw/providers/nginx/crd_reader.go
```diff
@@ -0,0 +1,83 @@
...
+			continue
+		}
+		obj := newObj()
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), obj); err != nil {
+			return nil, fmt.Errorf("failed to parse %s object %s/%s: %w", gvk.Kind, u.GetNamespace(), u.GetName(), err)
+		}
+		items = append(items, *obj)
+	}
+	return items, nil
+}
```
bjee19, line 54: golang spacing conventions

```suggestion
	for _, u := range list.Items {
		if namespace != "" && u.GetNamespace() != namespace {
			continue
		}
		
		obj := newObj()
		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.UnstructuredContent(), obj); err != nil {
			return nil, fmt.Errorf("failed to parse %s object %s/%s: %w", gvk.Kind, u.GetNamespace(), u.GetName(), err)
		}
		
		items = append(items, *obj)
	}
	
	return items, nil
}
```

pkg/i2gw/providers/nginx/resource_reader.go
```diff
@@ -41,6 +55,31 @@ func newResourceReader(conf *i2gw.ProviderConf) *resourceReader {
...
+func newVirtualServer() *nginxv1.VirtualServer { return &nginxv1.VirtualServer{} }
+func newVirtualServerRoute() *nginxv1.VirtualServerRoute {
+	return &nginxv1.VirtualServerRoute{}
+}
+func newTransportServer() *nginxv1.TransportServer {
+	return &nginxv1.TransportServer{}
+}
+func newGlobalConfiguration() *nginxv1.GlobalConfiguration {
+	return &nginxv1.GlobalConfiguration{}
+}
```
bjee19, line 68:
```suggestion
func newVirtualServer() *nginxv1.VirtualServer { return &nginxv1.VirtualServer{} }

func newVirtualServerRoute() *nginxv1.VirtualServerRoute {
	return &nginxv1.VirtualServerRoute{}
}

func newTransportServer() *nginxv1.TransportServer {
	return &nginxv1.TransportServer{}
}

func newGlobalConfiguration() *nginxv1.GlobalConfiguration {
	return &nginxv1.GlobalConfiguration{}
}
```

pkg/i2gw/providers/nginx/resource_reader.go
```diff
@@ -41,6 +55,31 @@ func newResourceReader(conf *i2gw.ProviderConf) *resourceReader {
...
+
+// parseNamespacedName parses a string in the format "namespace/name" or just "name"
+// Returns namespace and name. If no namespace is specified, uses the default namespace.
+func parseNamespacedName(namespacedName, defaultNamespace string) (namespace, name string) {
```
bjee19, line 72: what if namespace or name contain "/", or is that not allowed?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+
+// CreateNamespaceGateway creates a single Gateway for all VirtualServers and TransportServers in the namespace
+func (f *NamespaceGatewayFactory) CreateNamespaceGateway() (map[types.NamespacedName]intermediate.GatewayContext, map[string][]GatewayListenerKey) {
+	gatewayName := DefaultGatewayName
+	gatewayKey := types.NamespacedName{
+		Namespace: f.namespace,
+		Name:      gatewayName,
+	}
+
+	gateways := make(map[types.NamespacedName]intermediate.GatewayContext)
```
bjee19, line 81: why is this called gateways when the function only creates a single gateway?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+		namespace:        namespace,
+		virtualServers:   virtualServers,
+		transportServers: transportServers,
+		notificationList: notifs,
+		listenerMap:      listenerMap,
+	}
+}
+
+// CreateNamespaceGateway creates a single Gateway for all VirtualServers and TransportServers in the namespace
+func (f *NamespaceGatewayFactory) CreateNamespaceGateway() (map[types.NamespacedName]intermediate.GatewayContext, map[string][]GatewayListenerKey) {
```
bjee19, line 74: Why does this function return the virtualServerMap as well? Can you include that in the comment above? 

Also since this function only returns a single gateway, i'm not sure why it returns a map.

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+					Protocol: gatewayv1.HTTPSProtocolType,
+					Hostname: (*gatewayv1.Hostname)(hostPtr),
+					TLS: &gatewayv1.GatewayTLSConfig{
+						Mode: Ptr(gatewayv1.TLSModeTerminate),
+						CertificateRefs: []gatewayv1.SecretObjectReference{
+							{
+								Group: Ptr(gatewayv1.Group(common.CoreGroup)),
+								Kind:  Ptr(gatewayv1.Kind(common.SecretKind)),
+								Name:  gatewayv1.ObjectName(secret),
+							},
```
bjee19, line 183: if the secret isn't specified should we still include this part?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+		transportServerMap[ts.Name] = append(transportServerMap[ts.Name], GatewayListenerKey{
+			gatewayName:  gatewayName,
+			listenerName: listenerName,
+		})
+	}
+
+	// Merge virtualServerMap and transportServerMap
+	for tsName, listeners := range transportServerMap {
+		virtualServerMap[tsName] = listeners
+	}
```
bjee19, line 264: I think its confusing re-using one of the maps and keeping it the same name. Could you instead just create a new map with a new name?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+	// Convert map to slice and sort for deterministic order
+	var listeners []gatewayv1.Listener
+	for _, listener := range uniqueListeners {
+		listeners = append(listeners, listener)
+	}
+
+	// Sort listeners by name for deterministic ordering
+	sort.Slice(listeners, func(i, j int) bool {
+		return string(listeners[i].Name) < string(listeners[j].Name)
+	})
```
bjee19, line 286: In these 3 sorts. Doesn't sort 1 and 3 do the same thing?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+
+	return listeners, virtualServerMap
+}
+
+// getListenerPorts determines which ports/protocols a VirtualServer needs
+func (f *NamespaceGatewayFactory) getListenerPorts(vs nginxv1.VirtualServer) (httpPort, httpsPort int) {
+	// Check if VirtualServer specifies custom listeners
+	if vs.Spec.Listener != nil {
+		// Use custom listeners from GlobalConfiguration
```
bjee19, line 295: Where does GlobalConfiguration come into play here?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+	} else {
+		// Use default ports
+		httpPort = namespaceGatewayHTTPPort
+		if vs.Spec.TLS != nil {
+			httpsPort = namespaceGatewayHTTPSPort
+		}
+	}
+	return httpPort, httpsPort
+}
```
bjee19, line 318:
```suggestion
func (f *NamespaceGatewayFactory) getListenerPorts(vs nginxv1.VirtualServer) (httpPort, httpsPort int) {
	// Check if VirtualServer specifies custom listeners
	if vs.Spec.Listener != nil {
		// Use custom listeners from GlobalConfiguration
		if vs.Spec.Listener.HTTP != "" {
			if listener, found := f.listenerMap[vs.Spec.Listener.HTTP]; found {
				if listener.Protocol == gatewayv1.HTTPProtocolType {
					httpPort = int(listener.Port)
				}
			}
		}
		
		if vs.Spec.Listener.HTTPS != "" {
			if listener, found := f.listenerMap[vs.Spec.Listener.HTTPS]; found {
				if listener.Protocol == gatewayv1.HTTPSProtocolType {
					httpsPort = int(listener.Port)
				}
			}
		}
	} else {
		// Use default ports
		httpPort = namespaceGatewayHTTPPort
		if vs.Spec.TLS != nil {
			httpsPort = namespaceGatewayHTTPSPort
		}
	}
	
	return httpPort, httpsPort
}
```

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+
+// getTransportServerPort determines the port for a TransportServer from GlobalConfiguration or defaults
+func (f *NamespaceGatewayFactory) getTransportServerPort(ts nginxv1.TransportServer) *int {
+	listenerName := ts.Spec.Listener.Name
+	if listener, exists := f.listenerMap[listenerName]; exists {
+		return Ptr(int(listener.Port))
+	}
+	return nil
+}
```
bjee19, line 327:
```suggestion
func (f *NamespaceGatewayFactory) getTransportServerPort(ts nginxv1.TransportServer) *int {
	listenerName := ts.Spec.Listener.Name
	if listener, exists := f.listenerMap[listenerName]; exists {
		return Ptr(int(listener.Port))
	}
	
	return nil
}
```

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+		// Use default ports
+		httpPort = namespaceGatewayHTTPPort
+		if vs.Spec.TLS != nil {
+			httpsPort = namespaceGatewayHTTPSPort
+		}
+	}
+	return httpPort, httpsPort
+}
+
+// getTransportServerPort determines the port for a TransportServer from GlobalConfiguration or defaults
```
bjee19, line 320: Also confused as to what this GlobalConfiguration is

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+	secret = strings.ReplaceAll(secret, "_", "-")
+	secret = strings.Trim(secret, "-")
+	if secret == "" {
+		return "nosecret"
+	}
+	if len(secret) > 20 { // Keep it shorter for section names
+		secret = secret[:20]
+	}
+	return secret
+}
```
bjee19, line 391:
```suggestion
// sanitizeHostname replaces special characters for use in sectionName
func sanitizeHostname(host string) string {
	if host == "" {
		return "catchall"
	}
	
	host = strings.ToLower(host)
	host = strings.ReplaceAll(host, ".", "-")
	host = strings.ReplaceAll(host, "*", "wildcard")
	host = strings.ReplaceAll(host, ":", "-")
	host = strings.ReplaceAll(host, "_", "-")
	host = strings.Trim(host, "-")
	if host == "" {
		return "catchall"
	}
	
	if len(host) > 30 { // Keep it shorter for section names
		host = host[:30]
	}
	
	return host
}

// sanitizeSecret replaces special characters in secret names for use in sectionName
func sanitizeSecret(secret string) string {
	if secret == "" {
		return "nosecret"
	}
	
	secret = strings.ToLower(secret)
	secret = strings.ReplaceAll(secret, ".", "-")
	secret = strings.ReplaceAll(secret, "_", "-")
	secret = strings.Trim(secret, "-")
	if secret == "" {
		return "nosecret"
	}
	
	if len(secret) > 20 { // Keep it shorter for section names
		secret = secret[:20]
	}
	
	return secret
}
```

go.mod
sarthyparty: I'll fix the go version update. The only one that should be directly changed was the kubernetes ingress controller import and then indirects.

pkg/i2gw/providers/nginx/annotations/header_manipulation.go
sarthyparty: yeah I felt like they weren't that needed

pkg/i2gw/providers/nginx/annotations/header_manipulation.go
sarthyparty: It felt redundant

pkg/i2gw/providers/nginx/annotations/path_rewrite.go
sarthyparty: Probably AI cleaning up my code, since there were too many comments

pkg/i2gw/providers/nginx/crd_reader.go
sarthyparty: Technically a user can filter by namespace when reading from a file. by default it doesn't filter by namespace

pkg/i2gw/providers/nginx/crd_reader.go
sarthyparty: yea fair enough

sarthyparty: I copied it over from reading from file, so for cluster reading it should be default

pkg/i2gw/providers/nginx/resource_reader.go
sarthyparty: its not allowed

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: yeah good catch, it used to create multiple and I didn't refactor the variable names properly

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: fixed the second part, Im not sure what you're asking in the first part

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: I think NIC requires it for HTTPS. I don't think I should do validation that NIC already does

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: sort 1 sorts the parent refs that a route would have and a sort 3 sorts the listeners on the gateway.

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: The listener map ports are populated via global configuration

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: Global configuration is a NIC custom CRD that maps listener names to a port. So a user can specify a customer listener like my-listener and then I go to GLobalCOnfiguration and need to read the port that my-listener has

pkg/i2gw/providers/nginx/common/types.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package common
+
+const (
+	CoreGroup   = ""
+	SecretKind  = "Secret"
+	ServiceKind = "Service"
+)
```
shaun-nx, line 23: Can this at a higher level path and be specific to NGINX?
It feels like all providers would need something like this.

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/intermediate"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/nginx/common"
+	nginxv1 "github.com/nginx/kubernetes-ingress/pkg/apis/configuration/v1"
+)
+
+const (
+	// DefaultGatewayName is the default name for gateways created from VirtualServers
```
shaun-nx:
```suggestion
	// DefaultGatewayName is the default name for gateways created from VirtualServers and TransportServers
```
Should this mention TransportServers as well or are Gateways only relevant to VirtualServers?

The `CreateNamespaceGateway` function mentions that it `creates a single Gateway for all VirtualServers and TransportServers in the namespace`.

pkg/i2gw/providers/nginx/crds/gateway_builder.go
shaun-nx: @sarthyparty what's the advantage of merging the maps? Can we just return both?

shaun-nx: Also, what happens if someone has a TransportServer and a VirtualServer with the same name? Will this override the VirtualServer with the same name?

pkg/i2gw/providers/nginx/nginx.go
```diff
@@ -25,10 +25,19 @@ import (
...
+
+	GlobalConfigurationFlag = "global-configuration"
+)
 
 func init() {
 	i2gw.ProviderConstructorByName[Name] = NewProvider
+
+	i2gw.RegisterProviderSpecificFlag(Name, i2gw.ProviderSpecificFlag{
+		Name:        GlobalConfigurationFlag,
+		Description: "Name of NIC GlobalConfiguration resource.",
```
shaun-nx:
```suggestion
		Description: "Namespace and Name of NIC GlobalConfiguration resource.",
```
When defining a GlobalConfiguration resource, you need to explicitly set the namespace/name of the resource.
https://docs.nginx.com/nginx-ingress-controller/configuration/global-configuration/command-line-arguments/#-global-configuration-string

pkg/i2gw/providers/nginx/common/filters.go
```diff
@@ -0,0 +1,77 @@
...
+	var headers []gatewayv1.HTTPHeader
+	for name, value := range headersToSet {
+		if value != "" && !strings.Contains(value, "$") {
+			headers = append(headers, gatewayv1.HTTPHeader{
+				Name:  gatewayv1.HTTPHeaderName(name),
+				Value: value,
+			})
+		}
+	}
+	if len(headers) == 0 {
```
salonichf5, line 52: question: why aren't we supporting headers to add and remove here? Not supported by NIC via annotations?

pkg/i2gw/providers/nginx/common/types.go
salonichf5: yeah they probably do but keeping it specific to NGINX is a better choice for reducing overhead for Sarthak's work for now

pkg/i2gw/providers/nginx/common/filters.go
sarthyparty: NIC proxy-set-headers modifies the request headers, proxy-hide-headers modifies the response headers.

salonichf5: as per our offline discussion, we should remove the check for `value != ""` --> NGF knows how to handle those cases 

but if the value is empty, it gets translated to remove headers in NGF. Only thing that bothers me about this is differentiating between set and add headers. I would add a warning about only supporting set headers and not add. which is what you have currently

```
      requestHeaderModifier:
        set:
        - name: My-Overwrite-Header
          value: this-is-the-only-value
        add:
        - name: Accept-Encoding
          value: compress
        - name: My-cool-header
          value: this-is-an-appended-value
```

pkg/i2gw/providers/nginx/crds/gateway_builder.go
```diff
@@ -0,0 +1,391 @@
...
+			}
+
+			virtualServerMap[vs.Name] = append(virtualServerMap[vs.Name], GatewayListenerKey{
+				gatewayName:  gatewayName,
+				listenerName: string(uniqueListeners[key].Name),
+			})
+		}
+	}
+
+	// Process TransportServers to create TCP/TLS/UDP listeners
```
salonichf5, line 196: we dont support UDP listeners

pkg/i2gw/providers/nginx/crds/gateway_builder.go
salonichf5: are you not addressing this?

pkg/i2gw/providers/nginx/crds/gateway_builder.go
sarthyparty: Yeah but the conversion tool will support converting

go.mod
sarthyparty: I'm going to leave the change unless it should be reverted, because it does not let me revert the change without changing my own go version

sarthyparty:
@rikatz @LiorLieberman This PR is a smaller split PR of the full CRD to Gateway API conversion that I'm trying to add. Is this split PR ok? Its not a full conversion, since I didn't want to drop too big of a PR on you guys.

PR #249, opened 2025-10-12
Ensures Gateway-API object fields obey kubebuilder max item validation

### LINKED ISSUE #124: Make sure generated Gateway-API object fields obey kubebuilder max item validation

carmal891:
**What type of PR is this?**
/kind feature

Fixes #124

```release-note
Added validation for Gateway API maxItems constraints. Now returns appropriate error messages when field limits are exceeded during I2GW conversion
```

### DIFF
--- a/pkg/i2gw/validate.go
+++ b/pkg/i2gw/validate.go
@@ -1,0 +1,133 @@
+package i2gw
+
+import (
+	"fmt"
+	"go/ast"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"golang.org/x/tools/go/packages"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+const (
+	KubebuilderMaxItemsMarker = "kubebuilder:validation:MaxItems"
+	MaxItemsPrefix            = "MaxItems="
+)
+
+// ValidateMaxItems validate fields presnet in the gateway resources object
+func ValidateMaxItems(gwResources *GatewayResources) field.ErrorList {
+	var errs field.ErrorList
+	maxItemsMap, err := loadGatewayAPITypes()
+	if err != nil {
+		return append(errs, field.InternalError(nil,
+			fmt.Errorf("unable to fetch kubebuilder max items : %v", err)))
+	}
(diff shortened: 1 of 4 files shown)


pkg/i2gw/validate.go
```diff
@@ -0,0 +1,129 @@
...
+
+		v := reflect.ValueOf(gw)
+		if v.Kind() == reflect.Ptr {
+			//Get the struct
+			v = v.Elem()
+		}
+
+		spec := v.FieldByName("Spec")
+		if !spec.IsValid() || spec.IsZero() {
+			return errs
```
manju956, line 46: Should we not append errors list with invalid spec error here?

carmal891:
@rikatz Could you help review the same ?

manju956 (approved):
LGTM

carmal891:
@Stevenjin8 Is this PR still useful ? It fixes issue #124

Stevenjin8:
@carmal891 I like the idea of doing kubebuilder validation, I just don't know if this is the way to go. There are a bunch of other constaints (regex, enum values, oneof) that crds expose. Is there a generic way we could validate all of them?

carmal891:
> @carmal891 I like the idea of doing kubebuilder validation, I just don't know if this is the way to go. There are a bunch of other constaints (regex, enum values, oneof) that crds expose. Is there a generic way we could validate all of them?

ok I understand your concern. I will check more on how to implement a generic solution to handle all constraints. Thanks for the feedback

PR #250, opened 2025-10-15
Add dependency to httproute IR

Stevenjin8:
/kind cleanup
/kind feature

The current extension process is
1. Create generic translations from ingresses to gwapi crds
2. Read implementation-specific annotations and patch the generated crds

The problem is that step 1 does not output dependencies, which means that when doing step 2, we have to rebuild the dependency tree (which parts of the generated gwapi output correspond with which parts of the ingress output). This is annoying because we are doing double work and is error prone. For example, the ingress-nginx canary extension doesn't handle ingresses that have the same service in multiple paths.

```
# bad.yaml
kind: Ingress
metadata:
  name: myingress
spec:
  ingressClassName: nginx
  rules:
  - host: httpbin
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service: 
            name: httpbinb
            port: 
              number: 80
      - path: /a
        pathType: Prefix
        backend:
          service: 
            name: httpbin
            port: 
              number: 80
---
kind: Ingress
metadata:
  name: my-ingress-canary
  annotations:
    nginx.ingress.kubernetes.io/canary: "true"
    nginx.ingress.kubernetes.io/canary-weight: "10"
spec:
  ingressClassName: nginx
  rules:
  - host: httpbin
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service: 
            name: httpbinc
            port: 
              number: 80
      - path: /a
        pathType: Prefix
        backend:
          service: 
            name: httpbinb
            port: 
              number: 80
```

running the following command gives:

```bash
$ go run . print --input-file bad.yaml --providers ingress-nginx
...
  rules:
  - backendRefs:
    - name: httpbinc
      port: 80
      weight: 10  
    - name: httpbinb
      port: 80
      weight: 10  # this should be 90
    matches:
    - path:
        type: PathPrefix
        value: /
...
```

The bug comes from the fact that the current patching logic doesn't look at the generated httproute's `matches`, but I think the correct solution is to add an extension point where the `HTTPRoute.rules` get generated. This will also make adding filters super nice because I wouldn't need to worry about gateway merging.

This also makes it extremely easy to do things like canaries by header (also in this PR), which generate multiple rules per rule group.

```release-note
Fixed bug where ingress-nginx canary annotations produced incorrect weights.
```

### DIFF
--- a/pkg/i2gw/providers/common/converter.go
+++ b/pkg/i2gw/providers/common/converter.go
@@ -39,8 +39,8 @@ import (
 // consideration any provider specific logic.
 func ToIR(ingresses []networkingv1.Ingress, servicePorts map[types.NamespacedName]map[string]int32, options i2gw.ProviderImplementationSpecificOptions) (intermediate.IR, field.ErrorList) {
 	aggregator := ingressAggregator{
-		RuleGroups:   map[ruleGroupKey]*IngressRuleGroup{},
-		ServicePorts: servicePorts,
+		ruleGroups:   map[ruleGroupKey]*ingressRuleGroup{},
+		servicePorts: servicePorts,
 	}
 
 	var errs field.ErrorList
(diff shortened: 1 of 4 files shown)


pkg/i2gw/providers/common/converter.go
```diff
@@ -141,47 +146,42 @@ type ingressDefaultBackend struct {
 	name         string
 	namespace    string
 	ingressClass string
-	backend      networkingv1.IngressBackend
-}
-
-type ingressPath struct {
```
Stevenjin8: this got moved to i2gw package

Stevenjin8:
used wrong branch. See https://github.com/kubernetes-sigs/ingress2gateway/pull/251

PR #251, opened 2025-10-21
Fix canary and create new dependencies in IR

Stevenjin8:
/kind cleanup
/kind feature

The current extension process is
1. Create generic translations from ingresses to gwapi crds
2. Read implementation-specific annotations and patch the generated crds

The problem is that step 1 does not output dependencies, which means that when doing step 2, we have to rebuild the dependency tree (which parts of the generated gwapi output correspond with which parts of the ingress output). This is annoying because we are doing double work and is error prone. For example, the ingress-nginx canary extension doesn't handle ingresses that have the same service in multiple paths.

```
# bad.yaml
kind: Ingress
metadata:
  name: myingress
spec:
  ingressClassName: nginx
  rules:
  - host: httpbin
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service: 
            name: httpbinb
            port: 
              number: 80
      - path: /a
        pathType: Prefix
        backend:
          service: 
            name: httpbin
            port: 
              number: 80
---
kind: Ingress
metadata:
  name: my-ingress-canary
  annotations:
    nginx.ingress.kubernetes.io/canary: "true"
    nginx.ingress.kubernetes.io/canary-weight: "10"
spec:
  ingressClassName: nginx
  rules:
  - host: httpbin
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service: 
            name: httpbinc
            port: 
              number: 80
      - path: /a
        pathType: Prefix
        backend:
          service: 
            name: httpbinb
            port: 
              number: 80
```

running the following command gives:

```bash
$ go run . print --input-file bad.yaml --providers ingress-nginx
...
  rules:
  - backendRefs:
    - name: httpbinc
      port: 80
      weight: 10  
    - name: httpbinb
      port: 80
      weight: 10  # this should be 90
    matches:
    - path:
        type: PathPrefix
        value: /
...
```

The bug comes from the fact that the current patching logic doesn't look at the generated httproute's `matches`, but I think the correct solution is to add an extension point where the `HTTPRoute.rules` get generated. This will also make adding filters super nice because I wouldn't need to worry about gateway merging.

This also makes it extremely easy to do things like canaries by header (also in this PR), which generate multiple rules per rule group.

I still need to write some tests and update the docs, want to get some high level discussion in before making any more changes.

Also, see https://github.com/kubernetes-sigs/ingress2gateway/issues/229

```release-note
* Added new fields in IR to track the source Ingress of each HTTPRoute BackendRef.
* Fixed incorrect canary weight assignment when a Service appeared under different paths in both canary and non-canary Ingresses.
* Added some more validations to ingress-nginx canary annotation parsing.
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/canary.go
+++ b/pkg/i2gw/providers/ingressnginx/canary.go
@@ -49,7 +49,7 @@ func parseCanaryConfig(ingress *networkingv1.Ingress) (canaryConfig, error) {
 	}
 
 	if weight := ingress.Annotations[canaryWeightAnnotation]; weight != "" {
-		w, err := strconv.Atoi(weight)
+		w, err := strconv.ParseInt(weight, 10, 32)
 		if err != nil {
 			return config, fmt.Errorf("invalid canary-weight annotation %q: %w", weight, err)
 		}
@@ -60,7 +60,7 @@ func parseCanaryConfig(ingress *networkingv1.Ingress) (canaryConfig, error) {
 	}
 
 	if total := ingress.Annotations[canaryWeightTotalAnnotation]; total != "" {
-		wt, err := strconv.Atoi(total)
+		wt, err := strconv.ParseInt(total, 10, 32)
 		if err != nil {
 			return config, fmt.Errorf("invalid canary-weight-total annotation %q: %w", total, err)
 		}
@@ -77,7 +77,7 @@ func parseCanaryConfig(ingress *networkingv1.Ingress) (canaryConfig, error) {
 	return config, nil
 }
 
-func canaryFeature(ingresses []networkingv1.Ingress, servicePorts map[types.NamespacedName]map[string]int32, ir *intermediate.IR) field.ErrorList {
+func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, ir *intermediate.IR) field.ErrorList {
 	ruleGroups := common.GetRuleGroups(ingresses)
 	var errList field.ErrorList
 


pkg/i2gw/providers/common/converter.go
```diff
@@ -251,7 +266,8 @@ func (a *ingressAggregator) toHTTPRoutesAndGateways(options i2gw.ProviderImpleme
 			})
 		}
 
-		httpRoutes = append(httpRoutes, httpRoute)
+		// Default backends don't have source tracking (they come from spec.defaultBackend)
```
Stevenjin8: Still thinking about this

pkg/i2gw/intermediate/intermediate_representation.go
```diff
@@ -95,3 +100,17 @@ type ProviderSpecificServiceIR struct {
...
+type BackendSource struct {
+	// Source Ingress that contributed this backend
+	Ingress *networkingv1.Ingress
+
+	// Exactly one of Path or DefaultBackend must be non-nil.
+	// Path points to the specific HTTPIngressPath that contributed this backend.
+	Path *networkingv1.HTTPIngressPath
+	
+	// DefaultBackend points to the Ingress's spec.defaultBackend that contributed this backend.
+	DefaultBackend *networkingv1.IngressBackend
```
Stevenjin8, line 114: These two don't get used, but might be useful in the future?

pkg/i2gw/providers/common/converter.go
```diff
@@ -365,9 +397,17 @@ func (rg *ingressRuleGroup) configureBackendRef(servicePorts map[types.Namespace
...
+		// Track source for this backend
+		sources = append(sources, intermediate.BackendSource{
+			Ingress: path.sourceIngress,
+			Path:    &path.path,
+		})
 	}
 
-	return removeBackendRefsDuplicates(backendRefs), errors
+	// Remove duplicates while maintaining source tracking
+	// uniqueBackendRefs, uniqueSources := removeBackendRefsDuplicatesWithSources(backendRefs, sources)
```
Stevenjin8: hmmm, what to do if we have duplicates with different sources?

pkg/i2gw/intermediate/intermediate_representation.go
Stevenjin8: these two being Path and DefaultBackend

pkg/i2gw/providers/common/converter.go
```diff
@@ -251,7 +269,18 @@ func (a *ingressAggregator) toHTTPRoutesAndGateways(options i2gw.ProviderImpleme
...
+			sources = [][]intermediate.BackendSource{
+				{
+					{
+						Ingress:        db.sourceIngress,
+						DefaultBackend: &db.backend,
+					},
+				},
+			}
+		}
+		httpRoutes = append(httpRoutes, httpRouteWithSources{route: httpRoute, sources: sources})
```
LiorLieberman: this section can benefit from a quick comment

pkg/i2gw/providers/common/converter.go
LiorLieberman: why does it matter? you get a list of backendRefs and you want to dedup it. Am I missing something?

pkg/i2gw/providers/ingressnginx/canary.go
LiorLieberman, line 1: is it all changes or you changed the order? If its the latter mind pointing on your changes exactly? the diff is unclear

Stevenjin8: yeah the diff is super unclear. TBH a better way to read it is just look at the new version the implementation is quite different

pkg/i2gw/providers/common/converter.go
Stevenjin8: Right now, each backendref has one source. If we want to remove duplicate backendrefs, we have to pick the correct source to keep. For example, if there are are duplicate backendrefs that come from the different ingresses, and one ingress has a canary annotations, we would want to keep the one without the canary annotation, but I guess we can have multiple sources per backendref.

The case I'm somewhat more worried about (maybe irrationally so), is the case were you have the same implementation specific match criteria on the ingress, but the semantics of the match are different due to the ingresses having different annotations (or some other way of configuration). E.g. you have the same implementation specific match, in two ingresses, but they are different flavors of regex. I guess this could also be solved by allowing multiple sources of a backend ref. I just generally have an aversion to processing data when there is a change that it might get undone in the future.

TLDR: no strong feelings on this.

pkg/i2gw/providers/common/converter.go
Stevenjin8: @LiorLieberman added

pkg/i2gw/providers/common/converter.go
Stevenjin8: Update: strong feelings about this. Just because an httproute rule has two backend refs that are the same (in the IR) does not mean we should merge them. For example, by keeping them seperate, provider specific logic could add weight *and* filters to one backendref but not another.

LiorLieberman: in the IR maybe, but the resulted backendRef, in the final route should not be the same.

LiorLieberman: I think this dedup came out of a bug where we had two backednRefs that are exactly the same resulted in a final route output

Stevenjin8: ok makes sense, I'll make a fix

pkg/i2gw/providers/common/utils.go
```diff
@@ -221,14 +227,19 @@ func removeBackendRefsDuplicates(backendRefs []gatewayv1.HTTPBackendRef) []gatew
 			k.Port = *backendRef.Port
 		}
 
-		if _, exists := uniqueKeys[k]; exists {
-			continue
+		if oldRef, exists := uniqueBackendRefs[k]; exists {
+			if oldRef.Weight != nil && backendRef.Weight != nil {
+				*oldRef.Weight += *backendRef.Weight
+			}
```
LiorLieberman, line 233: I am not sure exactly what are we doing here, why we are adding the weights?  What use-case it solves?

pkg/i2gw/providers/ingressnginx/canary_test.go
LiorLieberman, line 1: any reasons we get rid of these tests?

LiorLieberman (commented):
We discussed offline, overall looks good left one question

pkg/i2gw/providers/common/utils.go
LiorLieberman: discussed offline: while thats probably not a common thing to have the same backend from different sources, I think the logic makes sense

pkg/i2gw/providers/ingressnginx/canary_test.go
Stevenjin8: I originally go rid of them because a lot of the logic didn't apply and I thought the existing tests were robust enough, but giving it another look, I think we can add some more testing in as well as improve error handling.

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -42,29 +42,44 @@ type canaryConfig struct {
...
 
 	if weight := ingress.Annotations[canaryWeightAnnotation]; weight != "" {
-		if w, err := strconv.Atoi(weight); err == nil {
-			config.weight = int32(w)
+		w, err := strconv.Atoi(weight)
+		if err != nil {
+			return config, fmt.Errorf("invalid canary-weight annotation %q: %w", weight, err)
 		}
+		if w < 0 {
+			return config, fmt.Errorf("canary-weight must be non-negative, got %d", w)
```
LiorLieberman: nit: does ingress-nginx allow setting a negative number in the annotation?

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -42,29 +42,44 @@ type canaryConfig struct {
...
 
 	if total := ingress.Annotations[canaryWeightTotalAnnotation]; total != "" {
-		if wt, err := strconv.Atoi(total); err == nil && wt > 0 {
-			config.weightTotal = int32(wt)
+		wt, err := strconv.Atoi(total)
+		if err != nil {
+			return config, fmt.Errorf("invalid canary-weight-total annotation %q: %w", total, err)
+		}
+		if wt <= 0 {
+			return config, fmt.Errorf("canary-weight-total must be positive, got %d", wt)
```
LiorLieberman: nit: can be a common function if used twice

LiorLieberman:
/lgtm
/approve

Thanks @Stevenjin8 for all the hard work on this!

PR #255, opened 2025-11-13
feature(ingress-nginx):Add ImplementationSpecific path support

obukhovaa:
**What type of PR is this?**

/kind feature

Implemented support for ImplementationSpecific path type in the ingress-nginx provider, currently it fails with error when converting ingress resource.

```release-note
Implemented support for ImplementationSpecific path type in the ingress-nginx provider
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/implementation_specific.go
+++ b/pkg/i2gw/providers/ingressnginx/implementation_specific.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The Kubernetes Authors.
+Copyright 2026 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/i2gw/providers/ingressnginx/implementation_specific_test.go
+++ b/pkg/i2gw/providers/ingressnginx/implementation_specific_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The Kubernetes Authors.
+Copyright 2026 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -79,8 +79,8 @@ func Test_implementationSpecificHTTPPathTypeMatch(t *testing.T) {
 			expectedValue: "/api/v3/.*",
 		},
 		{
-			name:      "path without use-regex annotation defaults to Prefix",
-			inputPath: "/page/track",
+			name:        "path without use-regex annotation defaults to Prefix",
+			inputPath:   "/page/track",
 			annotations: map[string]string{
 				// No use-regex annotation
 			},
@@ -429,4 +429,3 @@ func Test_implementationSpecificHTTPPathTypeMatch_withCanary(t *testing.T) {
 		})
 	}
 }
-


pkg/i2gw/providers/ingressnginx/implementation_specific.go
```diff
@@ -0,0 +1,29 @@
...
+package ingressnginx
+
+import (
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// implementationSpecificHTTPPathTypeMatch handles the ImplementationSpecific path type
+// for ingress-nginx. When ImplementationSpecific is used, ingress-nginx treats paths
+// as regular expressions, so we convert them to PathMatchRegularExpression.
+func implementationSpecificHTTPPathTypeMatch(path *gatewayv1.HTTPPathMatch) {
```
Stevenjin8: This is only true if `nginx.ingress.kubernetes.io/use-regex: "true"

obukhovaa: addressed

Stevenjin8:
@obukhovaa Hey Artem, sorry I've been kinda MIA in this issue. Taking a look now.

pkg/i2gw/providers/ingressnginx/implementation_specific.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	useRegexAnnotation = "nginx.ingress.kubernetes.io/use-regex"
+)
+
+// implementationSpecificHTTPPathTypeMatch handles the ImplementationSpecific path type
+// for ingress-nginx. When ImplementationSpecific is used, ingress-nginx treats paths
+// as regular expressions only if the annotation "nginx.ingress.kubernetes.io/use-regex: true"
+// is present. Otherwise, it defaults to Prefix matching.
+func implementationSpecificHTTPPathTypeMatch(path *gatewayv1.HTTPPathMatch, ingress *networkingv1.Ingress) {
```
Stevenjin8: Ok, apologies for not mentioning this earlier, I'm still getting familiar with Ingress NGINX. A few things

1. I don't think this logic is exactly right. I think it should be (you should double check me on this)

```
if use-regex == IS:
   return pmRegex
if pathType == IS or pathType == prefix
  return pmPrefix
return pmExact
```

2. 
The other, much more annoying thing, that I worry about, is that if an ingress is canary, it might implicitly have `use-regex=true`. That is, if you have an ingress with regex, and have a corresponding canary ingress, the canary ingress implicitly has `use-regex = "true"`.

3.
Last thing is that `useRegex` can equal anything that `strconv.ParseBool`s to a bool.

obukhovaa: Hey, thanks for the review

1. If I understood your concern correctly, then there's nothing to worry about, since `implementationSpecificHTTPPathTypeMatch` only can be called when pathType is `ImplementationSpecific` and in this case you only have 2 branches, either Regexp if `use-regex: true` or Prefix otherwise (by default Nginx would act like Prefix match, not Exact AFAIK).

2. Agree, if there's a canary (means more than 1 ingress to check) then we shouldn't rely on picking the first one, but rather find non-canary and use it. Accordingly to the docs, any annotations beyond canary related are being ignored on canary ingress in favour of main ingress. I've added a generic way to pick the right ingress to check against, so it can be adjust on provider level, with the same "pick first" default to ensure backward compatibility. 

3. Right.

Stevenjin8: 1. right

Stevenjin8: Ok, I need to do a better job of pointing out issues and suggesting fixes; rather than pointing out issues, getting a response, and then making a suggestion. So this is my bad. Anyways.

I don't think we really have to worry about backcompatibility. This is a CLI tools, so the only API we really expose is the CLI args.

My point being that we can just update 

```diff
- type ImplementationSpecificHTTPPathTypeMatchConverter func(*gatewayv1.HTTPPathMatch, *networkingv1.Ingress)
+ type ImplementationSpecificHTTPPathTypeMatchConverter func(*gatewayv1.HTTPPathMatch, []networkingv1.Ingress)
```

We can get rid of the `selectRepresentativeIngress` interface and for ingress nginx, move it to its `ImplementationSpecificHTTPPathTypeMatchConverter`. So we would end up with something like

```
func implementationSpecificHTTPPathTypeMatch(path *gatewayv1.HTTPPathMatch, ingresses []networkingv1.Ingress) {
	pmPrefix := gatewayv1.PathMatchPathPrefix
	pmRegex := gatewayv1.PathMatchRegularExpression

	if the non canary ingress in ingresses has the regex annotation {
		path.Type = &pmRegex
		return
	}
	path.Type = &pmPrefix
}
```

obukhovaa: Makes sense, let's simplify and isolate logic within implementation specific handler. I've implemented proposed change.

NOTE: regarding backward compatibility, it was rather related to the expected output behaviour, not the input contract; but we kept it consistent anyway, so no problems here.

Stevenjin8:
@obukhovaa sorry, I just realized that I don't think we this isn't as simple as it seems because we need to convert everything with the `use-regex: true` annotation to be a regex path, regardless of whether its implementation specific or not.

I didn't realize there were two PRs addressing the same issue, but ultimately, I think https://github.com/kubernetes-sigs/ingress2gateway/pull/307 is a cleaner solution. Again, this was my mistake as a maintainer and I should have flagged this/realized this earlier.

If you would still like to contribute, I'm looking to get integration tests for regex paths in.

TOHUHM:
Hello 

May I know if the issue solved?  I download v0.5.0, but still got this error

error:
spec.rules[0].http.paths[0].pathType: Invalid value: "ImplementationSpecific": implementationSpecific path type is not supported in generic translation, and your provider does not provide custom support to translate it

version print:

ingress2gateway version: 0.5.0
Built with Go version: go1.25.5

PR #258, opened 2025-11-17
Read resources from multiple files

### LINKED ISSUE #106: Read resources from multiple files

The flag `--input-file` can be used to read input resources from a specific file. It'd be beneficial to improve such a flag to allow providing:
- multiple files
- entire directories to explore and fetch yaml files from.

For the last point, we could even consider a `--recursive` flag to recursively explore nested directories.

**Why this is needed**:

To allow converting multiple files at once.

carmal891:
**What type of PR is this?**

/kind feature

Read resources from multiple files

Fixes #106 

```release-note
The --input-file flag now supports multiple files and directories. When multiple inputs are provided, each file's conversion results are displayed under individual file headers.
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -18,7 +18,9 @@ package cmd
 
 import (
 	"fmt"
+	"io"
 	"os"
+	"path/filepath"
 	"slices"
 	"strings"
 
@@ -54,8 +56,8 @@ type PrintRunner struct {
 	// Defaults to YAML.
 	outputFormat string
 
-	// The path to the input yaml config file. Value assigned via --input-file flag
-	inputFile string
+	// inputFile contains the paths to YAML manifest files to process. Value assigned via --input-file flag.
+	inputFile []string
 
 	// The namespace used to query Gateway API objects. Value assigned via
 	// --namespace/-n flag.
@@ -100,7 +102,58 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
 		return fmt.Errorf("failed to initialize namespace filter: %w", err)
 	}
 
-	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.emitter, pr.getProviderSpecificFlags(), pr.allowExperimentalGatewayAPI)
+	allFiles := []string{}
+
+	for _, path := range pr.inputFile {
+		// Check if path is a directory
+		info, statErr := os.Stat(path)
+		if statErr != nil {
+			return fmt.Errorf("input path does not exist: %s", path)
+		}
+		if info.IsDir() {
+			return fmt.Errorf("provided input path %s is a directory", path)
+		}
+		allFiles = append(allFiles, path)
+	}
+	var gatewayResources []i2gw.GatewayResources
+	var notificationTablesMap map[string]string
+
+	if len(allFiles) == 0 {
+		return fmt.Errorf("no input files specified. Use --input-file to specify at least one file")
+	}
+
+	var readers []io.Reader
+	var inputReader io.Reader
+
+	for i, file := range allFiles {
+		cleanPath := filepath.Clean(file)
+
+		f, openErr := os.Open(cleanPath)
+		if openErr != nil {
+			return fmt.Errorf("error reading file %s: %w", file, openErr)
+		}
+
+		readers = append(readers, f)
+
+		if i < len(allFiles)-1 {
+			readers = append(readers, strings.NewReader("\n---\n"))
+		}
+
+		defer func(f *os.File, path string) {
+			if closeErr := f.Close(); closeErr != nil {
+				fmt.Fprintf(os.Stderr, "warning: failed to close file %s: %v\n", path, closeErr)
+			}
+		}(f, cleanPath)
+	}
+
+	if len(readers) == 1 {
+		inputReader = readers[0]
+	} else {
+		inputReader = io.MultiReader(readers...)
+	}
+
+	gatewayResources, notificationTablesMap, err = i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, inputReader, pr.providers, pr.emitter, pr.getProviderSpecificFlags(), pr.allowExperimentalGatewayAPI)
+
 	if err != nil {
 		return err
 	}
@@ -293,8 +346,10 @@ func (pr *PrintRunner) initializeResourcePrinter() error {
 // 3. If namespace is specified, it filters resources based on that namespace.
 // 4. If no namespace is specified and reading from the cluster, it attempts to get the namespace from the cluster; if unsuccessful, initialization fails.
 func (pr *PrintRunner) initializeNamespaceFilter() error {
+	hasFileInput := len(pr.inputFile) > 0
+
 	// When we should use all namespaces, empty string is used as the filter.
-	if pr.allNamespaces || (pr.inputFile != "" && pr.namespace == "") {
+	if pr.allNamespaces || (hasFileInput && pr.namespace == "") {
 		pr.namespaceFilter = ""
 		return nil
 	}
@@ -302,7 +357,7 @@ func (pr *PrintRunner) initializeNamespaceFilter() error {
 	// If namespace flag is not specified, try to use the default namespace from the cluster
 	if pr.namespace == "" {
 		ns, err := getNamespaceInCurrentContext()
-		if err != nil && pr.inputFile == "" {
+		if err != nil && !hasFileInput {
 			// When asked to read from the cluster, but getting the current namespace
 			// failed for whatever reason - do not process the request.
 			return err
@@ -350,8 +405,8 @@ func newPrintCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&pr.outputFormat, "output", "o", "yaml",
 		"Output format. One of: (yaml, json, kyaml).")
 
-	cmd.Flags().StringVar(&pr.inputFile, "input-file", "",
-		`Path to the manifest file. When set, the tool will read ingresses from the file instead of reading from the cluster. Supported files are yaml and json.`)
+	cmd.Flags().StringSliceVar(&pr.inputFile, "input-file", []string{},
+		`Path to manifest files. When set, the tool will read ingresses from the files instead of reading from the cluster. Supported files are yaml and json.`)
 
 	cmd.Flags().StringVarP(&pr.namespace, "namespace", "n", "",
 		`If present, the namespace scope for this CLI request.`)
--- a/pkg/i2gw/provider.go
+++ b/pkg/i2gw/provider.go
@@ -18,6 +18,7 @@ package i2gw
 
 import (
 	"context"
+	"io"
 	"sync"
 
 	emitterir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate"
@@ -65,7 +66,7 @@ type CustomResourceReader interface {
 
 	// ReadResourcesFromFile reads custom resources associated with
 	// the underlying Provider implementation from the file.
-	ReadResourcesFromFile(ctx context.Context, filename string) error
+	ReadResourcesFromFile(ctx context.Context, reader io.Reader) error
 }
 
 // The ResourcesToIRConverter interface specifies conversion functions from Ingress
--- a/cmd/print_test.go
+++ b/cmd/print_test.go
@@ -149,10 +149,14 @@ func Test_getNamespaceFilter(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
+			var inputFile []string
+			if tc.inputfile != "" {
+				inputFile = []string{tc.inputfile}
+			}
 			pr := PrintRunner{
 				namespace:     tc.namespace,
 				allNamespaces: tc.allNamespaces,
-				inputFile:     tc.inputfile,
+				inputFile:     inputFile,
 			}
 			err = pr.initializeNamespaceFilter()
 
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -17,8 +17,10 @@ limitations under the License.
 package i2gw
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"sort"
 
 	common_emitter "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitters/common_emitter"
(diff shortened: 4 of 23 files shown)


cmd/print.go
```diff
@@ -88,7 +92,43 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+			allFiles = append(allFiles, dirFiles...)
+		} else {
+			allFiles = append(allFiles, path)
+		}
+	}
+
+	if len(allFiles) > 0 {
+		for _, inputFile := range allFiles {
+			fmt.Fprintf(os.Stdout, "=== File: %s ===\n", inputFile)
+			gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, inputFile, pr.providers, pr.getProviderSpecificFlags())
```
Stevenjin8: My understanding here is this implementation is that the following would be more or less equivalent (i forget the specific flag names)

```
i2gw print --input-file a.yaml b.yaml c.yaml 
```

and

```
i2gw print --input-file a.yaml
i2gw print --input-file b.yaml
i2gw print --input-file c.yaml
```

I'm not super familiar with Ingress merging semantics, but could this not mean that we have duplicate resources in the output (maybe with slightly different names).

Ideally, the same set of kubernetes resources should create the same output regardless of which file each resources lives in.

cmd/print.go
```diff
@@ -326,8 +368,11 @@ func newPrintCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&pr.outputFormat, "output", "o", "yaml",
 		"Output format. One of: (yaml, json, kyaml).")
 
-	cmd.Flags().StringVar(&pr.inputFile, "input-file", "",
-		`Path to the manifest file. When set, the tool will read ingresses from the file instead of reading from the cluster. Supported files are yaml and json.`)
+	cmd.Flags().StringSliceVar(&pr.inputPaths, "input-file", []string{},
+		`Path to manifest file(s) or directory(s). When set, the tool will read ingresses from the specified files or all YAML files in directories. Supported files are yaml and json.`)
+
+	cmd.Flags().BoolVar(&pr.recursive, "recursive", false,
```
Stevenjin8: This one is more personal, but I think we can omit this path since users can already do `-input-file manifest-dir/**/*.{yaml,yml,json}`, but if this is a common enough pattern in other places I'm happy to include.

cmd/print.go
carmal891: Thanks for highlighting this. I got your point. So duplicate resource check should happen amongst the set of yaml files when provided as a set like `i2gw print --input-file a.yaml b.yaml c.yaml`

Stevenjin8: Its more like

```
i2gw print --input-file a.yaml b.yaml c.yaml
```
and
```
i2gw print --input-file <(cat a.yaml b.yaml c.yaml)
```
should be equivalent.

cmd/print.go
carmal891: Corrected

cmd/print.go
carmal891: This is corrected now

cmd/print.go
```diff
@@ -96,7 +100,69 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
 		return fmt.Errorf("failed to initialize namespace filter: %w", err)
 	}
 
-	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.emitter, pr.getProviderSpecificFlags())
+	allFiles := []string{}
+
+	for _, path := range pr.inputFile {
+		// Check if path is a directory
+		if info, err := os.Stat(path); err == nil && info.IsDir() {
```
manju956: we need to handle an error case where path is incorrect.

cmd/print.go
```diff
@@ -96,7 +100,69 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+			if err != nil {
+				return fmt.Errorf("error reading directory %s: %w", path, err)
+			}
+			allFiles = append(allFiles, dirFiles...)
+		} else {
+			allFiles = append(allFiles, path)
+		}
+	}
+	var gatewayResources []i2gw.GatewayResources
+	var notificationTablesMap map[string]string
```
manju956, line 119: Can leave a comment explaining the map usage?

cmd/print.go
```diff
@@ -96,7 +100,69 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+
+	if len(allFiles) == 0 {
+		return fmt.Errorf("no input files or directories specified. Use --input-file to specify at least one file")
+	}
+
+	var inputPath string
+
+	if len(allFiles) > 1 {
```
manju956: Can we include a case for single input file as well?

cmd/print.go
carmal891: The notificationTablesMap is used in multiple portions of the code already as part of existing logic. Refer below for its useage:
https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/notifications/notifications.go#L61-L62

cmd/print.go
carmal891: already handled. Refer below else condition

carmal891:
@Stevenjin8 Kindly help to review this once again when possible. It would be much appreciated

cmd/print.go
Stevenjin8: its still here

cmd/print.go
```diff
@@ -289,16 +364,18 @@ func (pr *PrintRunner) initializeResourcePrinter() error {
...
+	if pr.allNamespaces || (hasFileInput && pr.namespace == "") {
 		pr.namespaceFilter = ""
 		return nil
 	}
 
 	// If namespace flag is not specified, try to use the default namespace from the cluster
 	if pr.namespace == "" {
 		ns, err := getNamespaceInCurrentContext()
-		if err != nil && pr.inputFile == "" {
+		if err != nil && len(pr.inputFile) == 0 {
```
Stevenjin8: you can use `hasInputFile` here

cmd/print.go
```diff
@@ -96,7 +100,78 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+		var combinedContent strings.Builder
+
+		for _, file := range allFiles {
+			cleanPath := filepath.Clean(file)
+			content, readErr := os.ReadFile(cleanPath)
+			if readErr != nil {
+				return fmt.Errorf("error reading file %s: %w", file, readErr)
+			}
+			combinedContent.Write(content)
+			combinedContent.WriteString("\n---\n")
```
Stevenjin8: I'd really rather not have it write a file. Instead, could you use https://pkg.go.dev/io#MultiReader to concat the file readers and change `ToGatewayApiResources` to take in a reader rather than a file path

cmd/print.go
carmal891: I misunderstood the comment earlier and thought it was referring to my prior use of input-files instead of input-file. I have now removed the recursive logic as suggested

cmd/print.go
carmal891: Updated logic to use multi reader. Modified ToGatewayApiResources to take in a reader rather than a file path and updated the providers code accordingly

pkg/i2gw/provider.go
```diff
@@ -65,7 +66,8 @@ type CustomResourceReader interface {
 
 	// ReadResourcesFromFile reads custom resources associated with
 	// the underlying Provider implementation from the file.
-	ReadResourcesFromFile(ctx context.Context, filename string) error
+	//ReadResourcesFromFile(ctx context.Context, filename string) error
```
Stevenjin8:
```suggestion
```

cmd/print.go
```diff
@@ -100,7 +102,64 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+		if statErr != nil {
+			return fmt.Errorf("input path does not exist: %s", path)
+		}
+		if info.IsDir() {
+			// Discover manifest files in directory
+			dirFiles, discoveryErr := discoverManifestFiles(path)
+			if discoveryErr != nil {
+				return fmt.Errorf("error reading directory %s: %w", path, discoveryErr)
+			}
+			allFiles = append(allFiles, dirFiles...)
```
Stevenjin8:
```suggestion
			// Discover manifest files in directory
			notify that path is a directory
			continue
```

cmd/print.go
```diff
@@ -100,7 +102,64 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+		} else {
+			allFiles = append(allFiles, path)
+		}
+	}
+	var gatewayResources []i2gw.GatewayResources
+	var notificationTablesMap map[string]string
+
+	if len(allFiles) == 0 {
+		return fmt.Errorf("no input files or directories specified. Use --input-file to specify at least one file")
```
Stevenjin8:
```suggestion
		return fmt.Errorf("no input files specified. Use --input-file to specify at least one file")
```

cmd/print.go
```diff
@@ -350,8 +411,8 @@ func newPrintCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&pr.outputFormat, "output", "o", "yaml",
 		"Output format. One of: (yaml, json, kyaml).")
 
-	cmd.Flags().StringVar(&pr.inputFile, "input-file", "",
-		`Path to the manifest file. When set, the tool will read ingresses from the file instead of reading from the cluster. Supported files are yaml and json.`)
+	cmd.Flags().StringSliceVar(&pr.inputFile, "input-file", []string{},
+		`Path to manifest file(s) or directory(s). When set, the tool will read ingresses from the specified files or all YAML files in directories. Supported files are yaml and json.`)
```
Stevenjin8:
```suggestion
		`Path to the manifest files. When set, the tool will read ingresses from the files instead of reading from the cluster. Supported files are yaml and json.`)
```

cmd/print.go
```diff
@@ -426,3 +487,31 @@ func PrintUnstructuredAsYaml(obj *unstructured.Unstructured) error {
...
+	}
+
+	return files, err
+}
+
+// isManifestFile checks if a file is a YAML or JSON manifest
+func isManifestFile(path string) bool {
+	ext := strings.ToLower(filepath.Ext(path))
+	return ext == ".yaml" || ext == ".yml" || ext == ".json"
+}
```
Stevenjin8:
```suggestion
```

cmd/print.go
```diff
@@ -100,7 +102,59 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+	allFiles := []string{}
+
+	for _, path := range pr.inputFile {
+		// Check if path is a directory
+		info, statErr := os.Stat(path)
+		if statErr != nil {
+			return fmt.Errorf("input path does not exist: %s", path)
+		}
+		if info.IsDir() {
+			fmt.Printf("provided input path %s is a directory \n", path)
```
Stevenjin8:
```suggestion
			fmt.Errorf("provided input path %s is a directory \n", path)
```

carmal891: @Stevenjin8 Please help clarify. The current intent is to inform the user that a given path is a directory and continue processing the remaining inputs. As fmt.Errorf returns an error, do you expect to propagate it and stop execution instead? I’m concerned that this might not be ideal when multiple paths are supplied.

Stevenjin8: oh oops, I thought this was going to get logged to stderr. I still think we should fail early if a directory is passed. Anyways, most of this will change once https://github.com/kubernetes-sigs/ingress2gateway/pull/341 is merged

carmal891:
@Stevenjin8 All review suggestions are addressed

cmd/print.go
```diff
@@ -100,7 +102,58 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+		}
+
+		readers = append(readers, f)
+
+		if i < len(allFiles)-1 {
+			readers = append(readers, strings.NewReader("\n---\n"))
+		}
+
+		defer func(f *os.File, path string) {
+			if closeErr := f.Close(); err != nil {
```
bexxmodd: This should check `closeErr != nil`

cmd/print.go
```diff
@@ -100,7 +102,58 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+	allFiles := []string{}
+
+	for _, path := range pr.inputFile {
+		// Check if path is a directory
+		info, statErr := os.Stat(path)
+		if statErr != nil {
+			return fmt.Errorf("input path does not exist: %s", path)
+		}
+		if info.IsDir() {
+			return fmt.Errorf("provided input path %s is a directory \n", path)
```
bexxmodd: in idiomatic Go, error strings should generally not end with a newline or punctuation.

cmd/print.go
```diff
@@ -100,7 +102,58 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
+
+		f, openErr := os.Open(cleanPath)
+		if openErr != nil {
+			return fmt.Errorf("error reading file %s: %w", file, openErr)
+		}
+
+		readers = append(readers, f)
+
+		if i < len(allFiles)-1 {
+			readers = append(readers, strings.NewReader("\n---\n"))
```
bexxmodd, line 139: Are we sure this will not break JSONs ?

cmd/print.go
```diff
@@ -100,7 +102,58 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
...
-	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.emitter, pr.getProviderSpecificFlags(), pr.allowExperimentalGatewayAPI)
+	allFiles := []string{}
+
+	for _, path := range pr.inputFile {
+		// Check if path is a directory
+		info, statErr := os.Stat(path)
+		if statErr != nil {
+			return fmt.Errorf("input path does not exist: %s", path)
+		}
+		if info.IsDir() {
```
bexxmodd, line 113: It is common for Kubernetes tools to support directory inputs by recursively searching for `.yaml` or `.json` files. Maybe we should consider the same here?

cmd/print.go
carmal891: Yes this was looked into, the  [NewYAMLOrJSONDecoder](https://github.com/kubernetes-sigs/ingress2gateway/blob/b5c9d056ee0f8c723cf4211bff2a78b3bf27fb3b/pkg/i2gw/providers/common/resource_reader.go#L133) supports multi-document streams and treats JSON as valid YAML. So when we add --- between files, the decoder simply interprets it as a document separator and cleanly decodes each object one by one. Therefore, it does not break JSON inputs.

cmd/print.go
carmal891: I think as @Stevenjin8 has mentioned [here](https://github.com/kubernetes-sigs/ingress2gateway/pull/258#pullrequestreview-3760356536), users can already achieve the same behaviour via shell expansion:

`path/to/my/folder/*.{json,yaml,yml}`

Since this provides equivalent functionality without adding complexity to our implementation, I think it’s reasonable to keep the current behaviour and avoid handling directories internally.

carmal891:
@bexxmodd I have addressed your comments. Please go through once again

carmal891:
@Stevenjin8 Can this be merged now that it has been approved by you and @bexxmodd ?

Stevenjin8:
@bexxmodd  can you `/lgtm`

PR #262, opened 2025-11-19
Initial Support for Implementation Emitters

### LINKED ISSUE #260: Add Kgateway as an Emitter

The ability to translate the [providers](https://github.com/kubernetes-sigs/ingress2gateway?tab=readme-ov-file#supported-providers) into Gateway API+[Kgateway](https://github.com/kgateway-dev/kgateway/tree/main/api/v1alpha1) resources that conform to [GEP 713](https://gateway-api.sigs.k8s.io/geps/gep-713/).

**Why this is needed**:

Allows users to migrate to Gateway API conformant implementations such as Kgateway.

danehans:
**What type of PR is this?**
/kind feature

- Adds support for implementation emitters, allowing Gateway API implementations to add implementation-specific resources to the list of output resources.
- Adds kgateway as an initial implementation emitter.
- Adds body buffer feature to ingress-nginx provider (to exercise implementation-specific resources).

Fixes #260 

```release-note
Adds implementation emitters. Adds Kgateway as an initial implementation emitter. Adds body buffer size feature to ingress-nginx provider to exercise initial Kgateway emitter implementation.
```


Stevenjin8:
@danehans We still need to debate some details of having implementation specific emitters, but I do want to get some discussion going on the implementation as I've been thinking about this too. I like this PR much more than the last one, but I do have some thoughts on how we want to do this.

There are two points points.

First, I would really want to have *all the IR be provider agnostic*. Admittedly, the code is not structured this way right now, but from a maintainability POV, having emitters read provider-specific IR is just not tractable as the number of providers and emitters grows. To start, we can have model this IR so its similar to the ingress nginx one.

The second point (that I don't feel as strongly about) is to think about failures as a first class citizen that is  embedded into the API. I'm not too sure how this would work, but I have some ideas. Each section of the IR would also have some sort of status. As the emitter parses it, it will write the status saying it was parsed and if there are any caveats the user should know about. That way if we add a new section of IR, they will be parsed as unamrked automatically by all emitters.

I also realize that there is a good amount of urgency to get something working specifically for ingress nginx (I definitely feel it), but if we really want users to use this tool, then I think its worth taking a step back and thinking a bit.

danehans:
@Stevenjin8 thanks for the review

> First, I would really want to have all the IR be provider agnostic. Admittedly, the code is not structured this way right now, but from a maintainability POV, having emitters read provider-specific IR is just not tractable as the number of providers and emitters grows. To start, we can have model this IR so its similar to the ingress nginx one.

+1 I added the Policy type to IngressNginxHTTPRouteIR, but this could be part of the top-level IR so it's not tied to a specific provider. IMHO refactoring all of the IR to be provider agnostic should not prevent us from proceeding with this PR or some other approach that supports emitting implementation-specific resources. Are you able to work on refacting the IR to be provider agnostic? If so, how long do you anticipate this taking?

> The second point (that I don't feel as strongly about) is to think about failures as a first class citizen that is embedded into the API. I'm not too sure how this would work, but I have some ideas. Each section of the IR would also have some sort of status. As the emitter parses it, it will write the status saying it was parsed and if there are any caveats the user should know about. That way if we add a new section of IR, they will be parsed as unamrked automatically by all emitters.

+1 Is this something we can iterate on so we're not blocked by making progress with the foundational work?

LiorLieberman:
@danehans can you please review/engage on https://docs.google.com/document/d/1EJV0nFtcmIK96_4V0C9UV8jlKqjPmmF1MFVfmpF_6lc/edit?tab=t.0

Stevenjin8:
> +1 Is this something we can iterate on so we're not blocked by making progress with the foundational work?

I really think we should figure out the foundational work **before** we move onto things that rely on the foundation. I don't think this is hard code. So the faster we can get consensus, the faster we can make progress. I'm putting almost all my time into this project right now.

danehans:
xref https://docs.google.com/document/d/1EJV0nFtcmIK96_4V0C9UV8jlKqjPmmF1MFVfmpF_6lc/edit?usp=sharing for discussion related to implementation-specific emitters.

danehans:
Emitters design doc PR xref: https://github.com/kubernetes-sigs/ingress2gateway/pull/265

pkg/i2gw/ingress2gateway.go
```diff
@@ -78,13 +84,40 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
...
+				errs = append(errs, field.InternalError(field.NewPath("implementation", implName), err))
+			}
+			implObjs = append(implObjs, objs...)
+		}
+
 		providerGatewayResources, conversionErrs := provider.ToGatewayResources(ir)
 		errs = append(errs, conversionErrs...)
+
+		// Convert client.Objects to unstructured and append to GatewayExtensions.
+		for _, obj := range implObjs {
```
Stevenjin8, line 106: Ok, I was thinking about this but want to get your throughts.

Right now, GCE is the only provider that takes advantage of provider-specific extensions. Do you think that we could have a GCE case and a non-GCE case?

pkg/i2gw/implementation.go
```diff
@@ -0,0 +1,37 @@
...
+
+// ImplementationEmitter generates GEP-713 style implementation-specific resources
+// from the merged IR.
+type ImplementationEmitter interface {
+	// Name of the implementation.
+	Name() string
+
+	// Emit can mutate the IR (e.g. to inject HTTPRoute filters) and returns
+	// implementation-specific objects.
+	Emit(ir *intermediate.IR) ([]client.Object, error)
```
Stevenjin8, line 33: What do you think of having a `ProviderIR` and a `EmitterIR`. ProviderIR would be the current IR (and we could probably shed a bunch of unused fields). It would serve as an initial translation where all the Ingress resources are read without their annotations, the provider would then take that and do things like change certain httproutes to tlsroutes and fill in provider-agnostic data.

pkg/i2gw/intermediate/provider_ingressnginx.go
```diff
@@ -16,6 +16,24 @@ limitations under the License.
...
 package intermediate
 
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
 type IngressNginxGatewayIR struct{}
-type IngressNginxHTTPRouteIR struct{}
+
+type IngressNginxHTTPRouteIR struct {
```
Stevenjin8, line 25: I think we should refactor the way we are doing provider-specific extensions. That is, we should get rid of as many of them as possible. Right now, only GCE uses them and we could keep them around for backcompatibility if we wanted. Ideally, any IR passed to the emitter would have no reference to the provider (or any ingress-related resource).

Stevenjin8:
@danehans sorry I thought this was a new PR. Do you still want me to do the provider-neutral emitters. It will probably completely overwrite this PR.

Stevenjin8:
Here is my (WIP) PR https://github.com/kubernetes-sigs/ingress2gateway/pull/273

pkg/i2gw/implementation.go
danehans: My initial thought is that ^ may be overcomplicating the solution.

pkg/i2gw/intermediate/provider_ingressnginx.go
danehans: > any IR passed to the emitter would have no reference to the provider (or any ingress-related resource)

Implementations may want to know the source resource that generated the IR.

danehans: IMHO the `Policies` field can be added to the IR instead of a provider-specific IR since all implementations should follow GEP 713.

danehans:
> @danehans sorry I thought this was a new PR. Do you still want me to do the provider-neutral emitters. It will probably completely overwrite this PR.

@Stevenjin8 thanks for the review and feedback. I'll review https://github.com/kubernetes-sigs/ingress2gateway/pull/273 and I can refactor this PR to fit into your emitter work.

danehans:
Closing since https://github.com/kubernetes-sigs/ingress2gateway/pull/265 and https://github.com/kubernetes-sigs/ingress2gateway/pull/273 merged.

PR #263, opened 2025-11-20
feat(conversion): enable best-effort output and silence usage on errors

Cna59:
**What type of PR is this?**

/kind feature

This PR improves the error handling behavior of the `ingress2gateway print` command to support best-effort conversion:

1. **Removes usage display on errors**: When conversion errors occur, the usage information is no longer displayed, providing cleaner error output.
2. **Enables best-effort conversion**: The tool now continues processing and outputs valid converted resources even when some resources contain errors, respecting the `--output` format flag.

Previously, when encountering errors (e.g., unsupported `ImplementationSpecific` pathType), the command would:

- Display the full usage help, cluttering the error output
- Not output any successfully converted resources in the requested format

After this change, the command:

- Shows only the error messages without usage information
- Continues conversion and outputs all valid resources in the requested format (YAML/JSON)
- Displays errors at the end for visibility

Fixes #

**Does this PR introduce a user-facing change?**: Yes

```release-note
The `ingress2gateway print` command now supports best-effort conversion, continuing to output valid resources even when errors are encountered. Error output no longer includes usage information for cleaner display.
```

---

## Testing

To test this PR, use the following command with a file containing both valid and invalid resources:

```bash
./ingress2gateway print --input-file best-effort-input.yaml --providers ingress-nginx --output yaml
```

### Test Input (best-effort-input.yaml)

```yaml
apiVersion: v1
kind: Service
metadata:
  name: valid-service
  namespace: test
spec:
  ports:
  - port: 80
    targetPort: 8080
    name: http
---
apiVersion: v1
kind: Service
metadata:
  name: invalid-service
  namespace: test
spec:
  ports:
  - port: 80
    targetPort: invalid-port
    name: http
---
kind: Ingress
metadata:
  name: valid-ingress
  namespace: test
spec:
  ingressClassName: test-class
  rules:
  - host: valid.example.com
    http:
      paths:
      - path: /valid
        pathType: Prefix
        backend:
          service:
            name: valid-service
            port:
              number: 80
---
kind: Ingress
metadata:
  name: invalid-backend-ingress
  namespace: test
spec:
  ingressClassName: test-class
  rules:
  - host: invalid.example.com
    http:
      paths:
      - path: /invalid
        pathType: Prefix
        backend:
          service:
            name: invalid-service
            port:
              name: non-existent-port
```

### Expected Output

The command should output valid converted Gateway and HTTPRoute resources in YAML format, followed by error messages:

```text
kind: Gateway
metadata:
  annotations:
    gateway.networking.k8s.io/generator: ingress2gateway-v0.5.0-rc1-2-g1877e50-dirty
  name: nginx
  namespace: default
spec:
  gatewayClassName: nginx
  listeners:
  - hostname: nginx.example.com
    name: nginx-example-com-http
    port: 80
    protocol: HTTP
status: {}
---
kind: HTTPRoute
metadata:
  annotations:
    gateway.networking.k8s.io/generator: ingress2gateway-v0.5.0-rc1-2-g1877e50-dirty
  name: test-nginx-nginx-example-com
  namespace: default
spec:
  hostnames:
  - nginx.example.com
  parentRefs:
  - name: nginx
status:
  parents: []
Error:

# Encountered 1 errors # spec.rules[0].http.paths[0].pathType: Invalid value: "ImplementationSpecific": implementationSpecific path type is not supported in generic translation, and your provider does not provide custom support to translate it
```

### DIFF
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -89,15 +89,20 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
 	}
 
 	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.getProviderSpecificFlags())
-	if err != nil {
-		return err
-	}
 
 	for _, table := range notificationTablesMap {
 		fmt.Fprintln(os.Stderr, table)
 	}
 
+	// Output partial results even if there are conversion errors (best-effort conversion)
 	pr.outputResult(gatewayResources)
+
+	// Return error after outputting partial results
+	if err != nil {
+		// Silence usage on conversion errors since these are not command usage errors
+		cmd.SilenceUsage = true
+		return err
+	}
 
 	return nil
 }
--- a/pkg/i2gw/providers/apisix/converter.go
+++ b/pkg/i2gw/providers/apisix/converter.go
@@ -50,9 +50,6 @@ func (c *resourcesToIRConverter) convertToIR(storage *storage) (intermediate.IR,
 	// Convert plain ingress resources to gateway resources, ignoring all
 	// provider-specific features.
 	ir, errs := common.ToIR(ingressList, storage.ServicePorts, c.implementationSpecificOptions)
-	if len(errs) > 0 {
-		return intermediate.IR{}, errs
-	}
 
 	for _, parseFeatureFunc := range c.featureParsers {
 		// Apply the feature parsing function to the gateway resources, one by one.
--- a/pkg/i2gw/ingress2gateway.go
+++ b/pkg/i2gw/ingress2gateway.go
@@ -84,7 +84,8 @@ func ToGatewayAPIResources(ctx context.Context, namespace string, inputFile stri
 	}
 	notificationTablesMap := notifications.NotificationAggr.CreateNotificationTables()
 	if len(errs) > 0 {
-		return nil, notificationTablesMap, aggregatedErrs(errs)
+		// Return partial results along with errors to support best-effort conversion
+		return gatewayResources, notificationTablesMap, aggregatedErrs(errs)
 	}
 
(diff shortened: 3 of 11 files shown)


cmd/print.go
```diff
@@ -89,16 +89,21 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
 	}
 
 	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.getProviderSpecificFlags())
-	if err != nil {
```
Cna59, line 92: This enables best-effort conversion and allows users to see what could be converted

pkg/i2gw/providers/apisix/converter.go
```diff
@@ -50,9 +50,6 @@ func (c *resourcesToIRConverter) convertToIR(storage *storage) (intermediate.IR,
 	// Convert plain ingress resources to gateway resources, ignoring all
 	// provider-specific features.
 	ir, errs := common.ToIR(ingressList, storage.ServicePorts, c.implementationSpecificOptions)
-	if len(errs) > 0 {
```
Cna59, line 53: This enables best-effort conversion and allows users to see what could be converted

Etc..

Cna59:
If you do not want the "best-effort" output i can recreate a pull request with only the "silence usage on errors"

z15alall:
Tested and it's work's usefull to be able to have an output even if there is "error" !

Stevenjin8:
I'm very supportive of this.

Stevenjin8:
@Cna59 we're in the middle of a refactor, but now that I'm a maintainer, I'll leave some more detailed thoughts

 Right now, we have these errors and we have some notifications. Its not clear when we should use one over another. I'm inclined to say that conversion errors should always be treated as notifications (and thus allow conversion). Would love to hear your thoughts.

PR #265, opened 2025-11-25
Emitters

Stevenjin8:
/kind documentation

Please approve if you agree.

For some context: https://docs.google.com/document/d/1EJV0nFtcmIK96_4V0C9UV8jlKqjPmmF1MFVfmpF_6lc/edit?tab=t.0

### DIFF
--- a/docs/emitters.md
+++ b/docs/emitters.md
@@ -13,14 +13,50 @@ This document outlines the design of the emitter system and the policies around 
 
 At a high level, ingress2gateway will have two main components: providers and emitters.
 Providers will read in Ingress-related resources and output an intermediate representation (IR) that is ingress-implementation-neutral.
-Specifically, providers will output Gateway API resources and some IR that captures any additional information that cannot be expressed in Gateway API.
-Emitters will read the IR and output Gateway API resources and implementation-specific resources.
+Specifically, providers will output `EmitterIR` (standard Gateway API resources and some IR that captures any additional information that cannot be expressed in Standard Gateway API).
+
+There will be a common emitter that reads the `EmitterIR` of the provider and translates it to potentially nonstandard Gateway API resources depending on configuration.
+This gives us a common component that will implement logic "use any Gateway API feature vs only use stable Gateway API features".
+The common emitter will output `EmitterIR`.
+
+Implementation-specific Emitters will read the `EmitterIR` from the common emitter and output Gateway API resources along with implementation-specific resources.
 Both providers and emitters MUST log any information that is lost in translation.
 Ideally, when there is new IR and an emitter does not implement it, ingress2gateway should automatically emit notifications.
 
 Note that the emitters MAY modify the Gateway API resources as needed.
 That said, emitters MUST NOT output non-Gateway API resources unless absolutely necessary.
 For example, once Gateway API's CORS support is stable, ingress2gateway should no longer output [Envoy Gateway CORS](https://gateway.envoyproxy.io/docs/tasks/security/cors/).
+
+The architecture looks as follows.
+The `provider` and `emitter` are implementation-specific components.
+So the `provider` could be an "Ingress NGINX provider" that understands Ingress NGINX annotations,
+and the `emitter` could be an "Envoy Gateway emitter" that outputs Envoy Gateway resources.
+
+
+```
++------------------------------+
+|(Ingress + related Resources) |
++-----+------------------------+
+      |
+      | 1) provider.ToIR(resources)
+      v
++-----+--------+
+| Emitter IR 1 |
++-----+--------+
+      |
+      | 2) CommonEmitter.Emit(IR 1)
+      v
++-----+--------+
+| Emitter IR 2 |
++-----+--------+
+      |
+      | 3) emitter.Emit(IR 2)
+      v
++-----+-----------------------------+
+| Gateway API Resources             |
+| Implementation-specific Resources |
++-----------------------------------+
+```
 
 ## Governance and Maintainability
 


docs/emitters.md
```diff
@@ -0,0 +1,72 @@
+# Emitters
+
+## Motivation
+
+At Kubecon 2025, SIG Networking announced the [retirement of Ingress NGINX](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/) giving users only four months to migrate to another solution
+There is general consensus that the ingress2gateway migration tool should be the centerpiece of this migration.
```
LiorLieberman:
```suggestion
At Kubecon 2025, SIG Networking announced the [retirement of Ingress NGINX](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/) giving users only four months to migrate to another solution. 
There is general consensus that the ingress2gateway migration tool should be the centerpiece of this migration.
```

docs/emitters.md
```diff
@@ -0,0 +1,72 @@
...
+
+For ingress2gateway to be useful, we need to output implementation-specific resources.
+We expect third-party implementation to write and maintain this code.
+Thus, we need explicit processes and expectations to govern and maintain third-party code.
+
+## Proposal
+
+Currently, ingress2gateway supports specifying an ingress provider (Ingress NGINX, NGINX Ingress, etc) as input.
+This allows reading configuration and annotations specific to that Ingress implementation.
+We should a configuration vector allows using a specific emitter that outputs implementation-specific resources.
```
LiorLieberman: "we should a configuration" -- typo?

docs/emitters.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Architecture
+
+Even though we have a strong focus on Ingress NGINX, we should try to be ingress provider-neutral.
+That is, an input provider should translate the input Ingress resources into as much Gateway API as possible and some other form of Ingress-neutral IR.
+Then, the emitter will use the Gateway API resources and IR to output Gateway API and implementation specific resources.
+Note that the emitters may modify the Gateway API resources as needed.
+That said, emitters should not output non-Gateway API resources unless absolutely necessary.
+For example, once Gateway API's CORS support is stable, we should not output [Envoy Gateway CORS](https://gateway.envoyproxy.io/docs/tasks/security/cors/)
```
LiorLieberman: does that also include the removal of the support for outputting it?

docs/emitters.md
```diff
@@ -0,0 +1,72 @@
...
+spec:
+  urlRewrite:
+    path:
+      type: ReplaceRegexMatch
+      replaceRegexMatch:
+        pattern: '^/service/([^/]+)(/.*)$'
+        substitution: '\2/instance/\1'
+```
+
+But, we would not allow anything resembling raw Envoy config, such as [Istio's EnvoyFilter](https://istio.io/latest/docs/reference/config/networking/envoy-filter/).
```
LiorLieberman: worth adding something at the end of the sentence like "since Istio EnvoyFilter is not copmliant with Gateway API etc" (note: I am not sure compliant is the right word here. Its more about HTTPRouteFilter being a "blessed/recognized" way to expand gateway api while EnvoyFilter is not)

docs/emitters.md
```diff
@@ -0,0 +1,72 @@
...
+The codeowner can stop the removal process if they demonstrate commitment to the above policy as decided by the maintainers.
+
+## Scope
+
+To keep ingress2gateway from blowing up, we restrict the output ingress2gateway to
+
+* upstream Gateway API resources
+* any GEP 713-compatible resource that is in the scope of upstream Gateway API as decided by the maintainers of ingress2gateway.
+
+For example, we would allow [`HTTPRouteFilter`](https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilter)
```
LiorLieberman:
```suggestion
For example, we would allow [`envoygateway's HTTPRouteFilter`](https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilter)
```

docs/emitters.md
Stevenjin8: technically, EnvoyFilter can be GEP 713 compliant as it has targetRef and you will soon be able to targetRef services to have it attach to waypoints. The reason we don't want EnvoyFilter is that such an resource would never become part of GWAPI core. I'll update the doc.

docs/emitters.md
Stevenjin8: That was my thought

docs/emitters.md
```diff
@@ -0,0 +1,68 @@
...
+At Kubecon 2025, SIG Networking announced the [retirement of Ingress NGINX](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/) giving users only four months to migrate to another solution.
+Ingress NGINX has ~100 custom annotations that extend the Kubernetes Ingress resource that don't have a direct mapping to Gateway API.
+Thus, for ingress2gateway to be useful at helping users migrate, we need to output implementation-specific resources to maximize coverage of said annotations.
+
+We expect third-party implementations to write and maintain emitters for their specific Gateway API implementations.
+This document outlines the design of the emitter system and the policies around third-party code contributions.
+
+## Architecture
+
+At a high level, ingress2gateway will have two main components: Input providers and 0utput emitters.
```
danehans: nit: s/0/o/

docs/emitters.md
```diff
@@ -0,0 +1,68 @@
...
+Providers will read in Ingress-related resources and output an intermediate representation (IR) that is ingress-implementation-neutral.
+Specifically, providers will output Gateway API resources and some IR that captures any additional information that cannot be expressed in Gateway API.
+Emitters will read the IR and output Gateway API resources and implementation-specific resources.
+Both emitters and providers MUST log any information that is lost in translation.
+Ideally, when there is new IR and an emitter does not implement it, the tool should automatically emit notifications.
```
danehans: nit: s/is new/is a new/

docs/emitters.md
logica0419: How about simply mentioning "providers and emitters" here?
I wanna know why you put "Input" and "Output".

docs/emitters.md
```diff
@@ -0,0 +1,68 @@
...
+We expect third-party implementations to write and maintain emitters for their specific Gateway API implementations.
+This document outlines the design of the emitter system and the policies around third-party code contributions.
+
+## Architecture
+
+At a high level, ingress2gateway will have two main components: Input providers and 0utput emitters.
+Providers will read in Ingress-related resources and output an intermediate representation (IR) that is ingress-implementation-neutral.
+Specifically, providers will output Gateway API resources and some IR that captures any additional information that cannot be expressed in Gateway API.
+Emitters will read the IR and output Gateway API resources and implementation-specific resources.
+Both emitters and providers MUST log any information that is lost in translation.
```
logica0419:
```suggestion
Both providers and emitters MUST log any information that is lost in translation.
```

docs/emitters.md
logica0419: Does "the tool" here mean Ingres2Gateway? Or are they providers and emitters?

docs/emitters.md
```diff
@@ -0,0 +1,68 @@
...
+* The code will be removed in the following minor release or within 30 days, whichever comes later.
+* During this time, ingress2gateway maintainers will reach out to the codeowners with two emails spaced one week apart notifying them of the impending removal as well as a GitHub issue.
+
+The codeowner can stop the removal process if they demonstrate commitment to the above policy as decided by the maintainers.
+
+Existing providers will be grandfathered in, but may not be fully integrated into the provider/emitter model.
+
+## Scope
+
+We restrict the output ingress2gateway to keep the project maintainable.
```
logica0419:
```suggestion
We restrict the output of ingress2gateway to keep the project maintainable.
```

docs/emitters.md
Stevenjin8: I was thinking that it specifies that providers read input and emitters write output. But I like it better without

Stevenjin8:
@logica0419 @danehans I think I got most grammar issues. Please take a look!

docs/emitters.md
```diff
@@ -0,0 +1,74 @@
+# Emitters: Design and Policy
```
LiorLieberman: nit:
```suggestion
# Emitters: Design and Governance
```

docs/emitters.md
LiorLieberman: I am not sure I am convinced by the language you added. Obviously we are both on the same page but `because such a resource conflicts with Gateway API's goal of being implementation-neutral.` is confusing. 

To be clear, there may be extensions that gateway api will never bring to core, but they are compliant with GEP-713 and gateway api intention.

Its a nit, but worth getting the language right. @mikemorris is usually the pro in phrasing these things correctly :)

LiorLieberman:
left two nits, LGTM overall

/appove
/hold to fix the last things

LiorLieberman:
dang, 

/approve

Stevenjin8:
@LiorLieberman i only see one nit.

EDIT: oh i resolved the other one i think

docs/emitters.md
Stevenjin8: @LiorLieberman I don't get it. Envoyfilter supports `target_refs` and is probably GEP-713 compliant. There are lots of other reasons why envoyfilter would never be part of gateway api, but the implementation-neutral argument is the one I found most convincing/easiest to articulate.

Is the difference what could be in gatewayapi in principle vs in practice? For example, there may be some feature that could be implemented with gateway api, but wouldn't be because there is little demand it would be too much work for all proxies to support it. This would be an example of something that is possible in principle but not in practice.

kkk777-7:
thanks for leading the work on this!

docs/emitters.md
LiorLieberman: There are many reasons while other things would never be Gateway API, and therefore there are Extensions. I dont think this process diffirentiates between what and what would not be gateway API in the future, nor that we have this guidelines. 

Lots of Envoy Gateway and NGINX specific stuff (among others, I just picked on these two) that will never be Gateway upstream, and will stay as extension, and that fine. The scope, as defined here and before is translating to whatever is Gateway GEP-713 compliant, we can not do this distinction of "and given that it will likely be gateway api in the future" cause no one really knows how to judge that, and thats also the whole point of extensions

Stevenjin8: @LiorLieberman we discussed this a bit in the docs:
* https://docs.google.com/document/d/1EJV0nFtcmIK96_4V0C9UV8jlKqjPmmF1MFVfmpF_6lc/edit?disco=AAABuB9bSfo
* https://docs.google.com/document/d/1EJV0nFtcmIK96_4V0C9UV8jlKqjPmmF1MFVfmpF_6lc/edit?disco=AAABwfhYrpE

Personally, I don't feel too strongly about this as long as we have a way of not reading and writing configs that are essentially raw dataplane configuration (e.g. EnvoyFilter)

Stevenjin8: Also I don't mean "something that might be in gwapi in the future" rather "somethign that is in scope for gateway api". In other words that if follows gateway api's [design principles](https://kubernetes.io/docs/concepts/services-networking/gateway#design-principles). The aspect that I really care the most about is "portability"

danehans:
@Stevenjin8 thanks for leading this effort!

/lgtm

docs/emitters.md
LiorLieberman: Ok. I am confused now. based on [that doc comment your linked](https://docs.google.com/document/d/1EJV0nFtcmIK96_4V0C9UV8jlKqjPmmF1MFVfmpF_6lc/edit?disco=AAABuB9bSfo), is the scope resources that are GEP-713 compliant AND has gateway api in the works?


If so, it will be good to clarify that, and I believe a lot of the extensions proposed now will not be in scope. Its hard to tell whats in the works and whats not IMO, but I also understand @youngnick 's concern from that doc.

The way I read this PR now, I can not tell if its "GEP 713 compliant, period (.)" or "GEP 713 compliant AND has gateway api work in progress"

Not trying to be hard here, just want to set this very clear so the review process for vendors will be super smooth and wont require back and forth. 

/cc @youngnick @kflynn @robscott @rikatz

robscott: @youngnick clarified in that doc that "has gateway api work in progress" should be replaced with "if the Gateway API maintainers and community agree that the config is in scope for Gateway API upstream". I'd agree with that, though I'd simplify it even further to remove "and community" to avoid any possible ambiguity since it may be unclear what signal we should use to determine what the broader community thinks about the topic. I'd expect maintainers to consider perspectives from community members, but it's likely simpler to make the final call here clearer on GW API maintainers.

We should likely maintain an allowlist in ingress2gateway describing features that Gateway API maintainers have deemed as likely future scope for Gateway API (and thus acceptable for ingress2gateway).

Stevenjin8: No worries. My main link was to @youngnick's comment. I kept the grayed out line (i.e. "GEP 713 compliant AND has gateway api work in progress") just to keep the comment. I don't think this PR has anything along the lines of "gateway api work in progress".

Whether something will or will not be part of gateway api is impossible to determine because for something to become part of gateway api requires people to work on the feature and dataplanes to adopt it. But there are things that are fundamentally incompatible for gateway API (namely any type of feature that allows for raw and arbitrary proxy configuration).

What I want to convey is:
* the status/absence of a feature in gateway api should not, in itself, disqualify the feature from become part of ingress2gateway. For example, AFAIK there is no work on regex path substitution support in gateway api, but I'd be happy to include this in ingress2gateway.
* Rather, a feature should only be disqualified from ingress2gateway if it is fundamentally incompatible with gateway api's design (e.g. if the feature is not portable).

Stevenjin8: So right now, this is what the doc says:

```md
We restrict the output of ingress2gateway to keep the project maintainable.
Emitters MUST only output:

* upstream Gateway API resources
* any GEP 713-compatible resource that is in the scope of upstream Gateway API, **but not yet part of it**, as determined by the maintainers of ingress2gateway.

For example, we would allow [EnvoyGateway's `HTTPRouteFilter`](https://gateway.envoyproxy.io/docs/api/extension_types/#httproutefilter)
with a [`replaceRegexMatch`](https://gateway.envoyproxy.io/docs/api/extension_types/#replaceregexmatch):

apiVersion: gateway.envoyproxy.io/v1alpha1
kind: HTTPRouteFilter
metadata:
  name: regex-path-rewrite
spec:
  urlRewrite:
    path:
      type: ReplaceRegexMatch
      replaceRegexMatch:
        pattern: '^/service/([^/]+)(/.*)$'
        substitution: '\2/instance/\1'


But, we would not allow anything resembling raw Envoy config, such as [Istio's EnvoyFilter](https://istio.io/latest/docs/reference/config/networking/envoy-filter/),
because such a resource conflicts with Gateway API's goal of being implementation-neutral.
```

LiorLieberman: I thinks thats LGTM but probably `because such a resource conflicts with Gateway API's goal of being implementation-neutral.` should be removed as its confusing. Or just replace it to say that its not in scope for the Gateway API project.

LiorLieberman:
/lgtm
/approve
Thanks @Stevenjin8 ! 

Feel free to remove the hold when you are ready

docs/emitters.md
```diff
@@ -0,0 +1,74 @@
...
+
+Existing providers will be exempted from these policies, but may not be fully integrated into the provider/emitter model.
+
+## Scope
+
+We restrict the output of ingress2gateway to keep the project maintainable.
+Emitters MUST only output:
+
+* upstream Gateway API resources
+* any GEP 713-compatible resource that is in the scope of upstream Gateway API, **but not yet part of it**, as determined by the maintainers of ingress2gateway.
```
youngnick: Damn, I hate to be that guy again, but we've got some _technically_ incorrect stuff here. Easy to miss, but I think it's important to get it exactly right.

A HTTPRouteFilter object of the type listed below is not a GEP-713 compatible object, but it _is_ a compatible object for an extensionRef.

Also, I think it's important to be explicit that it's ingress2gateway maintainers _plus_ upstream Gateway API maintainers that both have to agree what's in scope.

With all that in mind, how about this?

```suggestion
* upstream Gateway API resources
* any GEP 713-compatible Policy resource that is in the scope of upstream Gateway API, but not yet part of it
* any extensionRef-compatible resource that is in scope of upstream Gateway API, but not yet part of it (this includes things like HTTPRoute filter extensionRefs)

as determined by the maintainers of ingress2gateway and the Gateway API maintainers together.
```

docs/emitters.md
mikemorris: Yea, I would think that ideally the emitter pattern is structured in such a way that the translation of annotations to general Gateway API configuration has already been done by the time the IR reaches the emitter (with something like "successfully translated YAML so far" and "unhandled annotations" passed to an emitter module) so that "removal of support" for outputting impl-specific variants can just be a non-blocking cleanup task because the emitter code path will just get skipped after support is added in core.

docs/emitters.md
mikemorris: I have another likely-controversial edge case - what about annotations populated through to the underlying pod via Gateway's [`.spec.infrastructure.annotations`](https://gateway-api.sigs.k8s.io/reference/1.4/spec/#gatewayinfrastructure)?

docs/emitters.md
```diff
@@ -0,0 +1,74 @@
+# Emitters: Design and Governance
+
+## Motivation
+
+At Kubecon 2025, SIG Networking announced the [retirement of Ingress NGINX](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/), giving users only four months to migrate to another solution.
```
mikemorris:
```suggestion
At KubeCon NA 2025 in Atlanta, SIG Network announced the [retirement of Ingress NGINX](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement/), giving users only four months to migrate to another solution.
```
Nit - there are many different regional KubeCon events, should be more specific.

docs/emitters.md
```diff
@@ -0,0 +1,74 @@
...
+At a high level, ingress2gateway will have two main components: providers and emitters.
+Providers will read in Ingress-related resources and output an intermediate representation (IR) that is ingress-implementation-neutral.
+Specifically, providers will output Gateway API resources and some IR that captures any additional information that cannot be expressed in Gateway API.
+Emitters will read the IR and output Gateway API resources and implementation-specific resources.
+Both providers and emitters MUST log any information that is lost in translation.
+Ideally, when there is new IR and an emitter does not implement it, ingress2gateway should automatically emit notifications.
+
+Note that the emitters MAY modify the Gateway API resources as needed.
+That said, emitters MUST NOT output non-Gateway API resources unless absolutely necessary.
+For example, once Gateway API's CORS support is stable, ingress2gateway should no longer output [Envoy Gateway CORS](https://gateway.envoyproxy.io/docs/tasks/security/cors/).
```
mikemorris, line 28: Have we had any consideration of something like a `--experimental` flag to optionally output core Gateway API configuration that is only available in the experimental channel (and pre-empt impl-specific emitter alternatives)?

Non-blocking question, doesn't need to be part of initial scope.

mikemorris (approved):
I'm wondering if we should add any language about _potential_ future re-architecture to move emitters out-of-tree _if needed_ for maintenance, support or feature scope concerns?

I think the current proposal and keeping thing in-tree for now is a great start that will let us make some much-needed progress on the tackling support for common ingress-nginx annotations, but I'm anticipating that the constraints on permissible outputs may eventually lead to tension with some emitter implementations that wish to support a broader scope, such as ingress-nginx features which may be outside the scope of Gateway API (such as dataplane buffer sizes, body and header length limits, etc).

docs/emitters.md
Stevenjin8: I don't think I'm fully against it. Is there an ingressnginx annotation that would map to something like this?

mikemorris: I'm specifically looking at things like `nginx.ingress.kubernetes.io/proxy-body-size` and `nginx.ingress.kubernetes.io/proxy-buffer-size` and possible outputs in https://istio.io/latest/docs/reference/config/annotations/#ProxyConfig, which has a proposal in progress to be extended as part of https://github.com/istio/istio/issues/57973

youngnick:
This LGTM as a place to start from - we can change this policy if we need to, just need to get all the maintainers on board.
I'll leave final LGTM and merge to one of the ingress2gateway approvers though.

LiorLieberman:
/approve
/lgtm

This is in a good shape to start. Thanks for all the work @Stevenjin8 !

LiorLieberman:
Again, feel free to remove the hold when you are ready.

docs/emitters.md
mikemorris: Thought about this some more and wrote up a more detailed proposal over in https://github.com/kubernetes-sigs/ingress2gateway/pull/273#issuecomment-3613267377, this can be resolved now.

docs/emitters.md
```diff
@@ -0,0 +1,113 @@
...
+We expect third-party implementations to write and maintain emitters for their specific Gateway API implementations.
+This document outlines the design of the emitter system and the policies around third-party code contributions.
+
+## Architecture
+
+At a high level, ingress2gateway will have two main components: providers and emitters.
+Providers will read in Ingress-related resources and output an intermediate representation (IR) that is ingress-implementation-neutral.
+Specifically, providers will output `EmitterIR` (standard Gateway API resources and some IR that captures any additional information that cannot be expressed in Standard Gateway API).
+
+There will be a common emitter that reads the `EmitterIR` of the provider and translates it to potentially nonstandard Gateway API resources depending on configuration.
```
mikemorris, line 18:
```suggestion
Specifically, providers will output `EmitterIR` (some IR that captures configuration expressible through Gateway API core or extended features in the standard or experimental channel, plus any additional information that cannot currently be expressed in Gateway API).

There will be a common emitter that reads the `EmitterIR` of the provider and translates it to Gateway API resources and fields defined as "core" functionality in the Standard channel, as well as possibly "extended" functionality or fields and resources from the Experimental channel depending on configuration and supported features of the specified emitter implementation. See Gateway API documentation on [release channels](https://gateway-api.sigs.k8s.io/concepts/versioning/#release-channels) and [support levels](https://gateway-api.sigs.k8s.io/concepts/conformance/#2-support-levels) for more detail on what may be emitted by the common emitter.
```
Suggesting a slight tweak - biggest impact would be the suggestion that providers only translate to IR, not directly to Gateway API resources, and all in-spec "to Gateway API" logic would be centralized in the common emitter.

This would require a bit more IR surface to cover core Gateway API functionality in the standard channel, but with the benefit of writing and maintaining it in one place rather than separately in each provider.

Stevenjin8: Usually IR is done in the image of the output. I'd much rather think about each step as building on top of the last. Admittedly, this does mean that the line between what should be done at the provider and the common emitter is a littlle fuzzy, but I think we can figure it out as we go.

Standard gatewayapi api components should always be backcompatible. So I think a the effort in translating ingress configuration to "gateway api expressible IR" is pretty comparable to just creating the corresponding gateway api.

mikemorris: Fair enough, that makes sense and SGTM

PR #273, opened 2025-11-26
Emitters Implementation

Stevenjin8:
/kind feature
This might be a long PR description so that this is easier to review.

Not including naming changes/moved code, this PR is actually very small

Currently, this is the flow

```
+--------------------------------+
|  (Ingress + related Resources) |
+--------------------------------+
      |           |
      |           | 1) common.ToIR(...)
      |           v
      |    +------+------------+
      |    |   Initial IR      |
      |    |  (rough skeleton) |
      |    +------+------------+
      |           |
      +---+-------+
          |
          | 2) provider refines
          |
          v
    +-----+-------------+
    |           IR      |
    +-----+-------------+
          |
          | 3) provider.ToGatewayResources(IR)
          |    (just common.ToGatewayResources(IR)
          |     except for GCE)
          v
+---------+----------------------------+
|        Gateway API Resources         |
| (Gateways, HTTPRoutes, etc. emitted) |
+--------------------------------------+
```

Note that step one is not required, rather an implementation detail that all providers use.
This PR changes the flow to be as follows:

```
+--------------------------------+
|  (Ingress + related Resources) |
+--------------------------------+
      |           |
      |           | 1) common.ToIR(...)
      |           v
      |    +------+------------+
      |    |   Provider IR     |
      |    |  (rough skeleton) |
      |    +------+------------+
      |           |
      +---+-------+
          |
          | 2) provider specific logic
          v
+-----+--------+
| Emitter IR 1 |
+-----+--------+
      |
      | 2) CommonEmitter.Emit(IR 1)
      v
+-----+--------+
| Emitter IR 2 |
+-----+--------+
      |
      | 3) emitter.Emit(IR 2)
      v
+-----+-----------------------------+
| Gateway API Resources             |
| Implementation-specific Resources |
+-----------------------------------+
```

As before, step 1 is an implementation detail that all providers implement for convenience.
Steps 1 and 2 are still done by the provider, but now step 3 is by the emitter.
As such, we need to make some changes

## Create ProviderIR and EmitterIR

We do this by turning the old `IR` into `ProviderIR`, but also stripping out as many unused fields as possible.
Fortunately, many of the unused/empty fields were also Provider specific (e.g. IngressNginxProviderIR).

EmitterIR is a whole new struct that providers output

```
type EmitterIR struct {
	Gateways   map[types.NamespacedName]GatewayContext
	HTTPRoutes map[types.NamespacedName]HTTPRouteContext

	GatewayClasses map[types.NamespacedName]GatewayClassContext
	TLSRoutes      map[types.NamespacedName]TLSRouteContext
	TCPRoutes      map[types.NamespacedName]TCPRouteContext
	UDPRoutes      map[types.NamespacedName]UDPRouteContext
	GRPCRoutes     map[types.NamespacedName]networkingv1.Ingress

	BackendTLSPolicies map[types.NamespacedName]BackendTLSPolicyContext
	ReferenceGrants    map[types.NamespacedName]ReferenceGrantContext
}

type GatewayContext struct {
	gatewayv1.Gateway
        // In the future this will be populated with fields that tell the emitter
        // it needs to add something to the gateway that gwapi itself does not offer.
}
```

## Create an Emitter interface

Looks like this

```

type Emitter interface {
	// Emit converts stored IR with the Provider into
	// Gateway API resources and extensions
	Emit(emitter_intermediate.IR) (GatewayResources, field.ErrorList)
}
```

## Open questions

While I was able to get rid of most implementation-specific IR fields because they were just empty structs, GCE was the exception. I think we have two options:
* remove all GCE logic
* Add GCE fields to the emitter IR and create a GCE emitter that is picked when the provider is GCE.

## For reviewing

Most of the changes are naming. Here are important logic/interface changes:

- pkg/i2gw/provider.go: Remove `IRToGatewayAPIConverter` interface; `Providers` no longer output Gateway API resources.
- pkg/i2gw/emitter.go: Add the `Emitter` interface; `Emitters` now handle IR -> Gateway API resource conversion.
- pkg/i2gw/ingress2gateway.go: Change control to use emitters
- pkg/i2gw/providers/gce/*: removed

```release-note
Initial support for Emitters and removed GCE provider
```

### DIFF
--- a/pkg/i2gw/emitter.go
+++ b/pkg/i2gw/emitter.go
@@ -60,4 +60,5 @@ type EmitterName string
 type EmitterConstructor func(conf *EmitterConf) Emitter
 
 type EmitterConf struct {
+	// TODO: add fields as needed.
 }


Stevenjin8:
@LiorLieberman I'm trying to keep the diff as small as possible

pkg/i2gw/providers/common/utils.go
```diff
@@ -190,56 +190,6 @@ func PtrTo[T any](a T) *T {
...
-type uniqueBackendRefsKey struct {
-	Name      gatewayv1.ObjectName
-	Namespace gatewayv1.Namespace
-	Port      gatewayv1.PortNumber
-	Group     gatewayv1.Group
-	Kind      gatewayv1.Kind
-}
-
-// removeBackendRefsDuplicates removes duplicate backendRefs from a list of backendRefs.
-func removeBackendRefsDuplicates(backendRefs []gatewayv1.HTTPBackendRef) []gatewayv1.HTTPBackendRef {
```
Stevenjin8, line 202: This just got moved to `pkg/i2gw/emitters/common/utils.go`

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -0,0 +1,78 @@
...
+
+	BackendTLSPolicies map[types.NamespacedName]BackendTLSPolicyContext
+	ReferenceGrants    map[types.NamespacedName]ReferenceGrantContext
+}
+
+type GatewayContext struct {
+	gatewayv1.Gateway
+}
+
+type HTTPRouteContext struct {
```
Stevenjin8, line 54: In the future, we would add provider specific information here, and the emitter would consume that information

pkg/i2gw/providers/apisix/apisix.go
```diff
@@ -50,14 +50,11 @@ func NewProvider(conf *i2gw.ProviderConf) i2gw.Provider {
...
 // including the apisix specific features.
-func (p *Provider) ToIR() (intermediate.IR, field.ErrorList) {
-	return p.resourcesToIRConverter.convertToIR(p.storage)
-}
-
-func (p *Provider) ToGatewayResources(ir intermediate.IR) (i2gw.GatewayResources, field.ErrorList) {
-	return common.ToGatewayResources(ir)
+func (p *Provider) ToIR() (emitter_intermediate.IR, field.ErrorList) {
+	ir, errs := p.resourcesToIRConverter.convertToIR(p.storage)
+	return provider_intermediate.ToEmitterIR(ir), errs
```
Stevenjin8: In reality, providers should not rely on `ToEmitterIR`. That is, `p.resourcesToIRConverter.convertToIR` should just return `EmitterIR`, but I added this function to minimize the code change.

Stevenjin8:
@danehans @kkk777-7 Ok, I was thinking of some other stuff wrt to this. Specifically how emitters should propagate errors when encountering IR it cannot parse. We want to be useful and let the user know the source of the error. So maybe each member of EmitterIR will implement some interface:

```
type EmitterIRInterface {
  SetParsed()
  SetError(string)
  IsParsed()
  GetError()
  GetSource()
```

This way, when an emitter parses it, it should call `SetParsed()`. If there is a specific reason it wasnt parsed, it can call `SetError` with the specific reason it failed. Then, when the emitter runs, the surrounding code can call `IsParsed` and if its false, it can create an error message with `GetError` and `GetSource`. `GetSource` would have information about the source of this feature (e.g. an annotation on an ingress)

kkk777-7:
> This way, when an emitter parses it, it should call `SetParsed()`. If there is a specific reason it wasnt parsed, it can call `SetError` with the specific reason it failed. Then, when the emitter runs, the surrounding code can call `IsParsed` and if its false, it can create an error message with `GetError` and `GetSource`. `GetSource` would have information about the source of this feature (e.g. an annotation on an ingress)

+1 !

When I was working on the draft PR, I had the same idea in mind.
In the draft PR, I had a map to manage the status, but it feels better to define an interface on the IR itself.
https://github.com/kubernetes-sigs/ingress2gateway/pull/274/files#diff-1d223d8c60e8f1578a5e42034c3e090e1becc6a254edafc4d71ee59e0cd50786R103-R105

It makes sense for Emitter to mark whether it was able to process the IR that the Provider produced.
With such a mechanism, if an Emitter does not support the implementation of a certain Feature,  won't mark it as processed, 
allowing us to detect unsupported features in `ingress2gateway.go`.

Since not all Emitters will be able to immediately support newly added Extension Features,
 I think it would be useful for users to know which Emitters do not support which Features.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -0,0 +1,78 @@
...
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+)
+
+// IR holds specifications of Gateway Objects for supporting Ingress extensions,
+// annotations, and proprietary API features not supported as Gateway core
+// features. An IR field can be mapped to core Gateway-API fields,
+// or provider-specific Gateway extensions.
+type IR struct {
```
kkk777-7: I'm probably just not fully understanding it yet, but what is the benefit of splitting the IR? (Provider IR,  Emitter IR)

mikemorris:
I think we might want to have an additional processing phase in the middle, with the objective being able to centralize as much common "to Gateway API" logic as possible and automatically pre-empt implementation-specific outputs (by preprocessing IR for common/standard features _before_ handing to the final emitter) as the spec adds new functionality. Something like this:

```
+--------------------------------+
|  (Ingress + related Resources) |
+--------------------------------+
      |           |
      |           | 1) common.ToIR(...)
      |           v
      |    +------+------------+
      |    |   Provider IR     |
      |    |  (rough skeleton) |
      |    +------+------------+
      |           |
      +---+-------+
          |
          | 2) provider specific logic
          v
    +-----+-------------+
    |   Core IR      |
    +-----+-------------+
          |
          | 3) common.Emit(IR)
          |
          v
    +-----+-------------+
    |   Emitter IR      |
    +-----+-------------+
          |
          | 4) emitter.Emit(resources{Gateways, HTTPRoutes, etc}, unhandled? IR)
          |
          v
+---------+----------------------------+
|        Gateway API Resources         |
| (Gateways, HTTPRoutes, etc. emitted) |
+--------------------------------------+
```
The idea here is what I was trying to convey in https://github.com/kubernetes-sigs/ingress2gateway/pull/265#discussion_r2587071253 - it could let emitter adoption of new core Gateway API features be a lagging cleanup task rather than a priority, and let ingress2gateway maintainers/contributors write core logic once instead of relying on every provider to translate common functionality.

This `common` intermediate emitter could start with just handling standard/core parts of the spec, but it could be really neat if emitters had a way to specify [GEP-2162 SupportedFeatures](https://gateway-api.sigs.k8s.io/geps/gep-2162/) to configure what extended or experimental resources/fields could be output by this `common` emitter, something like:

```
ingress2gateway --provider=ingress-nginx --emitter=envoy-gateway
```
where the Envoy Gateway emitter module contains a YAML file like https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/reports/v1.4.0/envoy-gateway/experimental-v1.6.0-default-report.yaml#L48-L74 defining these extended or experimental features the common emitter should handle if supported (and a common API on the Emitter interface like Emitter.GetSupportedFeatures() to avoid loading/parsing the file directly).

What I'm not quite clear on is how to best convey an "already parsed" or "unhandled" state in the IR passed from the common stage to the final emitter.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
Stevenjin8: So the provider IR is more of a vestage of the past. IR is probably not the best name for it either. I kept it like this to minimize the diff. also take a look at the updated PR description if you haven't

This is more of a helper struct that providers call internally in their `ToIR` function. So a provider calls `common.ToIR` (maybe we should rename this) and it gives back a very low-effort translation of ingress resources to gateway api (only Gateway and HTTPRoute). This kicker is that ProviderIR also includes source maps. Right now it only exists for httproute and it basically says, "these routes came from these ingresses". Then the provider can parse annotations and do what it needs to do (e.g add weights to backendRefs if there are Ingresses with canary annotations). The provider will then use the ProviderIR + provider specific logic to create EmitterIR, which is the best translation it can do with only GatewayAPI resources and puts any missing things in the EmitterIR. An example would probably be helpful.

Suppose we have an an ingress. The ingress has some cors annotations and there is also path rewrite (I know that CORS is experimental in gwapi, but lets pretend it doesnt exist).

First, the provider would call common.ToIR on the ingresses to create ProviderIR. This will create the equivalent gwapi if the ingress annotations are ignored. Since path rewrite and cors exist via annotations, `common.ToIR` completely ignores them.

Then, the provider can then read the Ingress resource's annotations that come with the ProviderIR and Parse annotations. Using that information, it would create an EmitterIR with an HTTPRoute with a path rewrite filter because that's part of gwapi. It will fill in some other field in the EmitterIR saying that "this route needs these CORS semantics added".

Finally, the emitter would take the EmitterIR and either add the CORS semantics (with an httproute filter in the envoygateway case) or send a warning saying that the CORS semantics could not be implemented.

Stevenjin8: TLDR ProviderIR is an artifact of how the repo used to work and is a intermediate/helper struct for providers to create EmitterIR. Emitters will only ever see EmitterIR.

Stevenjin8:
> I think we might want to have an additional processing phase in the middle, with the objective being able to centralize as much common "to Gateway API" logic as possible and automatically pre-empt implementation-specific outputs (by preprocessing IR for common/standard features _before_ handing to the final emitter) as the spec adds new functionality. Something like this:
> 
> ```
> +--------------------------------+
> |  (Ingress + related Resources) |
> +--------------------------------+
>       |           |
>       |           | 1) common.ToIR(...)
>       |           v
>       |    +------+------------+
>       |    |   Provider IR     |
>       |    |  (rough skeleton) |
>       |    +------+------------+
>       |           |
>       +---+-------+
>           |
>           | 2) provider specific logic
>           v
>     +-----+-------------+
>     |   Core IR      |
>     +-----+-------------+
>           |
>           | 3) common.Emit(IR)
>           |
>           v
>     +-----+-------------+
>     |   Emitter IR      |
>     +-----+-------------+
>           |
>           | 4) emitter.Emit(resources{Gateways, HTTPRoutes, etc}, unhandled? IR)
>           |
>           v
> +---------+----------------------------+
> |        Gateway API Resources         |
> | (Gateways, HTTPRoutes, etc. emitted) |
> +--------------------------------------+
> ```
> 
> The idea here is what I was trying to convey in [#265 (comment)](https://github.com/kubernetes-sigs/ingress2gateway/pull/265#discussion_r2587071253) - it could let emitter adoption of new core Gateway API features be a lagging cleanup task rather than a priority, and let ingress2gateway maintainers/contributors write core logic once instead of relying on every provider to translate common functionality.
> 
> This `common` intermediate emitter could start with just handling standard/core parts of the spec, but it could be really neat if emitters had a way to specify [GEP-2162 SupportedFeatures](https://gateway-api.sigs.k8s.io/geps/gep-2162/) to configure what extended or experimental resources/fields could be output by this `common` emitter, something like:
> 
> ```
> ingress2gateway --provider=ingress-nginx --emitter=envoy-gateway
> ```
> 
> where the Envoy Gateway emitter module contains a YAML file like https://github.com/kubernetes-sigs/gateway-api/blob/main/conformance/reports/v1.4.0/envoy-gateway/experimental-v1.6.0-default-report.yaml#L48-L74 defining these extended or experimental features the common emitter should handle if supported (and a common API on the Emitter interface like Emitter.GetSupportedFeatures() to avoid loading/parsing the file directly).
> 
> What I'm not quite clear on is how to best convey an "already parsed" or "unhandled" state in the IR passed from the common stage to the final emitter.

Makes sense, will implement probably next week @kkk777-7 I don't think will really impact any stuff you need to do

youngnick:
I like @mikemorris' suggestion here too. Thanks again @Stevenjin8, this looks great.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
kkk777-7: thanks for explanation, I understand better!

mikemorris:
Reviewed briefly and most of the diff makes sense (just a lot of splitting up the IR between providers and emitter instead of emitting Gateway API resources directly from the providers), but I didn't look too closely at the functionality or implications of removing the GCE provider - @robscott is there someone you could get to take a closer look at that specifically and advise on what functionality/scope/architecture might make sense for your needs there?

cmd/print.go
```diff
@@ -33,7 +33,7 @@ import (
 	// Call init function for the providers
 	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/apisix"
 	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/cilium"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/gce"
+	// _ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/gce"
```
robscott: Why is GCE singled out here?

pkg/i2gw/providers/gce/extensions/input_extensions.go
```diff
@@ -1,92 +0,0 @@
-/*
```
robscott, line 1: It looks like this PR is removing >2k lines of code specific to the GCE implementation, and removing the corresponding functionality. I think the minimum requirement for a PR that's doing as much refactoring as this one is to maintain the previous capabilities of ingress2gateway unless there's a clear reason we can't do that.

robscott (commented):
Thanks for the work on this @Stevenjin8! Have some questions about removal of GCE provider that I'd like to address before we move forward with this.

/hold

cmd/print.go
Stevenjin8: I talk about it in the PR comments, but GCE both reads GCE CRDs and outputs GCE CRDs so it didn't fit in quite as well into the provider/emitter architecture. I'll add it back, was just curious for some thoughts.

pkg/i2gw/emitters/utils/utils.go
```diff
@@ -0,0 +1,126 @@
+/*
```
Stevenjin8, line 1: This code was just moved

PROVIDER.md
```diff
@@ -1,207 +0,0 @@
-# Adding Providers
```
Stevenjin8, line 1: I plan on adding something like this back with providers and emitters.

cmd/print.go
LiorLieberman: We are planning to have providers and emitters, both overridable, right? 

Most of the current providers right now has support for both input and ouput, except from those who were only ingress or only gateway.

cmd/print.go
```diff
@@ -72,6 +75,9 @@ type PrintRunner struct {
 
 	// Provider specific flags --<provider>-<flag>.
 	providerSpecificFlags map[string]*string
+
+	// emitter indicates which emitter is used to generate the Gateway API resources.
+	emitter string
```
LiorLieberman, line 82: should we have a type, perhaps also maintain a list of supportedOnes for better validation and errors for the user?

cmd/print.go
```diff
@@ -336,6 +342,9 @@ func newPrintCommand() *cobra.Command {
 		`If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even
 if specified with --namespace.`)
 
+	cmd.Flags().StringVar(&pr.emitter, "emitter", "default",
```
LiorLieberman: whats the default one does? only Gateway resources?

pkg/i2gw/emitter.go
```diff
@@ -0,0 +1,63 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
LiorLieberman: 2025

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -0,0 +1,78 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
LiorLieberman: 2025

cmd/print.go
Stevenjin8: Right now, the API allows providers to read anything and output anything really, but in practice, GCE is the only one that has beskpoke logic. The other provders just use a generic `ToGatewayResources` function. So, if we ignore GCE, the existing providers match to the new architecture really nicely. I'll work on a GCE emitter and some one-off logic for backcompatiblity

cmd/print.go
Stevenjin8: totally, there are a few things that this PR leaves out mostly around the api for emitting warnings of untranslatable fields. Do you want me to add that in this PR or make an issue and leave it for another PR

cmd/print.go
Stevenjin8: yes

Stevenjin8:
> Thanks @Stevenjin8!
> 
> Started to review and leaving a few comments but have a more fundamental question.
> 
> Commenting on the diagram in the description:
> 
> > As before, step 1 is an implementation detail that all providers implement for convenience.
> 
> The reason we made **commonIR** as the first step, while still being implemented by the provider (cause an input provider will know best how to translate their stuff to the common IR) is because we wanted to support exactly what mike mentioned. Provider A --> Emitter B. (which is also the most common case we will have with ingress-nginx).
> 
> Why are we moving to a Provider-specific IR instead? what does it give us?

Yeah I think I could have described it better/better naming.  The goal is to still have a common interface between providers and emitters and to support use cases like ingress nginx to envoy gateway. Heck, we could even do a bunch of different ingresses to envoy gateay.

`ProviderIR` is not provider specific. It is meant as IR that is read by the provider. Similarly, `EmitterIR` is IR created by the provider and *read by the emitter*.

The old `IR` struct became `ProviderIR` mainly to minimize the diff, but also because it has source map, and we don't want to expose the emitter to the raw `Ingress` resources

cmd/print.go
mikemorris: Adding guardrails and especially untranslatable warnings feels like scope that could saved for a followup to me to get the initial framework landed first.

LiorLieberman: ack, I am good with a followup

cmd/print.go
LiorLieberman: worth describing this somewhere. will also be easier for new contributors to understand

cmd/print.go
```diff
@@ -314,11 +321,23 @@ func newPrintCommand() *cobra.Command {
...
+			// Auto-set emitter for GCE provider
+			gceProviderUsed := slices.Contains(pr.providers, "gce")
+			emitterFlagChanged := cmd.Flags().Changed("emitter")
+
+			if gceProviderUsed {
+				if emitterFlagChanged && pr.emitter != "gce" {
+					return fmt.Errorf("when using the gce provider, the emitter must be 'gce' (got '%s')", pr.emitter)
+				}
+				pr.emitter = "gce"
+			}
```
LiorLieberman, line 340: I guess I am struggling to understand why GCE is still different? 

We talked about it in different forums now, isnt the framework we are getting in here is "custom provider" and "custom emitter"? If yes, why GCE is special?

pkg/i2gw/intermediate/utils.go
```diff
@@ -96,11 +94,3 @@ func mergeGatewayContexts(irs []IR) (map[types.NamespacedName]GatewayContext, fi
...
 	return newGatewayContexts, errs
 }
-
-func mergedGatewayIR(current, existing ProviderSpecificGatewayIR) ProviderSpecificGatewayIR {
-	var mergedGatewayIR ProviderSpecificGatewayIR
-	// TODO(issue #190): Find a different way to merge GatewayIR, instead of
-	// delegating them to each provider.
-	mergedGatewayIR.Gce = mergeGceGatewayIR(current.Gce, existing.Gce)
-	return mergedGatewayIR
-}
```
LiorLieberman, line 106: is this not needed anymore?

pkg/i2gw/intermediate/provider_gce.go
```diff
@@ -45,18 +45,3 @@ type HealthCheckConfig struct {
...
-	}
-	if existing == nil {
-		return current
-	}
-
-	// If both GceGatewayIRs are not nil, merge their fields.
-	var mergedGatewayIR GceGatewayIR
-	mergedGatewayIR.EnableHTTPSRedirect = current.EnableHTTPSRedirect || existing.EnableHTTPSRedirect
-	return &mergedGatewayIR
-}
```
LiorLieberman, line 62: what replaces this?

pkg/i2gw/emitters/common_emitter/emitter.go
```diff
@@ -14,27 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package nginx
+package commonemitter
```
LiorLieberman: nit: common_emitter ? is this an emitter we are going to use? if not, better to not call the package x_emitter. Maybe emitter_helpers or something idk

pkg/i2gw/emitters/standard/standard_test.go
```diff
@@ -14,41 +14,60 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package common
+package standard
```
LiorLieberman: nit: "standard" package does not communicate anything.. should this be `standard_emitter`?

Also, you changed it from `default` to `standard` but forgot to change the folder name which is still `default`

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -86,41 +110,38 @@ func addGatewayPolicyIfConfigured(gatewayNamespacedName types.NamespacedName, ga
...
 
-func buildGceServiceExtensions(ir intermediate.IR, gatewayResources *i2gw.GatewayResources) {
-	for svcKey, serviceIR := range ir.Services {
+func buildGceServiceExtensions(ir emitterir.EmitterIR, gatewayResources *i2gw.GatewayResources) {
+	for svcKey, serviceIR := range ir.GceServices {
 		bePolicy := addGCPBackendPolicyIfConfigured(svcKey, serviceIR)
 		if bePolicy != nil {
 			obj, err := i2gw.CastToUnstructured(bePolicy)
 			if err != nil {
-				notify(notifications.ErrorNotification, "Failed to cast GCPBackendPolicy to unstructured", bePolicy)
```
LiorLieberman, line 100: why removing the `bePolicy` here?

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -47,23 +71,23 @@ func (c *irToGatewayResourcesConverter) irToGateway(ir intermediate.IR) (i2gw.Ga
...
+func buildGceGatewayExtensions(ir emitterir.EmitterIR, gatewayResources *i2gw.GatewayResources) {
 	for gwyKey, gatewayContext := range ir.Gateways {
-		gwyPolicy := addGatewayPolicyIfConfigured(gwyKey, gatewayContext.ProviderSpecificIR)
+		gwyPolicy := addGatewayPolicyIfConfigured(gwyKey, &gatewayContext)
 		if gwyPolicy == nil {
 			continue
 		}
 		obj, err := i2gw.CastToUnstructured(gwyPolicy)
 		if err != nil {
-			notify(notifications.ErrorNotification, "Failed to cast GCPGatewayPolicy to unstructured", gwyPolicy)
```
LiorLieberman, line 58: why removing the gwyPolicy here?

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -86,41 +110,38 @@ func addGatewayPolicyIfConfigured(gatewayNamespacedName types.NamespacedName, ga
...
 			}
 			gatewayResources.GatewayExtensions = append(gatewayResources.GatewayExtensions, *obj)
 		}
 
-		hcPolicy := addHealthCheckPolicyIfConfigured(svcKey, serviceIR)
+		hcPolicy := addHealthCheckPolicyIfConfigured(svcKey, &serviceIR)
 		if hcPolicy != nil {
 			obj, err := i2gw.CastToUnstructured(hcPolicy)
 			if err != nil {
-				notify(notifications.ErrorNotification, "Failed to cast HealthCheckPolicy to unstructured", hcPolicy)
```
LiorLieberman, line 110: same comment as above for hcPolicy

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -86,41 +110,38 @@ func addGatewayPolicyIfConfigured(gatewayNamespacedName types.NamespacedName, ga
...
 			gatewayResources.GatewayExtensions = append(gatewayResources.GatewayExtensions, *obj)
 		}
 	}
 }
 
-func addGCPBackendPolicyIfConfigured(serviceNamespacedName types.NamespacedName, serviceIR intermediate.ProviderSpecificServiceIR) *gkegatewayv1.GCPBackendPolicy {
-	if serviceIR.Gce == nil {
-		return nil
-	}
+func addGCPBackendPolicyIfConfigured(serviceNamespacedName types.NamespacedName, serviceIR gce.ServiceIR) *gkegatewayv1.GCPBackendPolicy {
```
LiorLieberman: nit: since there is not serviceIR.GCE anymore do s/serviceIR/gceServiceIR/g?

pkg/i2gw/emitters/gce/gce_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
```
LiorLieberman: nit: not sure whats the guidance here but when you change a file and not create a new one not sure if should update

/cc @rikatz @robscott do you know?

pkg/i2gw/emitters/gce/output_extensions.go
```diff
@@ -14,48 +14,49 @@ See the License for the specific language governing permissions and
...
 import (
 	gkegatewayv1 "github.com/GoogleCloudPlatform/gke-gateway-api/apis/networking/v1"
-	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/intermediate"
+	emitterir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate/gce"
 )
 
-func BuildGCPBackendPolicySessionAffinityConfig(serviceIR intermediate.ProviderSpecificServiceIR) *gkegatewayv1.SessionAffinityConfig {
-	affinityType := serviceIR.Gce.SessionAffinity.AffinityType
+func BuildGCPBackendPolicySessionAffinityConfig(serviceIR gce.ServiceIR) *gkegatewayv1.SessionAffinityConfig {
```
LiorLieberman: s/serviceIR/gceServiceIR

LiorLieberman (commented):
Overall lgtm, (I think 🙂), left a few more nits

pkg/i2gw/emitters/gce/gce_test.go
rikatz: yeah, you should not bump the year unless this is a new file. There are some ongoing discussions from steering or whatever place I saw about not adding years anymore on files (we need to adopt it!) but old file headers must not be changed

pkg/i2gw/emitters/standard/standard_test.go
Stevenjin8: oops yeah. Go package names are supposed to be one word. `standard_emitter` gives a linting error. We can disable this check though. Up to you.

pkg/i2gw/emitters/gce/gce_test.go
Stevenjin8: Yeah the difference between what is a new and what is moved is kinda ambiguous. I'll just use git's determination as reference.

cmd/print.go
Stevenjin8: This is from @robscott comment to preserve functionality/backcompatibility when using `--provider gce`. We can remove it too.

pkg/i2gw/intermediate/utils.go
Stevenjin8: no, but maybe there is a good argument for keeping it around. The only place this gets called is with the Kong provider, which is orthogonal to Gce, so it was dead code.

pkg/i2gw/emitters/common_emitter/emitter.go
Stevenjin8: yeah, this emitter will do basic gwapi translation before implementation specific ones. That said, `_` give linting errors. I think we should disable that linter.

pkg/i2gw/intermediate/utils.go
LiorLieberman: Are we retaining Kong functionality as well? I think GCE also used this.. so what replaces this?

pkg/i2gw/emitters/common_emitter/emitter.go
LiorLieberman: whats the default between this and "standard" one then? 

I dont think you can provide --emitter=common, can you?

Stevenjin8: no. I just think its useful to have a common/shared spot for translation layer from the provider to emitter. @youngnick and @mikemorris commented about this, but it would be useful if we want to do things like, "only emit standard gwapi." So common gets run for everything and the output gets passed into the chosen emitter.

pkg/i2gw/intermediate/utils.go
Stevenjin8: we are retaining that functionality. I just don't see a code path where this gets run currently.

Stevenjin8: > The only place this gets called is with the Kong provider, which is orthogonal to Gce, so it was dead code.

I can elaborate on this more. The only code path where `mergedGatewayIR` gets called is when the provider is `Kong`, but the only way `mergedGatewayIR` does anything is when the provider is GCE since it only considers GCE-specific fields.

pkg/i2gw/emitters/common_emitter/emitter.go
rikatz: I think what is missing here is a comment that this emitter is still to be implemented. And probably calling it "base_emitter" if it will always be called sounds better?

pkg/i2gw/emitters/standard/standard_test.go
rikatz: it is weird to have a linter that blocks _ on names, even because one may have something like "bla_test" as a package name that would be used to unit test just the public facing part of an implementation.

I would disable this linter as well

pkg/i2gw/emitters/common_emitter/emitter.go
Stevenjin8: I'll add some comments TODOs. I think `common` its better because it runs for everything. base suggests that every emitter inherits/derives from it.

pkg/i2gw/intermediate/utils.go
LiorLieberman: what happens if --providers=[kong,gce]?

Stevenjin8: @LiorLieberman That's fine. Here is the relevant code block

https://github.com/kubernetes-sigs/ingress2gateway/blob/e79e5af32e28968ade10bcd3f0fca55c9b8d0292/pkg/i2gw/providers/kong/converter.go#L57-L74

Here, `ingressList` is of type `networkingv1.Ingress` and `storage.TCPIngresses` is of type `[]kongv1beta1.TCPIngress`. Since all the `Ingress/TCPIngress` -> IR conversion is done inside Kong's `convert` function, GCE-specific fields will not be set. 

Then, we call `MergeIRs(irs, tcpGatewayIR)`, where the GCE-specific logic will not run because the corresponding fields aren't set.

More generally, when we specify multiple providers, we have the following loop:

https://github.com/kubernetes-sigs/ingress2gateway/blob/e79e5af32e28968ade10bcd3f0fca55c9b8d0292/pkg/i2gw/ingress2gateway.go#L78-L84

The point is that providers never share IR with each other.

LiorLieberman: got it, thanks for the explanation!

LiorLieberman:
Thanks for all the work on this @Stevenjin8 !

/approve
/hold for final lgtm from @rikatz @youngnick @robscott

cmd/print
rikatz, line 1: is this file left intentionally empty?

pkg/i2gw/emitter.go
```diff
@@ -0,0 +1,63 @@
...
+// EmitterConstructorByName is a map of EmitterConstructor functions by a
+// emitter name. Different Emitter implementations should add their construction
+// func at startup.
+var EmitterConstructorByName = map[EmitterName]EmitterConstructor{}
+
+type EmitterName string
+
+type EmitterConstructor func(conf *EmitterConf) Emitter
+
+type EmitterConf struct {
```
rikatz, line 62: followup: I guess this structure will be populated soon, right? if so on a followup add a TODO: here to make it clear

rikatz:
/lgtm
Let's move forward with the implementation. I still look into some more docs on it :)

PR #287, opened 2026-01-05
Ingress NGINX - Support more canary

### LINKED ISSUE #267: [Ingress NGINX] Support more canary
Right now, we only do canary by weight. We should support 

```
nginx.ingress.kubernetes.io/canary-by-header
nginx.ingress.kubernetes.io/canary-by-header-value
```

and emit warnings for 

```
nginx.ingress.kubernetes.io/canary-by-header-pattern
nginx.ingress.kubernetes.io/canary-by-cookie
```

jgreeer:
**What type of PR is this?**

Fixes #267 

```release-note
added support for `nginx.ingress.kubernetes.io/canary-by-header` & `nginx.ingress.kubernetes.io/canary-by-header-value` and emit warnings for `nginx.ingress.kubernetes.io/canary-by-header-pattern` & `nginx.ingress.kubernetes.io/canary-by-cookie`
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/canary.go
+++ b/pkg/i2gw/providers/ingressnginx/canary.go
@@ -181,8 +181,8 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
 			}
 
 			// Set weights if isWeight is true or both header and weight are not set (all traffic should go to non-canary)
-			if canaryConfig.isWeight || !canaryConfig.isHeader {
-				canaryWeight := canaryConfig.weight
+			if canaryConf.isWeight || !canaryConf.isHeader {
+				canaryWeight := canaryConf.weight
 
 				canaryBackend.Weight = &canaryWeight
 				nonCanaryWeight := canaryConf.weightTotal - canaryWeight
@@ -192,10 +192,10 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
 					canarySourceIngress.Namespace, canarySourceIngress.Name, canaryWeight, nonCanaryWeight, canaryConf.weightTotal), &httpRouteContext.HTTPRoute)
 			}
 
-			if canaryConfig.isHeader {
+			if canaryConf.isHeader {
 				var header = "always"
-				if canaryConfig.headerValue != "" {
-					header = canaryConfig.headerValue
+				if canaryConf.headerValue != "" {
+					header = canaryConf.headerValue
 				}
 				canaryBackendCopy := *canaryBackend
 				// Remove weight from the header-matched backend
@@ -205,7 +205,7 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
 						{
 							Headers: []gatewayv1.HTTPHeaderMatch{
 								{
-									Name:  gatewayv1.HTTPHeaderName(canaryConfig.header),
+									Name:  gatewayv1.HTTPHeaderName(canaryConf.header),
 									Value: header,
 								},
 							},
@@ -223,7 +223,7 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
 
 				// When canary-by-header-value is not specified, we need to add a rule for "never"
 				// that routes to the non-canary backend
-				if canaryConfig.headerValue == "" {
+				if canaryConf.headerValue == "" {
 					nonCanaryBackendCopy := *nonCanaryBackend
 					nonCanaryBackendCopy.Weight = nil
 					neverRule := gatewayv1.HTTPRouteRule{
@@ -231,7 +231,7 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
 							{
 								Headers: []gatewayv1.HTTPHeaderMatch{
 									{
-										Name:  gatewayv1.HTTPHeaderName(canaryConfig.header),
+										Name:  gatewayv1.HTTPHeaderName(canaryConf.header),
 										Value: "never",
 									},
 								},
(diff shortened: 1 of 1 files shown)


pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -162,6 +188,51 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
 				notify(notifications.InfoNotification, fmt.Sprintf("parsed canary annotations of ingress %s/%s and set weights (canary: %d, non-canary: %d, total: %d)",
 					canarySourceIngress.Namespace, canarySourceIngress.Name, canaryWeight, nonCanaryWeight, canaryConfig.weightTotal), &httpRouteContext.HTTPRoute)
 			}
+
+			if canaryConfig.isHeader {
+				var header = "always"
```
Stevenjin8, line 196: what's so special about "always"?

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -162,6 +188,51 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
...
+									Value: header,
+								},
+							},
+						},
+					},
+					BackendRefs: []gatewayv1.HTTPBackendRef{canaryBackendCopy},
+				}
+
+				// Add the new rule to HTTPRoute
+				httpRouteContext.HTTPRoute.Spec.Rules = append(httpRouteContext.HTTPRoute.Spec.Rules, newRule)
```
Stevenjin8, line 218: The other thing we want to do here is to update `httpRouteContext.RuleBackendSources` so that other feature parses can parse other annotations

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -42,7 +46,23 @@ func parseCanaryConfig(ingress *networkingv1.Ingress) (canaryConfig, error) {
...
+	if ingress.Annotations[CanaryByHeader] != "" {
+		config.isHeader = true
+	}
+	config.header = ingress.Annotations[CanaryByHeader]
+	config.headerValue = ingress.Annotations[CanaryByHeaderValue]
 	if weight := ingress.Annotations[CanaryWeightAnnotation]; weight != "" {
+		config.isWeight = true
 		w, err := strconv.ParseInt(weight, 10, 32)
 		if err != nil {
 			return config, fmt.Errorf("invalid canary-weight annotation %q: %w", weight, err)
```
Stevenjin8, line 68: There can only be one canary and that canary cannot be both by header/weight/cookie (double check me on this). So we want our code and types to reflect this (have one code path for headers, and one for weight)

> Canary rules are evaluated in order of precedence. Precedence is as follows: canary-by-header -> canary-by-cookie -> canary-weight

https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -31,6 +31,10 @@ import (
 
 // canaryConfig holds the parsed canary configuration from a single Ingress
 type canaryConfig struct {
+	isHeader    bool
+	header      string
+	headerValue string
+	isWeight    bool
```
Stevenjin8, line 37: It shouldn't be possible for both `isHeader == isWeight`. I think the code could be simplified a lot if we had a canaryType enum instead of a bunch of bool fields.

pkg/i2gw/providers/ingressnginx/canary.go
Stevenjin8: oh i see. The docs are so confusing

> nginx.ingress.kubernetes.io/canary-by-header: The header to use for notifying the Ingress to route the request to the service specified in the Canary Ingress. When the request header is set to always, it will be routed to the canary. When the header is set to never, it will never be routed to the canary. For any other value, the header will be ignored and the request compared against the other canary rules by precedence.

I'm still not sure if this is right.

Stevenjin8: yes, this is right.

pkg/i2gw/providers/ingressnginx/canary.go
jgreeer: I think the evaluation is at request handling time and not when the ingress is parsed. When combining both annotations on a canary you check if the header is there first + evaluate it if it is and, if not, route based on the weights. That is what happens when I test it, whether we want to support this conversion or just emit a warning is another question.

Stevenjin8: wait really... ok good thanks for checking

Stevenjin8: oh I get it now. Well that's super annoying. If the header value is always, it will always route to the canary. If tis never, it will never route to the canary, if its something else, then it looks at the weight. That means that we need to create two, not one additional HTTPRouteRules.

jgreeer: right, i'll add a new rule for that then

pkg/i2gw/providers/ingressnginx/canary.go
jgreeer: added this, but i change the array while we are iterating through it. i think this is fine though because we won't iterate through the new backend sources, they are just added on at the end.

pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -162,6 +191,89 @@ func canaryFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]
...
+						},
+					},
+					BackendRefs: []gatewayv1.HTTPBackendRef{canaryBackendCopy},
+				}
+
+				// Add the new rule to HTTPRoute
+				httpRouteContext.HTTPRoute.Spec.Rules = append(httpRouteContext.HTTPRoute.Spec.Rules, newRule)
+
+				// Add the canary and non-canary backend sources to RuleBackendSources for the new rule
+				newRuleBackendSources := []providerir.BackendSource{canaryBackendSource, nonCanaryBackendSource}
```
kkk777-7, line 221: curious why we add also `nonCanaryBackendSource`?

kkk777-7:
@jgreeer thanks for working this!
I left nits comment.

pkg/i2gw/providers/ingressnginx/canary.go
Stevenjin8: Its subtle and hacky (my idea), but a canary ingress uses the annotations of a non-canary annotation (with a few exceptions). So the non canary ingress' annotations (header manipulation, etc) will affect the canary ingress' HTTPRoute.

kkk777-7: thanks!

PR #283, opened 2025-12-15
implement header manipulation for nginxingress

eladmotola:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
adding support for header manipulation
**Which issue(s) this PR fixes**:
https://github.com/kubernetes-sigs/ingress2gateway/issues/271

Fixes #

**Does this PR introduce a user-facing change?**:

```release-note

```

Issue #272, opened 2025-11-25
[Ingress NGINX] Implement Path rewrites


Stevenjin8:
Implement the following annotations or log an error

```
nginx.ingress.kubernetes.io/rewrite-target
nginx.ingress.kubernetes.io/app-root
```

Stevenjin8:
careful with regex

Stevenjin8:
#288 fixes it

Issue #271, opened 2025-11-25
[Ingress NGINX] Implement Header Maniputation


Stevenjin8:
Try to implement these annotations or log error:

```
nginx.ingress.kubernetes.io/x-forwarded-prefix
nginx.ingress.kubernetes.io/custom-headers
nginx.ingress.kubernetes.io/connection-proxy-header
nginx.ingress.kubernetes.io/upstream-vhost
```

Stevenjin8:
[#288](https://github.com/kubernetes-sigs/ingress2gateway/pull/288) fixes this (custom-headers is out of scope for now)

PR #288, opened 2026-01-05
Feature/path rewrite

Stevenjin8:
depends on https://github.com/kubernetes-sigs/ingress2gateway/pull/283

**What type of PR is this?**

/kind feature

Fixes #

```release-note
Add support for `nginx.ingress.kubernetes.io/x-forwarded-prefix` and `nginx.ingress.kubernetes.io/rewrite-target` (non-regex case)
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/annotations.go
+++ b/pkg/i2gw/providers/ingressnginx/annotations.go
@@ -31,6 +31,6 @@ const (
 	ConnectionProxyHeaderAnnotation = "nginx.ingress.kubernetes.io/connection-proxy-header"
 	CustomHeadersAnnotation         = "nginx.ingress.kubernetes.io/custom-headers"
 
-	// R2gex
+	// Regex
 	UseRegexAnnotation = "nginx.ingress.kubernetes.io/use-regex"
 )
--- a/pkg/i2gw/providers/ingressnginx/rewrite_test.go
+++ b/pkg/i2gw/providers/ingressnginx/rewrite_test.go
@@ -1,0 +1,136 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ingressnginx
+
+import (
+	"reflect"
+	"testing"
+
+	emitterir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate"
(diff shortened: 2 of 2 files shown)


pkg/i2gw/providers/ingressnginx/annotations.go
```diff
@@ -30,4 +30,7 @@ const (
 	UpstreamVhostAnnotation         = "nginx.ingress.kubernetes.io/upstream-vhost"
 	ConnectionProxyHeaderAnnotation = "nginx.ingress.kubernetes.io/connection-proxy-header"
 	CustomHeadersAnnotation         = "nginx.ingress.kubernetes.io/custom-headers"
+
+	// R2gex
```
robscott:
```suggestion
	// Regex
```

pkg/i2gw/providers/ingressnginx/headers_test.go
```diff
@@ -34,85 +34,6 @@ func TestHeaderModifierFeature(t *testing.T) {
 		ingress         networkingv1.Ingress
 		expectedHeaders map[string]string
 	}{
-		{
-			name: "x-forwarded-prefix header with rewrite-target",
```
robscott, line 38: Why are these tests being removed?

robscott (approved):
Thanks @Stevenjin8!

pkg/i2gw/providers/ingressnginx/headers_test.go
Stevenjin8: These tests no longer apply here because the `X-Forwarded-Prefix` header is not part of the path rewrite logic, not the standard header logic.

Stevenjin8: on second thought, I do need to write some tests for rewrite

mikemorris:
Refs #271, #272

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -53,6 +53,19 @@ type GatewayContext struct {
...
+	// This is provider-neutral and applied by the common emitter.
+	PathRewriteByRuleIdx map[int]*PathRewrite
+}
+
+// PathRewrite represents provider-neutral path rewrite intent.
+// For now it only supports full-path replacement; more fields may be added later.
+type PathRewrite struct {
+	ReplaceFullPath string
+	// Headers to add on path rewrite.
+	Headers map[string]string
```
mikemorris, line 67: Does this belong inside PathRewrite? The logic below looks correct, just feels like either an odd structure or struct name (versus something like `RequestRewrite` if paths and headers should be siblings).

Stevenjin8: You mean `Headers`? Yeah its not clear to me either. Generally, its nice for one function to worry about one IR field, that way you don't have to worry about things already existing.

mikemorris: Yea having `Headers` as a field inside the `PathRewrite` struct feels weird...

Stevenjin8: oh I remember now why I did it this way. Since headers are pretty straightforward, they aren't really IR, they are just added as a header modifier and it felt weird to add that header into the HTTPRoute, but leave the PathRewrite stuff for later.

Issue #270, opened 2025-11-25
[Ingress NGINX] Support Timeouts


Stevenjin8:
Support the following annotations or log an error:

```
nginx.ingress.kubernetes.io/proxy-connect-timeout
nginx.ingress.kubernetes.io/proxy-send-timeout
nginx.ingress.kubernetes.io/proxy-read-timeout
nginx.ingress.kubernetes.io/proxy-next-upstream
nginx.ingress.kubernetes.io/proxy-next-upstream-timeout
nginx.ingress.kubernetes.io/proxy-next-upstream-tries
nginx.ingress.kubernetes.io/preserve-trailing-slash
nginx.ingress.kubernetes.io/proxy-body-size
```

Stevenjin8:
Ok I thought about this a bit more and I've come to the following conclusion:

* Seems like timeouts are one of the most popular features
* This cannot (directly) be implemented in gwapi at any support level

The conflict is that GWAPI timeouts refer to either the time taken for the whole HTTP response, or the whole HTTP request/reponse flow. OTOH, the nginx apis are at the TCP level. They are about the time between consecutive reads/writes. https://github.com/kubernetes-sigs/gateway-api/issues/1742#issuecomment-1811367123 points out this difference as well.

Although not directly implementable. The goals of these APIs is similar even though the mechanics are different, so there is probably a reasonable way to translate between them. We could have a feature flag that is on (?) by default. When said flag is one, we set `timeouts.request = (proxy_read_timeout + proxy_write_timeout) * multiplier`. We can set multiplier to something like 10.

robscott:
@Stevenjin8 Thanks for all the work to dig into this! This feels like something where we should make a best effort and then print a warning that points to some kind of static doc, maybe part of https://gateway-api.sigs.k8s.io/guides/getting-started/migrating-from-ingress-nginx/, that dives into the difference between timeouts in Ingress-NGINX and Gateway API.

Stevenjin8:
#289 fixes

PR #289, opened 2026-01-05
Feature/timeouts

Stevenjin8:
depends on https://github.com/kubernetes-sigs/ingress2gateway/pull/288

**What type of PR is this?**

/kind feature

Fixes #

```release-note
adds best-effort translation for
* `nginx.ingress.kubernetes.io/proxy-connect-timeout`
* `nginx.ingress.kubernetes.io/proxy-send-timeout`
* `nginx.ingress.kubernetes.io/proxy-read-timeout`
```

### DIFF
--- a/pkg/i2gw/emitter_intermediate/intermediate_representation.go
+++ b/pkg/i2gw/emitter_intermediate/intermediate_representation.go
@@ -64,6 +64,12 @@ type HTTPRouteContext struct {
 	// BodySizeByRuleIdx maps HTTPRoute rule indices to body size intent.
 	// This is provider-neutral and applied by each custom emitter.
 	BodySizeByRuleIdx map[int]*BodySize
+
+	// CorsPolicyByRuleIdx maps HTTPRoute rule indices to CORS policy intent.
+	// This map is populated by providers that support CORS (e.g., via annotations) and is
+	// applied by the CommonEmitter. This separation allows the CORS logic to be provider-neutral
+	// and consistently applied across different providers, subject to feature gating.
+	CorsPolicyByRuleIdx map[int]*gatewayv1.HTTPCORSFilter
 }
 
 // TCPTimeouts holds TCP-level timeout configuration for a single HTTPRoute rule.
--- a/cmd/print.go
+++ b/cmd/print.go
@@ -81,6 +81,9 @@ type PrintRunner struct {
 	// emitter indicates which emitter is used to generate the Gateway API resources.
 	// Defaults to "standard".
 	emitter string
+
+	// allowExperimentalGatewayAPI indicates whether Experimental Gateway API features (like CORS, URLRewrite) should be included in the output.
+	allowExperimentalGatewayAPI bool
 }
 
 // PrintGatewayAPIObjects performs necessary steps to digest and print
@@ -97,7 +100,7 @@ func (pr *PrintRunner) PrintGatewayAPIObjects(cmd *cobra.Command, _ []string) er
 		return fmt.Errorf("failed to initialize namespace filter: %w", err)
 	}
 
-	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.emitter, pr.getProviderSpecificFlags())
+	gatewayResources, notificationTablesMap, err := i2gw.ToGatewayAPIResources(cmd.Context(), pr.namespaceFilter, pr.inputFile, pr.providers, pr.emitter, pr.getProviderSpecificFlags(), pr.allowExperimentalGatewayAPI)
 	if err != nil {
 		return err
 	}
@@ -362,6 +365,8 @@ if specified with --namespace.`)
 
 	cmd.Flags().StringSliceVar(&pr.providers, "providers", []string{},
 		fmt.Sprintf("If present, the tool will try to convert only resources related to the specified providers, supported values are %v.", i2gw.GetSupportedProviders()))
+
+	cmd.Flags().BoolVar(&pr.allowExperimentalGatewayAPI, "allow-experimental-gw-api", false, "If present, the tool will include Experimental Gateway API fields (e.g. CORS, URLRewrite) in the output. Default is false.")
 
 	pr.providerSpecificFlags = make(map[string]*string)
 	for provider, flags := range i2gw.GetProviderSpecificFlagDefinitions() {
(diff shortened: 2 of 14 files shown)


pkg/i2gw/providers/ingressnginx/timeouts.go
```diff
@@ -0,0 +1,111 @@
...
+			maxTimeout *= timeoutMultiplier
+			gwDur := gatewayv1.Duration(maxTimeout.String())
+			eHTTPContext.RequestTimeouts[ruleIdx] = &gwDur
+
+			notify(notifications.InfoNotification, fmt.Sprintf("parsed ingress-nginx proxy timeouts (x%d) from %s/%s for HTTPRoute %s/%s rule %d (timeouts.request): %s",
+				timeoutMultiplier, ingress.Namespace, ingress.Name, key.Namespace, key.Name, ruleIdx, gwDur), &httpRouteContext.HTTPRoute)
+			notify(
+				notifications.WarningNotification,
+				"ingress-nginx only supports TCP-level timeouts; i2gw has made a best-effort translation to Gateway API timeouts.request." +
+				" Please verify that this meets your needs. See documentation: https://gateway-api.sigs.k8s.io/guides/http-timeouts/",
```
sjberman: So essentially you're taking the highest timeout value out of connect/send/read and multiplying by 10 to use that as the full HTTP request timeout?

Stevenjin8: yes, but definitely open to any other suggestions.

sjberman: Since nginx doesn't natively support full HTTP request timeouts, there's really no great solution.

Stevenjin8: correct

mikemorris:
Refs #270

pkg/i2gw/providers/ingressnginx/timeouts.go
mikemorris: @sjberman How did y'all implement the Gateway API timeouts in https://github.com/nginx/nginx-gateway-fabric? (Or have you not implemented timeouts due to this difficulty?)

sjberman: We haven't implemented it yet.

Users can inject raw nginx config, so they can set the TCP level timeouts (and we'll probably expose these soon in our own Policy), but no support for the Gateway API timeouts.

mikemorris: Reference material in https://gateway-api.sigs.k8s.io/geps/gep-1742/ and https://github.com/kubernetes-sigs/gateway-api/issues/1741 might be helpful for suggesting anything we could do to get a closer approximation, @kflynn any ideas from your research a while back?

sjberman: If the ultimate goal in Gateway API is to define a timeout from the point at which a client's request hits the gateway to the time that the gateway responds back to the client, that's just something that nginx won't be able to support today. Now, we may be able to work with the core nginx team to see if it can be built, or write our own custom nginx module, but that has yet to be prioritized.

The existing nginx timeouts just won't suffice for the desire of a true HTTP request timeout.

`proxy_connect_timeout` is the time that an upstream has to accept a connection
`proxy_read_timeout` is the time for reading a response from the upstream. It's set only between two successive read operations, not for the transmission of the whole response. So if a response comes back in parts, as long as each part is within that timeout, we'll read the response forever.
`proxy_send_timeout` is the time for transmitting a request to the upstream. Same rules as read timeout.

It's more likely that a client side connection would timeout first.

pkg/i2gw/providers/ingressnginx/timeouts.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("must be an integer number of seconds")
+	}
+	if secs <= 0 {
+		return 0, fmt.Errorf("must be > 0")
+	}
+	return time.Duration(secs) * time.Second, nil
+}
+
+func timeoutFeature(_ []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, pIR *providerir.ProviderIR, eIR *emitterir.EmitterIR) field.ErrorList {
```
kkk777-7: nits:
should we align the signatures of  `applyXXX` methods?
like [these](https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/providers/ingressnginx/ingressnginx.go#L64-L65)

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -53,6 +53,8 @@ type GatewayContext struct {
 
 type HTTPRouteContext struct {
 	gatewayv1.HTTPRoute
+	// RequestTimeoutsByRule holds desired HTTPRoute.spec.rules[i].timeouts.request values by rule index.
+	RequestTimeouts map[int]*gatewayv1.Duration
```
kkk777-7: nits:
RequestTimeoutByRuleIdx

pkg/i2gw/providers/ingressnginx/timeouts.go
```diff
@@ -0,0 +1,111 @@
...
+						maxTimeout = d
+					}
+				}
+			}
+
+			if !anyTimeout {
+				continue
+			}
+
+			maxTimeout *= timeoutMultiplier
```
kkk777-7: Could you share any links about timeoutMultiplier in ingress nginx?

pkg/i2gw/providers/ingressnginx/timeouts_test.go
kkk777-7, line 1: Non-blocking:
IMO, provider tests only verify that the Emitter IR is generated correctly, focusing on the provider’s responsibilities.

pkg/i2gw/providers/ingressnginx/timeouts.go
kkk777-7: sorry, I found the [thread](https://github.com/kubernetes-sigs/ingress2gateway/pull/289#discussion_r2663133583).

pkg/i2gw/providers/ingressnginx/timeouts.go
Stevenjin8: eventually, but id rather keep things as they are until we have too many.

kkk777-7:
LGTM, thanks!

/lgtm

PR #292, opened 2026-01-07
Fix data race in NotificationAggregator

johananl:
**What type of PR is this?**

/kind bug

I've stumbled upon a data race in NotificationAggregator while working on e2e tests for ingress2gateway. This fixes it.

To test the PR, run `make test` while commenting out the mutex lock in `CreateNotificationTables()`. This should make the test fail.

None

```release-note
NONE
```

### DIFF
--- a/pkg/i2gw/notifications/notifications.go
+++ b/pkg/i2gw/notifications/notifications.go
@@ -63,6 +63,9 @@ func (na *NotificationAggregator) DispatchNotification(notification Notification
 func (na *NotificationAggregator) CreateNotificationTables() map[string]string {
 	notificationTablesMap := make(map[string]string)
 
+	na.mutex.Lock()
+	defer na.mutex.Unlock()
+
 	for provider, msgs := range na.Notifications {
 		providerTable := strings.Builder{}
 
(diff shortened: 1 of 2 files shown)


johananl:
> Nice! Happen to know when exactly the race happens?
> 
> I thought we had
> 
> https://github.com/kubernetes-sigs/ingress2gateway/blob/2175d6334eebb43a893648f7058adb81e2fe1aaf/pkg/i2gw/notifications/notifications.go#L56-L59
> 
> to guard that.

Yes, it happens here:

https://github.com/kubernetes-sigs/ingress2gateway/blob/2175d6334eebb43a893648f7058adb81e2fe1aaf/pkg/i2gw/notifications/notifications.go#L66

We're reading `na.Notifications` without locking the mutex. We need to lock while reading, too, otherwise the value can change while it's being read.

You can see detailed output from the Go race detector by commenting out

```go
	na.mutex.Lock()
	defer na.mutex.Unlock()
```

in `CreateNotificationTables()` and running `go test -race -run TestNotificationAggregatorRace ./pkg/i2gw/notifications/`.

LiorLieberman (commented):
thanks! @johananl 

/lgtm
/approve

PR #294, opened 2026-01-07
Add e2e tests

### LINKED ISSUE #234: Implement e2e test for ingress2gateway

johananl:
**What type of PR is this?**

/kind test

This PR adds an initial implementation of an e2e testing framework for ingress2gateway.

The framework is in pure Go and avoids shelling out to tools such as kubectl or Helm. This makes it highly portable and easy to use.

Use `make e2e` to run the tests. See `README.md` for more info.

~**NOTE**: The tests are affected by a race in the codebase that is fixed in https://github.com/kubernetes-sigs/ingress2gateway/pull/292. Either cherry-pick the fix onto a temporary branch or wait until we merge the fix before running the e2e tests.~

Fixes #234 

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ kind: ensure-kind
 # tests.
 # See README.md for more info.
 .PHONY: e2e
-e2e: ## Run end-to-end tests.
+e2e: build ## Run end-to-end tests.
 	@if [ ! -z "$${KUBECONFIG}" ]; then \
 		echo "ERROR: KUBECONFIG is set in current shell. Refusing to run to avoid touching an"; \
 		echo "unrelated cluster."; \
@@ -133,7 +133,8 @@ e2e: ## Run end-to-end tests.
 		cleanup_kind=true; \
 	fi; \
 	set -x; \
-	KUBECONFIG=$${kubeconfig} go test $(I2GW_GO_TEST_ARGS) $(REPO_ROOT)/e2e; \
+	I2GW_BINARY_PATH=$(REPO_ROOT)/ingress2gateway KUBECONFIG=$${kubeconfig} \
+		go test $(I2GW_GO_TEST_ARGS) $(REPO_ROOT)/e2e; \
 	test_exit_code=$$?; \
 	set +x; \
 	if [ "$${cleanup_kind}" = "true" ] && [ "$${SKIP_CLEANUP}" != "1" ]; then \
--- a/e2e/testutil.go
+++ b/e2e/testutil.go
@@ -17,9 +17,16 @@ limitations under the License.
 package e2e
 
 import (
+	"bufio"
+	"bytes"
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"os"
+	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
@@ -33,24 +40,14 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/utils/ptr"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	"sigs.k8s.io/gateway-api/apis/v1alpha2"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
 	gwclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
-
-	// Import providers to register them.
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/apisix"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/cilium"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/gce"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/ingressnginx"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/istio"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/kong"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/nginx"
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/openapi3"
-
-	// Import emitters to register them.
-	_ "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitters/standard"
 )
 
 // A prefix for all namespaces used in the e2e tests.
@@ -169,16 +166,8 @@ func runTestCase(t *testing.T, tc *testCase) {
 
 	verifyIngresses(ctx, t, tc, ingressAddresses)
 
-	// We pass an empty input file to make i2gw read ingresses from the cluster.
-	res, notif, err := i2gw.ToGatewayAPIResources(ctx, appNS, "", tc.providers, "standard", tc.providerFlags)
-	require.NoError(t, err)
-
-	if len(notif) > 0 {
-		t.Log("Received notifications during conversion:")
-		for _, table := range notif {
-			t.Log(table)
-		}
-	}
+	// Run the ingress2gateway binary to convert ingresses to Gateway API resources.
+	res := runI2GW(ctx, t, kubeconfig, appNS, tc.providers, tc.providerFlags)
 
 	// TODO: Hack! Force correct gateway class since i2gw doesn't seem to infer that from the
 	// ingress at the moment.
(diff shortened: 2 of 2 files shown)


johananl:
# TODO

- [ ] Ensure the CI runs the e2e tests.

johananl:
I'm fixing the linter errors.

johananl:
This turned out to be a LOT of commits since I've changed directions a couple of times while working on this. Happy to squash or reorganize the commits to make reviewing easier.

johananl:
> I'm fixing the linter errors.

Done.

johananl:
> This turned out to be a LOT of commits since I've changed directions a couple of times while working on this. Happy to squash or reorganize the commits to make reviewing easier.

Done. Review should be much easier now.

johananl:
I'm trying to refactor `runTestCase()` to make it shorter and more readable.

e2e/gwapi_istio.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+	"log"
+	"time"
+
+	"helm.sh/helm/v4/pkg/cli"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	istioVersion   = "1.24.1"
```
Stevenjin8:
```suggestion
	istioVersion   = "1.28.2"
```

e2e/ingress_nginx.go
```diff
@@ -0,0 +1,135 @@
...
+	"time"
+
+	"helm.sh/helm/v4/pkg/cli"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	ingressNginxChartVersion = "4.11.0"
```
Stevenjin8: I think this is out of date too

e2e/gwapi_istio.go
johananl: Thanks, fixed.

e2e/ingress_nginx.go
johananl: Yup. Fixed.

e2e/README.md
```diff
@@ -0,0 +1,64 @@
...
+>deliberately refuse to run when this var is set to avoid accidental execution against an unrelated
+>cluster.
+
+### Running specific tests
+
+A specific subset of the tests can be run by passing a value to `go test -run` using
+`I2GW_RUN_TESTS`:
+
+```bash
+I2GW_RUN_TESTS="TestIngressNginx" make e2e
```
Stevenjin8: Can we expose just a general "args to pass to go test" so we can pass in things like `-count`, etc

e2e/dummy_app.go
```diff
@@ -0,0 +1,160 @@
...
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  name,
+							Image: "registry.k8s.io/e2e-test-images/agnhost:2.39",
```
Stevenjin8: not familiar with agnhost, but is what is the selling point of using it over httpbin? AFAIK, httpbin can do things like delay responses that would allow us to test things like timeouts. agnhost has tcp/udp support, but that's out of scope of ingress.

johananl: I went with agnhost because I saw it used in e2e tests by other k8s-related projects so I thought it's a more "official" image than httpbin. Happy to swap to httpbin (either the original Python-based or the Go clone) if you think that's better.

johananl: But I guess we'd have to host the image at a suitable location (and maybe maintain it...).

e2e/port_forward.go
```diff
@@ -0,0 +1,346 @@
...
+			}
+			return 0, fmt.Errorf("named port %q not found in pod", portName)
+		}
+	}
+
+	return 0, fmt.Errorf("service port %d not found", servicePort)
+}
+
+// Finds the service created by a Gateway API implementation for a Gateway object.
+func findGatewayService(
```
Stevenjin8, line 250: Have you looked at https://gateway-api.sigs.k8s.io/reference/1.4/spec/#gatewaystatusaddress

e2e/port_forward.go
```diff
@@ -0,0 +1,346 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("finding service for gateway %s/%s: %w", gatewayNamespace, gatewayName, err)
+	}
+
+	return out, nil
+}
+
+// Finds the service for an ingress controller.
+func findIngressControllerService(
```
Stevenjin8, line 319: same here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#ingressloadbalanceringress-v1-networking-k8s-io

e2e/scenarios_test.go
```diff
@@ -0,0 +1,614 @@
...
+)
+
+// NOTE: Setting the Host field in ingress rules and in verifiers is optional. When omitted, a
+// random host is generated and used automatically for all ingress objects and verifiers in the
+// test case. Most test cases likely don't need an explicit Host value since the value doesn't
+// matter as long as the verifier verifies the correct Host. In case a specific Host value is
+// important for some test cases, it's important to pay attention to duplicate Host values across
+// test cases: While k8s allows defining multiple ingress objects with identical Host values,
+// whether doing so makes sense (or even works) depends on the ingress controller and can influence
+// test results.
```
Stevenjin8, line 35: Is there any way to generate a random host and use it? Most users of ingress define a host.

Stevenjin8: should have read a bit more.

e2e/scenarios_test.go
```diff
@@ -0,0 +1,614 @@
...
+			runTestCase(t, &testCase{
+				gatewayImplementation: istio.ProviderName,
+				providers:             []string{ingressnginx.Name},
+				providerFlags: map[string]map[string]string{
+					ingressnginx.Name: {
+						ingressnginx.NginxIngressClassFlag: ingressnginx.NginxIngressClass,
+					},
+				},
+				ingresses: []*networkingv1.Ingress{
+					{
```
Stevenjin8: can we make a `newSimpleIngress` or something?

e2e/scenarios_test.go
```diff
@@ -0,0 +1,614 @@
...
+	// 	t.Run("basic conversion", func(t *testing.T) {
+	// 		runTestCase(t, &testCase{
+	// 			gatewayImplementation: cilium.Name,
+	// 			...
+	// 		})
+	// 	})
+	// })
+}
+
+func TestKong(t *testing.T) {
```
Stevenjin8: Since kong does both gwapi and ingress, lets make it specific

e2e/scenarios_test.go
Stevenjin8: this way we dont have stuff like

```go
                }
            }
        }
    }
}
```

e2e/scenarios_test.go
```diff
@@ -0,0 +1,614 @@
...
+							path: "/",
+						},
+					},
+				},
+			})
+		})
+	})
+}
+
+func TestMultipleProviders(t *testing.T) {
```
Stevenjin8, line 206: what is this testing that hasn't been tested before? that it works when there are two ingresses?

e2e/util.go
```diff
@@ -0,0 +1,48 @@
...
+// Generates a cryptographically random alphanumeric string of length n. Uses crypto/rand to ensure
+// uniqueness even when called from parallel tests.
+func randString(n int) (string, error) {
+	const letters = "abcdefghijklmnopqrstuvwxyz0123456789"
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	for i := range b {
+		b[i] = letters[b[i]%byte(len(letters))]
```
Stevenjin8:
```suggestion
		b[i] = letters[b[i % len(letters))]]
```

e2e/verifiers.go
```diff
@@ -0,0 +1,95 @@
...
+			log.Logf("Sending HTTP GET: %v", err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		if res.StatusCode != http.StatusOK {
+			res.Body.Close()
+			lastErr = fmt.Errorf("unexpected HTTP status code: got %d, want %d", res.StatusCode, http.StatusOK)
+			time.Sleep(1 * time.Second)
+			continue
+		}
```
Stevenjin8: can we decouple the retry behaviour and the verify behaviour? We can do a `withRetry(settings TimeoutSettings, verifier Verifier) Verifier)` or something like that.

e2e/dummy_app.go
johananl: We also might want to use consts for the URL and version (note to self).

e2e/README.md
johananl: Done.

e2e/dummy_app.go
johananl: Reference: https://pkg.go.dev/k8s.io/kubernetes/test/images/agnhost

Looking at https://hub.docker.com/r/kennethreitz/httpbin, the image was last pushed in 2018...

https://hub.docker.com/r/mccutchen/go-httpbin looks more recent, but still -- do we want the tests to depend on an image built and maintained by an individual developer?

I'm happy to replace agnhost with anything else, just trying to figure out what would make sense @Stevenjin8.

johananl: > We also might want to use consts for the URL and version (note to self).

Done.

e2e/port_forward.go
johananl: I haven't. Thanks.

But now that I looked at it, I think we have a problem: AFAICT the gateway status addresses are populated only after the gateway's service had been given a LoadBalancer address. In the current setup this never happens since we removed MetalLB 😦 

This makes me question whether the tests are useful when we use port forwarding to get to the gateway's service. I need to better understand the relationship between the gateway and its related service. Maybe we are currently **bypassing** the gateway.

e2e/port_forward.go
johananl: Same as with the gateway. I need to make sure we aren't bypassing the ingress controller.

e2e/port_forward.go
johananl: OK, I checked and we **aren't** bypassing the gateway. I know this because enabling access logs for Envoy shows the verifier requests. So we're utilizing the implementation's proxies even with the port forwarding trick. But we can't rely on things like gateway status address because there **are no** status addresses since we aren't reconciling the LoadBalancer service using MetalLB.

e2e/port_forward.go
johananl: >I need to make sure we aren't bypassing the ingress controller.

We aren't.

e2e/scenarios_test.go
johananl: Done. Switched to the builder pattern.

e2e/scenarios_test.go
johananl: Done.

e2e/scenarios_test.go
johananl: It's testing mainly that the e2e testing framework itself can handle the deployment of more than one ingress provider. We can probably remove it once we have at least one test case which uses multiple ingress providers. Do you think we should remove this now?

e2e/util.go
johananl: I don't understand this suggestion. Could you explain? Also - it doesn't compile for me.

johananl: I've added a comment in the code to clarify exactly what we're doing here since I think it isn't trivial.

e2e/verifiers.go
johananl: Good idea. Done.

johananl:
I found a bug which causes notifications to bleed between test cases. I'll open a fix soon (in a separate PR since it's a generic i2gw problem).

johananl:
> I found a bug which causes notifications to bleed between test cases. I'll open a fix soon (in a separate PR since it's a generic i2gw problem).

I resorted to calling `ingress2gateway` as a binary from the test code instead of calling the main Go function. Although I didn't like the idea initially, I am now convinced that it's a better approach (see [here](https://github.com/kubernetes-sigs/ingress2gateway/issues/234#issuecomment-3779861232) for elaboration) and as a byproduct it also solves the notifications bug.

e2e/clients.go
```diff
@@ -0,0 +1,85 @@
...
+	restConfig, err := cfg.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	// Increase rate limits for parallel test execution.
+	// Default QPS is 5, Burst is 10, which is too low for parallel e2e tests
+	// that make many API calls concurrently.
+	restConfig.QPS = 50
+	restConfig.Burst = 100
```
bexxmodd, line 82: Maybe make these two configurable through env variables (with 50 and 100 being defaults)?

Makefile
```diff
@@ -64,3 +74,75 @@ build: vet;$(info $(M)...Build the binary.)  @ ## Build the binary.
...
+		$(MAKE) kind || exit 1; \
+		kubeconfig="$(REPO_ROOT)/kind-kubeconfig"; \
+		cleanup_kind=true; \
+	fi; \
+	set -x; \
+	I2GW_BINARY_PATH=$(REPO_ROOT)/ingress2gateway KUBECONFIG=$${kubeconfig} \
+		go test $(I2GW_GO_TEST_ARGS) $(REPO_ROOT)/e2e; \
+	test_exit_code=$$?; \
+	set +x; \
+	if [ "$${cleanup_kind}" = "true" ] && [ "$${SKIP_CLEANUP}" != "1" ]; then \
```
bexxmodd, line 140: What if tests fails halfway through, will it leave some zombie resources? Maybe add `t.Cleanup` in Go test, also?

e2e/clients.go
johananl: Note that these settings affect only the k8s client. It's not a hard concurrency limit for the tests themselves. Does it still make sense to expose these as env vars?

Makefile
johananl: We're using `t.Cleanup()` in the Go code -- see `testutil.go`. I didn't spend a lot of time to make it airtight but in most cases we clean up resources by default, including in partial runs.

bexxmodd: SG. testutils was not loading for me :(

e2e/clients.go
bexxmodd: That's fine. We can always make that change later if needed.

Stevenjin8:
holding this until monday in case anyone wants to take a look

johananl:
Fixing the failure.

PR #296, opened 2026-01-07
Provides Clean Separation of Provider and Emitter IRs

### LINKED ISSUE #295: ResourcesToIRConverter Interface Should Return Provider IR
**What happened**:

`ResourcesToIRConverter()` returns an Emitter IR when it should instead return a Provider IR. `ToGatewayAPIResources()` should then iterate through each provider and do the following:

1. Call `ToIR()` to convert the source, e.g. Ingress with ingress-nginx annotations, to the provider IR.
2. Append any conversion errors to the error list.
3. Use the returned Provider IR when calling `ToEmitterIR()` to emit the Emitter IR.
4. Have the common Emitter emit the Emitter IR and append any conversion errors.
5. Pass the Emitter IR to `Emit()` generate the Gateway API resources.

`ResourcesToIRConverter()` to return the provider IR.

N/A

This bug is related to the implementation of `docs/emitters.md`.

danehans:
**What type of PR is this?**

/kind bug

Provides clean separation of Provider and Emitter IRs.

Fixes #295 

```release-note
NONE
```


danehans:
Closing due to https://github.com/kubernetes-sigs/ingress2gateway/issues/295#issuecomment-3739837989

PR #299, opened 2026-01-09
Ingress NGINX - Support Redirect

### LINKED ISSUE #268: [Ingress NGINX] Support Redirect
Support these annotations

```
nginx.ingress.kubernetes.io/force-ssl-redirect
nginx.ingress.kubernetes.io/ssl-redirect
nginx.ingress.kubernetes.io/permanent-redirect
nginx.ingress.kubernetes.io/permanent-redirect-code
nginx.ingress.kubernetes.io/temporal-redirect
nginx.ingress.kubernetes.io/temporal-redirect-code
nginx.ingress.kubernetes.io/from-to-www-redirect
nginx.ingress.kubernetes.io/proxy-redirect-from
nginx.ingress.kubernetes.io/proxy-redirect-to
```

or log errors if usable. See https://gateway-api.sigs.k8s.io/reference/spec/#httprequestredirectfilter

jgreeer:
**What type of PR is this?**

Fixes #268 

```release-note
Add support for `nginx.ingress.kubernetes.io/permanent-redirect` and `nginx.ingress.kubernetes.io/temporal-redirect` + warnings for `nginx.ingress.kubernetes.io/proxy-redirect-from` and `nginx.ingress.kubernetes.io/proxy-redirect-to`
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/redirect.go
+++ b/pkg/i2gw/providers/ingressnginx/redirect.go
@@ -34,8 +34,14 @@ import (
 )
 
 // redirectFeature converts permanent and temporal redirect annotations to Gateway API RequestRedirect filters.
-// - permanent-redirect uses a 301 status code
-// - temporal-redirect uses a 302 status code
+// This matches ingress-nginx's execution order: temporal redirect is checked first, then permanent.
+// If the temporal-redirect annotation key is present (even with an empty value), the function
+// short-circuits and permanent redirect annotations are never evaluated.
+//
+// Gateway API only supports status codes 301, 302, 303, 307, 308.
+// Intersecting with ingress-nginx's valid ranges:
+// - temporal-redirect defaults to 302, supported custom codes: 301, 302, 303, 307
+// - permanent-redirect defaults to 301, supported custom codes: 301, 302, 303, 307, 308
 func redirectFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, ir *providerir.ProviderIR) field.ErrorList {
 	var errs field.ErrorList
 
@@ -65,16 +71,18 @@ func redirectFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedNam
 					ingress.Namespace, ingress.Name, ProxyRedirectToAnnotation), ingress)
 			}
 
-			permanentRedirectURL, hasPermanent := ingress.Annotations[PermanentRedirectAnnotation]
 			temporalRedirectURL, hasTemporal := ingress.Annotations[TemporalRedirectAnnotation]
+			permanentRedirectURL, hasPermanent := ingress.Annotations[PermanentRedirectAnnotation]
 
 			// Skip if neither annotation is present
 			if !hasPermanent && !hasTemporal {
 				continue
 			}
 
-			// Determine redirect URL and status code
-			// If both are present, temporal takes priority
+			// Determine redirect URL and status code.
+			// Matching ingress-nginx execution order: temporal is checked first.
+			// If the temporal-redirect annotation key is present (even with an empty value),
+			// the function short-circuits — permanent redirect is never evaluated.
 			var redirectURL string
 			var statusCode int
 			var annotationUsed string
@@ -84,34 +92,37 @@ func redirectFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedNam
 				statusCode = 302
 				annotationUsed = TemporalRedirectAnnotation
 
-				// Warn if both annotations are present
+				// Warn if both annotations are present (permanent is ignored)
 				if hasPermanent {
-					notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s has both %s and %s annotations, using %s",
-						ingress.Namespace, ingress.Name, PermanentRedirectAnnotation, TemporalRedirectAnnotation, TemporalRedirectAnnotation), ingress)
+					notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s has both %s and %s annotations; temporal-redirect takes priority, permanent-redirect is ignored",
+						ingress.Namespace, ingress.Name, PermanentRedirectAnnotation, TemporalRedirectAnnotation), ingress)
 				}
 
-				// Check custom status code annotation
+				// Check custom status code annotation.
 				if codeStr := ingress.Annotations[TemporalRedirectCodeAnnotation]; codeStr != "" {
 					code, err := strconv.Atoi(codeStr)
-					if err != nil || code != 302 {
-						notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s uses unsupported status code %q in %s annotation, using default 302",
+					if err != nil || !isValidTemporalRedirectCode(code) {
+						notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s uses unsupported status code %q in %s annotation (Gateway API supports: 301, 302, 303, 307 for temporal redirects), using default 302",
 							ingress.Namespace, ingress.Name, codeStr, TemporalRedirectCodeAnnotation), ingress)
+					} else {
+						statusCode = code
 					}
-					// If code is 302, we allow it (it matches the default, so no change needed)
 				}
 			} else {
+				// Only reached if temporal-redirect annotation is completely absent
 				redirectURL = permanentRedirectURL
 				statusCode = 301
 				annotationUsed = PermanentRedirectAnnotation
 
-				// Check custom status code annotation
+				// Check custom status code annotation.
 				if codeStr := ingress.Annotations[PermanentRedirectCodeAnnotation]; codeStr != "" {
 					code, err := strconv.Atoi(codeStr)
-					if err != nil || code != 301 {
-						notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s uses unsupported status code %q in %s annotation, using default 301",
+					if err != nil || !isValidPermanentRedirectCode(code) {
+						notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s uses unsupported status code %q in %s annotation (Gateway API supports: 301, 302, 303, 307, 308 for permanent redirects), using default 301",
 							ingress.Namespace, ingress.Name, codeStr, PermanentRedirectCodeAnnotation), ingress)
+					} else {
+						statusCode = code
 					}
-					// If code is 301, we allow it (it matches the default, so no change needed)
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


jgreeer:
`nginx.ingress.kubernetes.io/from-to-www-redirect`, `nginx.ingress.kubernetes.io/permanent-redirect-code`, and `nginx.ingress.kubernetes.io/temporal-redirect-code` require 307/308 status code support, so they cannot be implemented until v1.5 release.

Stevenjin8:
tbh gwapi 1.5 is just around the corner. I think we can merge this as is. WDYT @mikemorris

jgreeer:
> tbh gwapi 1.5 is just around the corner. I think we can merge this as is. WDYT @mikemorris

this doesn't have the 307/308 status codes, it emits warnings for them. so we can merge now and then make another PR later for the status code annotations.

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
Stevenjin8:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +34,166 @@ import (
...
+			}
+
+			// Both annotations present is an error
+			if hasPermanent && hasTemporal {
+				errs = append(errs, field.Invalid(
+					field.NewPath("ingress", rule.Ingress.Namespace, rule.Ingress.Name, "metadata", "annotations"),
+					rule.Ingress.Annotations,
+					fmt.Sprintf("cannot use both %s and %s annotations simultaneously", PermanentRedirectAnnotation, TemporalRedirectAnnotation),
+				))
+				continue
```
Stevenjin8: Is this what ingress nginx does?

pkg/i2gw/providers/ingressnginx/annotations.go
```diff
@@ -21,10 +21,23 @@ const (
 	CanaryAnnotation            = "nginx.ingress.kubernetes.io/canary"
 	CanaryWeightAnnotation      = "nginx.ingress.kubernetes.io/canary-weight"
 	CanaryWeightTotalAnnotation = "nginx.ingress.kubernetes.io/canary-weight-total"
+	CanaryByHeader              = "nginx.ingress.kubernetes.io/canary-by-header"
+	CanaryByHeaderValue         = "nginx.ingress.kubernetes.io/canary-by-header-value"
+	CanaryByHeaderPattern       = "nginx.ingress.kubernetes.io/canary-by-header-pattern"
+	CanaryByCookie              = "nginx.ingress.kubernetes.io/canary-by-cookie"
```
Stevenjin8, line 27:
```suggestion
```

pkg/i2gw/providers/ingressnginx/redirect.go
jgreeer: i made an incorrect assumption here. when both are used it gives priority to the temporal redirect, will add this

pkg/i2gw/providers/ingressnginx/redirect_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright 2023 The Kubernetes Authors.
```
Stevenjin8: this one too

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +34,163 @@ import (
...
+
+				// Set hostname if present
+				if parsedURL.Hostname() != "" {
+					hostname := gatewayv1.PreciseHostname(parsedURL.Hostname())
+					redirectFilterConfig.Hostname = &hostname
+				}
+
+				// Set port if present
+				if parsedURL.Port() != "" {
+					port, err := strconv.Atoi(parsedURL.Port())
```
bexxmodd: Why are we ignoring this error? Let's append to errs.

Stevenjin8: This did get me thinking... we could have named ports here right?

jgreeer: this does gets caught by `url.Parse()` before we reach this, but i added a check to be safe. also not sure what you mean steven?

Stevenjin8: @jgreeer i just reread this and what I said makes no sense

Stevenjin8: @jgreeer Atoi can only really fail if the port overflows int, which shouldn't be possible unless you are actively trying to mess up. This seems fine.

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +34,170 @@ import (
...
+					// Clear backend refs as redirects don't route to backends
+					BackendRefs: nil,
+				}
+
+				// Prepend the redirect rule to existing rules
+				httpRouteContext.HTTPRoute.Spec.Rules = append(
+					[]gatewayv1.HTTPRouteRule{redirectRule},
+					httpRouteContext.HTTPRoute.Spec.Rules...,
+				)
+
```
kkk777-7, line 204: should we add BackendSource to adjust rule index?

`RuleBackendSources[i][j] is the source of the jth backend in the ith element of HTTPRoute.Spec.Rules.`
https://github.com/kubernetes-sigs/ingress2gateway/blob/main/pkg/i2gw/provider_intermediate/intermediate_representation.go#L70-L71

pkg/i2gw/providers/ingressnginx/redirect_test.go
kkk777-7, line 1: nits:
current tests look removed. e.g. `TestAddDefaultSSLRedirect_disabledByAnnotation`
could you rebase or merge main branch?

pkg/i2gw/providers/ingressnginx/redirect.go
jgreeer: right, fixed this by adding the rule in backendsources but with no backends since it just redirects

Stevenjin8: the semantics are a bit weird on this but this sounds reasonable to me

jgreeer: actually, couldn't you just add the requestRedirect to the existing rule? that might be better

jgreeer: this breaks when you have a path that redirects and a path that doesn't, trying to fix now

jgreeer:
Changed approach to the following:
- Iterate through all rules in the IR
- Find a rule that has a source ingress with a redirect annotation
- Add the redirect filter to that rule using values from URL.parse()
- Remove backendRef from that rule (yaml isn't accepted if we don't do this)
- Keep existing backendSource since it still contributed that rule

Also, always override the path with "/" if no path is provided in the redirect annotation. Gateway api will keep the path from the match in the redirect, ingress-nginx does not.

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +33,177 @@ import (
...
+	// Iterate over all HTTPRoutes in the IR
+	for key, httpRouteContext := range ir.HTTPRoutes {
+		// Iterate over each rule in the HTTPRoute
+		for ruleIndex := range httpRouteContext.HTTPRoute.Spec.Rules {
+			// Check if this rule has backend sources
+			if ruleIndex >= len(httpRouteContext.RuleBackendSources) {
+				continue
+			}
+
+			// Get ingress from the first backend source (all backends in a rule should come from the same ingress)
```
Stevenjin8: I don thtink this is true. What about canary

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +33,177 @@ import (
...
+				annotationUsed = TemporalRedirectAnnotation
+
+				// Warn if both annotations are present
+				if hasPermanent {
+					notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s has both %s and %s annotations, using %s",
+						ingress.Namespace, ingress.Name, PermanentRedirectAnnotation, TemporalRedirectAnnotation, TemporalRedirectAnnotation), ingress)
+				}
+
+				// Warn about unsupported custom status code annotation
+				if ingress.Annotations[TemporalRedirectCodeAnnotation] != "" {
```
Stevenjin8: If its the TemporalRedirectCodeAnnotation is 301, its ok though. Also isn't the default 308 anyways? so we need to make a warning anyways.

Stevenjin8: *what if the TemporalRedirectCodeAnnotation is already 301

jgreeer: Default codes are 301/302. I will add a check for if the redirect code annotation is already a default though

pkg/i2gw/providers/ingressnginx/redirect.go
jgreeer: It's not, but when you combine canary and redirect the redirect takes priority and no canary happens. Are there other situations where we might need to check other backend sources?

Stevenjin8: I don't think so.

Stevenjin8: @jgreeer can you fix

jgreeer: Ah, I thought you were agreeing. So fix to account for when canary is enabled? What would that look like?

Stevenjin8: Oh I was thinking of using'getNonCanaryIngress'

Stevenjin8:
@kkk777-7 @spencerhance @bexxmodd  could one of you take a look or lgtm this.

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +33,176 @@ import (
...
+				// Warn if both annotations are present
+				if hasPermanent {
+					notify(notifications.WarningNotification, fmt.Sprintf("ingress %s/%s has both %s and %s annotations, using %s",
+						ingress.Namespace, ingress.Name, PermanentRedirectAnnotation, TemporalRedirectAnnotation, TemporalRedirectAnnotation), ingress)
+				}
+
+				// Check custom status code annotation
+				if codeStr := ingress.Annotations[TemporalRedirectCodeAnnotation]; codeStr != "" {
+					code, err := strconv.Atoi(codeStr)
+					if err != nil || code != 302 {
```
bexxmodd: Isn't HTTPRequestedRedirectFilter supports 303, 307, and 308 now?

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -30,6 +33,176 @@ import (
...
+			// Check if this rule has backend sources
+			if ruleIndex >= len(httpRouteContext.RuleBackendSources) {
+				continue
+			}
+
+			// Get the non canary ingress for this rule
+			ingress := getNonCanaryIngress(httpRouteContext.RuleBackendSources[ruleIndex])
+
+			// Warn about unsupported proxy-redirect annotations
+			if ingress.Annotations[ProxyRedirectFromAnnotation] != "" {
```
bexxmodd, line 65: Don't we need to check `ingress != nil` ?

pkg/i2gw/providers/ingressnginx/redirect.go
jgreeer: yeah it's about to release with v1.5, i'll go ahead and add it to this PR now

pkg/i2gw/providers/ingressnginx/redirect.go
jgreeer: yes, you're right

PR #302, opened 2026-01-13
fix: setting proper secret group and kind in certificateRefs

cnvergence:
**What type of PR is this?**

/kind bug

This PR fixes generating null values for group and kind for SecretObjectReference in CertificateRefs.
I have migrated resources on my cluster and did notice certificateRefs being translated to this.
```yaml
      certificateRefs:
      - group: null
        kind: null
        name: tls-sectet
```
I know that there are defaults set in the gateway-api, but it is still a bit confusing, and I think it should not rely fully on defaulting.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/i2gw/providers/common/converter.go
+++ b/pkg/i2gw/providers/common/converter.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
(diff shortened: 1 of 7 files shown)


Stevenjin8:
@cnvergence Also, could you add a quick release note?

pkg/i2gw/providers/common/converter.go
```diff
@@ -229,7 +230,11 @@ func (a *ingressAggregator) toHTTPRoutesAndGateways(options i2gw.ProviderImpleme
 		}
 		for _, tls := range rg.tls {
 			listener.TLS.CertificateRefs = append(listener.TLS.CertificateRefs,
-				gatewayv1.SecretObjectReference{Name: gatewayv1.ObjectName(tls.SecretName)})
+				gatewayv1.SecretObjectReference{
+					Group: ptr.To(gatewayv1.Group("")),
```
youngnick, line 234: It's fine for now, but we should also consider using the `corev1.GroupName` constant instead?

PR #303, opened 2026-01-16
feat: Implement standard CORS support

chakravardhan:
**What type of PR is this?**

/kind feature

This is the first PR in a stacked series. This PR focuses on the internal infrastructure and providing robust, standard-compliant CORS support. It enables users to migrate complex CORS configurations seamlessly to `HTTPRoute` filters.
### Supported Annotations & Mappings
This PR adds a conversion mapping for the following Nginx annotations to Gateway API `HTTPRouteFilterCORS` fields:
| Nginx Annotation | Gateway API Field (`HTTPRoute.filters.cors`) | Default Behavior (if unset) |
| :--- | :--- | :--- |
| `nginx.ingress.kubernetes.io/enable-cors` | (Enables Filter) | Must be `"true"` to enable. |
| `cors-allow-origin` | `allowOrigins` | `*` |
| `cors-allow-methods` | `allowMethods` | `GET, PUT, POST, DELETE, PATCH, OPTIONS` |
| `cors-allow-headers` | `allowHeaders` | Nginx Standard List (DNT, Keep-Alive, User-Agent, etc.) |
| `cors-expose-headers` | `exposeHeaders` | (Empty) |
| `cors-allow-credentials` | `allowCredentials` | `true` (Matches Nginx default) |
| `cors-max-age` | `maxAge` | `1728000` (Matches Nginx default) |

### New flags - 

**Flag**:  --allow-experimental-gw-api
**Purpose**: Controls whether Experimental Gateway API fields are included in the output.
**Behavior**: Defaults to false. Must be explicitly set to generated CORS configuration, as it is considered an advanced/experimental feature in this tool context.

Fixes #

```release-note
Implemented support for Nginx CORS annotations (enable-cors, allow-origin, allow-methods, allow-headers, expose-headers, allow-credentials, max-age) mapping to Gateway API CORS filters with strict Nginx default compliance.
```

### DIFF
--- a/pkg/i2gw/emitter_intermediate/intermediate_representation.go
+++ b/pkg/i2gw/emitter_intermediate/intermediate_representation.go
@@ -17,12 +17,12 @@ limitations under the License.
 package emitterir
 
 import (
-        "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate/gce"
-        "k8s.io/apimachinery/pkg/api/resource"
-        "k8s.io/apimachinery/pkg/types"
-        gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
-        gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
-        gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate/gce"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/types"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
 // EmitterIR holds specifications of Gateway Objects for supporting Ingress extensions,
@@ -30,84 +30,84 @@ import (
 // features. An EmitterIR field can be mapped to core Gateway-API fields,
 // or provider-specific Gateway extensions.
 type EmitterIR struct {
-        Gateways   map[types.NamespacedName]GatewayContext
-        HTTPRoutes map[types.NamespacedName]HTTPRouteContext
+	Gateways   map[types.NamespacedName]GatewayContext
+	HTTPRoutes map[types.NamespacedName]HTTPRouteContext
 
-        GatewayClasses map[types.NamespacedName]GatewayClassContext
-        TLSRoutes      map[types.NamespacedName]TLSRouteContext
-        TCPRoutes      map[types.NamespacedName]TCPRouteContext
-        UDPRoutes      map[types.NamespacedName]UDPRouteContext
-        GRPCRoutes     map[types.NamespacedName]GRPCRouteContext
+	GatewayClasses map[types.NamespacedName]GatewayClassContext
+	TLSRoutes      map[types.NamespacedName]TLSRouteContext
+	TCPRoutes      map[types.NamespacedName]TCPRouteContext
+	UDPRoutes      map[types.NamespacedName]UDPRouteContext
+	GRPCRoutes     map[types.NamespacedName]GRPCRouteContext
 
-        BackendTLSPolicies map[types.NamespacedName]BackendTLSPolicyContext
-        ReferenceGrants    map[types.NamespacedName]ReferenceGrantContext
+	BackendTLSPolicies map[types.NamespacedName]BackendTLSPolicyContext
+	ReferenceGrants    map[types.NamespacedName]ReferenceGrantContext
 
-        GceServices map[types.NamespacedName]gce.ServiceIR
+	GceServices map[types.NamespacedName]gce.ServiceIR
 }
 
 type GatewayContext struct {
-        gatewayv1.Gateway
-        // Emitter IR should be provider/emitter neutral,
-        // But we have GCE for backcompatibility.
-        Gce *gce.GatewayIR
+	gatewayv1.Gateway
+	// Emitter IR should be provider/emitter neutral,
+	// But we have GCE for backcompatibility.
+	Gce *gce.GatewayIR
 }
 
 type HTTPRouteContext struct {
-        gatewayv1.HTTPRoute
-        // PathRewriteByRuleIdx maps HTTPRoute rule indices to path rewrite intent.
-        // This is provider-neutral and applied by the common emitter.
-        PathRewriteByRuleIdx map[int]*PathRewrite
+	gatewayv1.HTTPRoute
+	// PathRewriteByRuleIdx maps HTTPRoute rule indices to path rewrite intent.
+	// This is provider-neutral and applied by the common emitter.
+	PathRewriteByRuleIdx map[int]*PathRewrite
 
-        // BodySizeByRuleIdx maps HTTPRoute rule indices to body size intent.
-        // This is provider-neutral and applied by each custom emitter.
-        BodySizeByRuleIdx map[int]*BodySize
+	// BodySizeByRuleIdx maps HTTPRoute rule indices to body size intent.
+	// This is provider-neutral and applied by each custom emitter.
+	BodySizeByRuleIdx map[int]*BodySize
 
-        // CorsPolicyByRuleIdx maps HTTPRoute rule indices to CORS policy intent.
-        // This map is populated by providers that support CORS (e.g., via annotations) and is
-        // applied by the CommonEmitter. This separation allows the CORS logic to be provider-neutral
-        // and consistently applied across different providers, subject to feature gating.
-        CorsPolicyByRuleIdx map[int]*gatewayv1.HTTPCORSFilter
+	// CorsPolicyByRuleIdx maps HTTPRoute rule indices to CORS policy intent.
+	// This map is populated by providers that support CORS (e.g., via annotations) and is
+	// applied by the CommonEmitter. This separation allows the CORS logic to be provider-neutral
+	// and consistently applied across different providers, subject to feature gating.
+	CorsPolicyByRuleIdx map[int]*gatewayv1.HTTPCORSFilter
 }
 
 // PathRewrite represents provider-neutral path rewrite intent.
 // For now it only supports full-path replacement; more fields may be added later.
 type PathRewrite struct {
-        ReplaceFullPath string
-        // Headers to add on path rewrite.
-        Headers map[string]string
-        Regex   bool
+	ReplaceFullPath string
+	// Headers to add on path rewrite.
+	Headers map[string]string
+	Regex   bool
 }
 
 // BodySize represents provider-neutral body size intent.
 type BodySize struct {
-        BufferSize *resource.Quantity
-        MaxSize    *resource.Quantity
+	BufferSize *resource.Quantity
+	MaxSize    *resource.Quantity
 }
 
 type GatewayClassContext struct {
-        gatewayv1.GatewayClass
+	gatewayv1.GatewayClass
 }
 
 type TLSRouteContext struct {
-        gatewayv1alpha2.TLSRoute
+	gatewayv1alpha2.TLSRoute
 }
 
 type TCPRouteContext struct {
-        gatewayv1alpha2.TCPRoute
+	gatewayv1alpha2.TCPRoute
 }
 
 type UDPRouteContext struct {
-        gatewayv1alpha2.UDPRoute
+	gatewayv1alpha2.UDPRoute
 }
 
 type GRPCRouteContext struct {
-        gatewayv1.GRPCRoute
+	gatewayv1.GRPCRoute
 }
 
 type BackendTLSPolicyContext struct {
-        gatewayv1.BackendTLSPolicy
+	gatewayv1.BackendTLSPolicy
 }
 
 type ReferenceGrantContext struct {
-        gatewayv1beta1.ReferenceGrant
+	gatewayv1beta1.ReferenceGrant
 }
--- a/pkg/i2gw/providers/ingressnginx/cors.go
+++ b/pkg/i2gw/providers/ingressnginx/cors.go
@@ -137,8 +137,8 @@ func applyCorsToEmitterIR(pIR providerir.ProviderIR, eIR *emitterir.EmitterIR) f
 			// Nginx default is 1728000.
 			var maxAgeVal int32 = 1728000 // Default from Nginx
 			if maxAgeStr, ok := ing.Annotations[CorsMaxAgeAnnotation]; ok && maxAgeStr != "" {
-				// Try parsing as integer (seconds)
-				if val, err := strconv.Atoi(maxAgeStr); err == nil {
+				// Try parsing as integer (seconds) using ParseInt with bitSize 32
+				if val, err := strconv.ParseInt(maxAgeStr, 10, 32); err == nil {
 					maxAgeVal = int32(val)
 				} else {
 					errs = append(errs, field.Invalid(field.NewPath("metadata", "annotations", CorsMaxAgeAnnotation), maxAgeStr, "invalid cors-max-age value"))
--- a/pkg/i2gw/providers/ingressnginx/annotations.go
+++ b/pkg/i2gw/providers/ingressnginx/annotations.go
@@ -42,10 +42,11 @@ const (
 	SSLRedirectAnnotation = "nginx.ingress.kubernetes.io/ssl-redirect"
 
 	// CORS annotations
-	EnableCorsAnnotation           = "nginx.ingress.kubernetes.io/enable-cors"
-	CorsAllowOriginAnnotation      = "nginx.ingress.kubernetes.io/cors-allow-origin"
-	CorsAllowHeadersAnnotation     = "nginx.ingress.kubernetes.io/cors-allow-headers"
-	CorsAllowMethodsAnnotation     = "nginx.ingress.kubernetes.io/cors-allow-methods"
+	EnableCorsAnnotation       = "nginx.ingress.kubernetes.io/enable-cors"
+	CorsAllowOriginAnnotation  = "nginx.ingress.kubernetes.io/cors-allow-origin"
+	CorsAllowHeadersAnnotation = "nginx.ingress.kubernetes.io/cors-allow-headers"
+	CorsAllowMethodsAnnotation = "nginx.ingress.kubernetes.io/cors-allow-methods"
+	//nolint:gosec // false positive, this is an annotation key
 	CorsAllowCredentialsAnnotation = "nginx.ingress.kubernetes.io/cors-allow-credentials"
 	CorsExposeHeadersAnnotation    = "nginx.ingress.kubernetes.io/cors-expose-headers"
 	CorsMaxAgeAnnotation           = "nginx.ingress.kubernetes.io/cors-max-age"


pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -53,6 +53,7 @@ type GatewayContext struct {
 
 type HTTPRouteContext struct {
 	gatewayv1.HTTPRoute
+	Gce *gce.HTTPRouteIR
```
Stevenjin8:
```suggestion
```

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -68,9 +68,19 @@ func (c *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.Err
 	}
 	buildGceGatewayExtensions(ir, &gatewayResources)
 	buildGceServiceExtensions(ir, &gatewayResources)
+	updateGatewayClass(&gatewayResources)
 	return gatewayResources, nil
 }
 
+func updateGatewayClass(gatewayResources *i2gw.GatewayResources) {
+	for i, gw := range gatewayResources.Gateways {
+		if gw.Spec.GatewayClassName == "nginx" {
```
Stevenjin8: nginx allows for customizing gateway class name. Do you have to make this check?

pkg/i2gw/providers/ingressnginx/gce_converter.go
```diff
@@ -0,0 +1,31 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func gceFeature(_ []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, ir *providerir.ProviderIR) field.ErrorList {
+	// gceFeature is a hook for GCE-specific logic.
+	// Currently used as an integration point for future GCE capabilities.
+	return nil
+}
+
```
Stevenjin8: I could probably have documented this better, but providers should not have any emitter specific features.

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,158 @@
...
+
+	providerir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/provider_intermediate"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/utils/ptr"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+func corsFeature(_ []networkingv1.Ingress, _ map[types.NamespacedName]map[string]int32, ir *providerir.ProviderIR) field.ErrorList {
```
Stevenjin8: Sorry, there isn't a lot of precedent for this and I haven't had time to update the docs, but the Cors filter is currently experimental (https://gateway-api.sigs.k8s.io/geps/gep-1767/), so it seems a bit early to add it to ingress2gateway.

Stevenjin8: We could implement it now, but the flow would look a little different. You would follow the flow of something like https://github.com/kubernetes-sigs/ingress2gateway/pull/288 where we create the IR in the provider and then populate the fields in the common emitter. We would also have to add a feature flag/argument to the common emitter `--allow-alpha-gw-api`

Stevenjin8: Also, take a look at this PR: https://github.com/kubernetes-sigs/ingress2gateway/pull/305 that has the structure I'm looking for. Also, I'm going to ask you to sign yourself up as a codeowner of the GCE emitter if you want to add code there.

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,158 @@
...
+			// Nginx default is true. Only false if explicitly set to "false".
+			if creds, ok := ingress.Annotations[CorsAllowCredentialsAnnotation]; ok && creds == "false" {
+				corsFilter.AllowCredentials = ptr.To(false)
+			} else {
+				corsFilter.AllowCredentials = ptr.To(true)
+			}
+
+			// Max Age
+			// Nginx default is 1728000.
+			var maxAgeVal int32 = 1728000 // Default from Nginx
```
bexxmodd, line 138: I understand that nginx has this number set for max age, but do we want to just it, or let GW API controller handle its own defaults? The reason I'm asking is because CORS filter has 5 seconds max age as a default value, which is quite a mismatch from this

https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/httproute_types.go#L1576-L1582

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -68,9 +68,19 @@ func (c *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.Err
...
+			gw.Spec.GatewayClassName = "gke-l7-global-external-managed"
```
bexxmodd: Shouldn't we allow for this to be configurable?

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,158 @@
...
+			if maxAgeStr, ok := ingress.Annotations[CorsMaxAgeAnnotation]; ok && maxAgeStr != "" {
+				if val, err := strconv.Atoi(maxAgeStr); err == nil {
```
bexxmodd: Let's say if use sets "10s" instead of 10, this will silently fail. Maybe add err to ErrorList?

pkg/i2gw/providers/ingressnginx/gce_converter.go
chakravardhan: Ack. Removing this file entirely to address the architectural concern about emitter-specific features in providers.

chakravardhan: We can revisit where to place GCE-specific transformation logic (like specific policy generation) in future PRs, likely within the GCE emitter itself

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: I have implemented the `--allow-alpha-gw-api` flag in the root command and wired it through to the `CommonEmitter`.

By default (flag false), the `CommonEmitter` now actively filters out experimental features like CORS filters from the IR before it reaches the specific emitter.

When `--allow-alpha-gw-api is` passed, CORS filters are preserved and output.

chakravardhan: Ack. I will check PR #305 and follow up with a separate PR to add myself to the OWNERS file for the GCE emitter components once this infrastructure lands.

pkg/i2gw/emitters/gce/gce.go
chakravardhan: I removed the check for nginx class name. The GCE emitter now consistently sets the GatewayClassName to `gke-l7-global-external-managed` for all generated gateways, regardless of the input IngressClass.

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: Refactored the code to return the error

pkg/i2gw/emitters/gce/gce.go
chakravardhan: Ack. I've resolved this by making the `GatewayClassName` configurable via a new CLI flag `--gce-gateway-class-name`. It defaults to `gke-l7-global-external-managed` to preserve the existing behavior, but now users can override it

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: To ensure a safe migration and avoid a sudden spike in preflight requests (going from 20 days cache to 5 seconds), we've explicitly set it to 1728000 because that is the default behavior in Nginx Ingress when CORS is enabled. The Gateway API default (5s) is significantly lower. Wouldn't it be safer to explicitly preserve the 20-day behavior from Nginx?

Stevenjin8: yep, when I looked at it, there were slightly different defaults. We want to use the ignress nginx ones.

pkg/i2gw/providers/ingressnginx/cors.go
Stevenjin8: This isn't a blocker, but right now, we propagate errors in two different ways, `err` and `notify`. We should decide which one to use/when to use each one.

cmd/print.go
```diff
@@ -80,6 +80,9 @@ type PrintRunner struct {
 	// emitter indicates which emitter is used to generate the Gateway API resources.
 	// Defaults to "standard".
 	emitter string
+
+	// allowAlphaGatewayAPI indicates whether Experimental Gateway API features (like CORS, URLRewrite) should be included in the output.
+	allowAlphaGatewayAPI bool
```
bexxmodd: Given that we split the GW Api fetures as standard and experimental. Let's rename it to `allowExperimentalGatewayAPI`

bexxmodd (approved):
Left one comment regarding flag value. The rest LGTM

cmd/print.go
```diff
@@ -362,6 +365,8 @@ if specified with --namespace.`)
 	cmd.Flags().StringSliceVar(&pr.providers, "providers", []string{},
 		fmt.Sprintf("If present, the tool will try to convert only resources related to the specified providers, supported values are %v.", i2gw.GetSupportedProviders()))
 
+	cmd.Flags().BoolVar(&pr.allowAlphaGatewayAPI, "allow-alpha-gw-api", false, "If present, the tool will include Alpha/Experimental Gateway API fields (e.g. CORS, URLRewrite) in the output. Default is false.")
```
Stevenjin8:
```suggestion
	cmd.Flags().BoolVar(&pr.allowAlphaGatewayAPI, "allow-alpha-gw-api", false, "If present, the tool will include Experimental Gateway API fields (e.g. CORS, URLRewrite) in the output. Default is false.")
```

pkg/i2gw/emitter.go
```diff
@@ -60,5 +60,9 @@ type EmitterName string
 type EmitterConstructor func(conf *EmitterConf) Emitter
 
 type EmitterConf struct {
-	// TODO: add fields as needed.
+	// AllowAlphaGatewayAPI indicates whether Experimental Gateway API features (like CORS, URLRewrite) should be included in the output.
+	AllowAlphaGatewayAPI bool
+	// ProviderSpecificFlags contains the flags that were passed to the provider
+	ProviderSpecificFlags map[string]map[string]string
```
Stevenjin8: We shouldn't have any ProverSpecificFlags in Emitter.

pkg/i2gw/emitters/common_emitter/emitter.go
```diff
@@ -22,10 +22,12 @@ import (
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
-type Emitter struct{}
+type Emitter struct {
+	allowAlpha bool
```
Stevenjin8:
```suggestion
	allowExperimental bool
```

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -68,9 +78,29 @@ func (c *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.Err
 	}
 	buildGceGatewayExtensions(ir, &gatewayResources)
 	buildGceServiceExtensions(ir, &gatewayResources)
+	c.updateGatewayClass(&gatewayResources)
 	return gatewayResources, nil
 }
 
+func (c *Emitter) updateGatewayClass(gatewayResources *i2gw.GatewayResources) {
```
Stevenjin8: can we have all GCE stuff in another PR

pkg/i2gw/providers/ingressnginx/cors.go
Stevenjin8: @chakravardhan could you look at #305? The flow right now is add the CORSFilter in the provider and strip it out in the CommonEmitter before it gets sent to the final emitter. But what if the final emitter is Envoy Gateway and they want to implement it using [their CORS filter](https://gateway.envoyproxy.io/docs/tasks/security/cors/). Also, structuring it like #305 will help us in the future because we will know what was parsed and what wasn't.

pkg/i2gw/emitters/gce/gce.go
spencerhance: +1 - mentioned this offline as well

cmd/print.go
```diff
@@ -391,6 +396,13 @@ func (pr *PrintRunner) getProviderSpecificFlags() map[string]map[string]string {
 	providerSpecificFlags := make(map[string]map[string]string)
 	for flagName, value := range pr.providerSpecificFlags {
 		provider, found := lo.Find(pr.providers, func(p string) bool { return strings.HasPrefix(flagName, fmt.Sprintf("%s-", p)) })
+		if !found {
+			// If not found in providers, check if it matches the emitter
+			if strings.HasPrefix(flagName, fmt.Sprintf("%s-", pr.emitter)) {
+				provider = pr.emitter
+				found = true
+			}
+		}
```
spencerhance: I'm a bit confused by this part, but I might just be missing something.  Can you elaborate why this is necessary?  For example, I would think that if a GCE specific flag is used but the GCE provider isn't provided, it shouldn't work.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -53,10 +53,10 @@ type GatewayContext struct {
 
 type HTTPRouteContext struct {
 	gatewayv1.HTTPRoute
-
 	// PathRewriteByRuleIdx maps HTTPRoute rule indices to path rewrite intent.
 	// This is provider-neutral and applied by the common emitter.
 	PathRewriteByRuleIdx map[int]*PathRewrite
+
 }
```
spencerhance, line 70: nit: was this format change intentional?

cmd/print.go
chakravardhan: Agreed! Renamed

pkg/i2gw/emitter_intermediate/intermediate_representation.go
chakravardhan: fixed

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: Understood, Adopting a structure like #305. Let me know if this looks good

cmd/print.go
chakravardhan: removed this logic

pkg/i2gw/emitter.go
chakravardhan: agreed, removed

pkg/i2gw/emitters/gce/gce.go
chakravardhan: removed all GCE stuff and only added CORS related changes

pkg/i2gw/emitters/common_emitter/emitter.go
```diff
@@ -61,10 +61,27 @@ func applyPathRewrites(ir *emitterir.EmitterIR) {
...
+	for key, routeCtx := range ir.HTTPRoutes {
+		for ruleIdx, policy := range routeCtx.CorsPolicyByRuleIdx {
+			if policy == nil {
+				continue
+			}
+			routeCtx.Spec.Rules[ruleIdx].Filters = append(routeCtx.Spec.Rules[ruleIdx].Filters, gatewayv1.HTTPRouteFilter{
+				Type: gatewayv1.HTTPRouteFilterCORS,
+				CORS: policy,
+			})
+			// routeCtx.CorsPolicyByRuleIdx[ruleIdx] = nil -- REMOVED clearing to preserve intent
```
Stevenjin8:
```suggestion
			routeCtx.CorsPolicyByRuleIdx[ruleIdx] = nil -- REMOVED clearing to preserve intent
```

pkg/i2gw/emitters/standard/standard.go
```diff
@@ -21,21 +21,43 @@ import (
...
-	return &Emitter{}
+func NewEmitter(conf *i2gw.EmitterConf) i2gw.Emitter {
+	return &Emitter{conf: conf}
 }
 
 // Emit converts the provider intermediate representation to Gateway API resources.
 func (e *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.ErrorList) {
+	if e.conf != nil && !e.conf.AllowExperimentalGatewayAPI {
+		filterOutCORS(ir)
+	}
```
Stevenjin8: We don't need this anymore right

pkg/i2gw/provider_intermediate/conversion.go
```diff
@@ -35,10 +36,25 @@ func ToEmitterIR(pIR ProviderIR) emitterir.EmitterIR {
 	}
 
 	for k, v := range pIR.Gateways {
-		eIR.Gateways[k] = emitterir.GatewayContext{Gateway: v.Gateway}
+		ctx := emitterir.GatewayContext{Gateway: v.Gateway}
+		if v.ProviderSpecificIR.Gce != nil {
+			ctx.Gce = v.ProviderSpecificIR.Gce
+		}
+		eIR.Gateways[k] = ctx
```
Stevenjin8: can we get rid of GCE references in this PR

pkg/i2gw/provider_intermediate/intermediate_representation.go
```diff
@@ -69,6 +69,10 @@ type HTTPRouteContext struct {
 
 	// RuleBackendSources[i][j] is the source of the jth backend in the ith element of HTTPRoute.Spec.Rules.
 	RuleBackendSources [][]BackendSource
+
+	// CorsPolicyByRuleIdx maps HTTPRoute rule indices to CORS policy intent.
+	// This is provider-neutral and applied by the common emitter.
+	CorsPolicyByRuleIdx map[int]*gatewayv1.HTTPCORSFilter
```
Stevenjin8:
```suggestion
```

pkg/i2gw/providers/ingressnginx/cors.go
Stevenjin8: @chakravardhan. Thanks, this is almost what I'm looking for, but it would be ideal to not touch ProviderIR. See `applyBodySizeToEmitterIR` here https://github.com/kubernetes-sigs/ingress2gateway/pull/305/files

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,150 @@
...
+			sources := httpRouteContext.RuleBackendSources[i]
+
+			ingress := getNonCanaryIngress(sources)
+			if ingress == nil {
+				continue
+			}
+
+			// Check if CORS is enabled
+			enableCors, ok := ingress.Annotations[EnableCorsAnnotation]
+			if !ok || enableCors != "true" {
```
Stevenjin8: this can be any value that `strconv.ParseBool`s to `true`

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,150 @@
...
+				for _, h := range headerList {
+					h = strings.TrimSpace(h)
+					if h == "" {
+						continue
+					}
+					corsFilter.AllowHeaders = append(corsFilter.AllowHeaders, gatewayv1.HTTPHeaderName(h))
+				}
+			} else {
+				// Default headers from Nginx documentation
+				defaultHeaders := []string{"DNT", "Keep-Alive", "User-Agent", "X-Requested-With", "If-Modified-Since", "Cache-Control", "Content-Type", "Range", "Authorization"}
```
Stevenjin8, line 109: are these supposed to be appended? or are they the default value if headers are unspecified.

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,150 @@
...
+					h = strings.TrimSpace(h)
+					if h == "" {
+						continue
+					}
+					corsFilter.ExposeHeaders = append(corsFilter.ExposeHeaders, gatewayv1.HTTPHeaderName(h))
+				}
+			}
+
+			// Allow Credentials
+			// Nginx default is true. Only false if explicitly set to "false".
```
Stevenjin8: any string value that go can parse into a bool works.

pkg/i2gw/providers/ingressnginx/cors.go
```diff
@@ -0,0 +1,150 @@
...
+					maxAgeVal = int32(val)
+				} else {
+					// Try parsing as duration (e.g. "10s")
+					if d, err := time.ParseDuration(maxAgeStr); err == nil {
+						maxAgeVal = int32(d.Seconds())
+					} else {
+						errs = append(errs, field.Invalid(field.NewPath("metadata", "annotations", CorsMaxAgeAnnotation), maxAgeStr, "invalid cors-max-age value"))
+					}
+				}
+			}
```
Stevenjin8: I'm pretty sure ingress nginx doesn't support have `s` at the end.

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: Agreed, I've updated the logic to use `strconv.ParseBool` so it correctly handles all standard boolean string values

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: They are the default values. These defaults are only applied if the cors-allow-headers annotation is missing or empty.

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: agreed, updated

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: Removed

pkg/i2gw/provider_intermediate/conversion.go
chakravardhan: removed

pkg/i2gw/emitters/standard/standard.go
chakravardhan: yep, removed

pkg/i2gw/providers/ingressnginx/cors.go
chakravardhan: I've modified code similar to the mentioned PR above by adding this method `applyCorsToEmitterIR`. Kindly check this approach and let us know if this aligns well. Thanks! @Stevenjin8

chakravardhan:
> /approve

@Stevenjin8, would you be able to tag /ok-to-test for the pipeline to run?

Stevenjin8:
@chakravardhan lints are failing.

chakravardhan:
> @chakravardhan lints are failing.

Fixed now!

kkk777-7:
LGTM, thanks!

/lgtm

PR #298, opened 2026-01-09
feat: support route rule name

kkk777-7:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
support route rule name in output files.
this enable xPolicy CRD to attach to section name level.

**Which issue(s) this PR fixes**:

Fixes #297 

**Does this PR introduce a user-facing change?**:

```release-note
support route rule name in output.
```

PR #305, opened 2026-01-18
feat: buffer annotations and envoy gateway emitter

kkk777-7:
**What type of PR is this?**

/kind feature

~~depends on #298~~

Add Envoy Gateway emitter.
Support proxy body size annotations.
- `nginx.ingress.kubernetes.io/proxy-body-size`
- `nginx.ingress.kubernetes.io/client-body-buffer-size`

Fixes #

```release-note
Support for ingress nginx body size annotations (proxy-body-size, client-body-buffer-size). and Add Envoy Gateway emitter.
```


pkg/i2gw/emitters/envoygateway/emitter.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func NewEmitter(_ *i2gw.EmitterConf) i2gw.Emitter {
+	return &Emitter{
+		builderMap: NewBuilderMap(),
+	}
+}
+
+func (e *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.ErrorList) {
+	gatewayResources, errs := utils.ToGatewayResources(ir)
```
kkk777-7, line 45: I think that the custom emitter shouldn’t use `utils.ToGatewayResources` and should be handled by the common emitter.
if we finish implementing common emitter, I'll replace this line.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -53,6 +54,16 @@ type GatewayContext struct {
 
 type HTTPRouteContext struct {
 	gatewayv1.HTTPRoute
+
+	// BodySizeByRuleIdx maps HTTPRoute rule indices to body size intent.
+	// This is provider-neutral and applied by each custom emitter.
+	BodySizeByRuleIdx map[int]*BodySize
```
puertomontt, line 64: Shouldn't this be added to the providerIR instead?

kkk777-7: I think we should add this to emitterIR because each emitters should only be aware of the emitterIR.

kkk777-7:
@Stevenjin8 
I added CODEOWNERS. and agree to Governance :)

pkg/i2gw/providers/ingressnginx/bodysize.go
```diff
@@ -0,0 +1,196 @@
...
+	emitterir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
+	providerir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/provider_intermediate"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// Ref: https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/template/template.go#L1001
+var nginxSizeRegex = regexp.MustCompile(`^(\d+)([kKmM])?$`)
```
Stevenjin8: > var SizeRegex = regexp.MustCompile(`^(?i)\d+[bkmg]?$`)

https://github.com/kubernetes/ingress-nginx/blob/95febbcbced177df0a96a2bc678389d5ce03c25f/internal/ingress/annotations/parser/validators.go#L57

pkg/i2gw/providers/ingressnginx/bodysize.go
```diff
@@ -0,0 +1,196 @@
...
+
+			var (
+				maxSize          *resource.Quantity
+				bufferSize       *resource.Quantity
+				maxSizeSource    *networkingv1.Ingress
+				bufferSizeSource *networkingv1.Ingress
+			)
+
+			for _, source := range pRouteCtx.RuleBackendSources[ruleIdx] {
+				if source.Ingress.Annotations != nil {
```
Stevenjin8: here, we should do `getNonCanaryIngress`.

pkg/i2gw/emitters/envoygateway/emitter.go
Stevenjin8: We already have a common emitter, but you can implement this part how you like

pkg/i2gw/emitters/envoygateway/buffer.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw"
+	emitterir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/notifications"
+)
+
+func (e *Emitter) EmitBuffer(ir emitterir.EmitterIR, gwResources *i2gw.GatewayResources) {
+	for _, ctx := range ir.HTTPRoutes {
+		MergeBodySizeIR(&ctx, ctx.BodySizeByRuleIdx)
+
+		for idx, bs := range ctx.BodySizeByRuleIdx {
```
Stevenjin8, line 35: In the (near) future, we'll probably want to emit warnings for unprocessed IR, which means that we need to mark the IR as processed by either setting it to nil or deleting it. So at the end of this block, we should add something along the lines of `ctx.BodySizeByRuleIdx[idx] = nil`

pkg/i2gw/providers/ingressnginx/bodysize.go
kkk777-7: thanks! updated regex.

pkg/i2gw/providers/ingressnginx/bodysize.go
kkk777-7: thanks, updated logic.

kkk777-7:
Envoy Gateway’s latest version requires Go 1.25.3. But, golangci-lint v1 doesn’t support Go 1.25.
If we can migrate to golangci-lint v2, I’ll handle that in a separate PR.

pkg/i2gw/emitters/envoygateway/buffer.go
puertomontt: ctx is a value not a reference, so setting `ctx.BodySizeByRuleIdx[idx] = nil` won't change contents in ir.HTTPRoutes

kkk777-7: thanks! I've added write back process.

Stevenjin8:
@kkk777-7 tests are failing
/approve

kkk777-7:
Hi @Stevenjin8, golangci-lint v1 doesn't support Go v1.25. so happened CI failure.
I opened separate PR for migrating golangci-lint v2. https://github.com/kubernetes-sigs/ingress2gateway/pull/323

I’d appreciate it if we could merge #323 first, and then merge this one afterward.

PR #307, opened 2026-01-20
feat: Add support for use-regex annotation

chakravardhan:
**What type of PR is this?**
/kind feature

This PR implements support for the `nginx.ingress.kubernetes.io/use-regex` annotation, enabling Regex Path Matching in the generated Gateway API resources.

Key Changes

- Annotation: Added `use-regex` support.
- Mapping: Maps paths to `HTTPRoute.spec.rules[].matches[].path.type: RegularExpression` when the annotation is set to "true".
- Infrastructure: Added `implementationSpecificPathMatch` handler to allow ImplementationSpecific paths to pass through the common conversion layer before being upgraded to RegularExpression.

Verification
- Unit Tests: Added `regex_test.go` covering positive (regex) and negative (prefix default) cases.

```release-note

```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/regex_test.go
+++ b/pkg/i2gw/providers/ingressnginx/regex_test.go
@@ -131,7 +131,7 @@ func TestRegexFeature(t *testing.T) {
 
 			// Manual IR setup simulating what common.ToIR would produce
 			key := types.NamespacedName{Namespace: tc.ingress.Namespace, Name: common.RouteName(tc.ingress.Name, "example.com")}
-			
+
 			// Determine initial match type based on input PathType (simulating generic conversion)
 			var initialMatchType *gatewayv1.PathMatchType
 			if *tc.ingress.Spec.Rules[0].HTTP.Paths[0].PathType == networkingv1.PathTypePrefix {
@@ -140,7 +140,7 @@ func TestRegexFeature(t *testing.T) {
 				// ImplementationSpecific often defaults to Prefix if not handled, or just stays nil/impl-specific
 				// For the sake of this test, let's assume common.ToIR set it to something or we are testing the overwrite.
 				// But common.ToIR throws error for ImplSpecific if no custom converter.
-				// However, regexFeature runs AFTER common.ToIR. 
+				// However, regexFeature runs AFTER common.ToIR.
 				// Let's assume common.ToIR generated a Prefix match (soft default) or checks if feature handles it.
 				// Actually, simplified: we just want to see if regexFeature *updates* it.
 				// So we init with Prefix for both cases.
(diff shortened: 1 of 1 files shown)


pkg/i2gw/providers/ingressnginx/converter.go
```diff
@@ -59,3 +63,11 @@ func (c *resourcesToIRConverter) convert(storage *storage) (providerir.ProviderI
...
 	return ir, errs
 }
+
+func implementationSpecificPathMatch(path *gatewayv1.HTTPPathMatch) {
+	// Nginx Ingress Controller treats ImplementationSpecific as Prefix by default,
+	// unless regex characters are present (handled by regexFeature).
+	// We safely default to Prefix here to pass the common.ToIR check.
+	t := gatewayv1.PathMatchPathPrefix
+	path.Type = &t
+}
```
Stevenjin8, line 73: this is super annoying... IIRC the `pathType` is completely ignored when using regex.

pkg/i2gw/providers/ingressnginx/regex.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+			sources := httpRouteCtx.RuleBackendSources[ruleIdx]
+			if len(sources) == 0 {
+				continue
+			}
+
+			// Check if any source ingress has the regex annotation enabled.
+			// If merged from multiple sources, we assume consistency or take the first positive.
+			useRegex := false
+			for _, source := range sources {
```
Stevenjin8: you can use getNonCanaryIngress here.

pkg/i2gw/providers/ingressnginx/regex.go
```diff
@@ -0,0 +1,64 @@
...
+				if source.Ingress.Annotations[UseRegexAnnotation] == "true" {
```
Stevenjin8: this can be any truthy value. Use strconv.ParseBool

pkg/i2gw/providers/ingressnginx/converter.go
chakravardhan: Valid point. Since the upstream validation requires a known path type, I've kept the logic here to default `ImplementationSpecific` to Prefix so it passes the common ToIR check. As you noted, the regex feature will later correctly override this to `RegularExpression` in the final output.

pkg/i2gw/providers/ingressnginx/regex.go
chakravardhan: Done. Updated to use `getNonCanaryIngress` to ensure we pick the correct primary ingress.

pkg/i2gw/providers/ingressnginx/regex.go
chakravardhan: Done. Switched to `strconv.ParseBool` to handle all truthy values correctly

Stevenjin8:
@chakravardhan, I jsut realized that https://github.com/kubernetes-sigs/ingress2gateway/pull/255 does the same thing, so I'm going to favour that one just because it came first.

Stevenjin8 (approved):

PR #313, opened 2026-01-21
Feat/gke session cookie stacked

rajashish:
What type of PR is this? /kind feature

What this PR does / why we need it: This PR adds parsing support for Nginx Ingress session affinity annotations to the ingress-nginx provider and maps them to the GCE emitter.

Specifically, it:

Adds session-cookie-name support: Parses nginx.ingress.kubernetes.io/session-cookie-name into the Provider IR.
Note: While parsed, this is currently not emitted to 
GCPBackendPolicy
 because the underlying gke-gateway-api (v1.3.0) does not yet support custom cookie names for GENERATED_COOKIE. A comment has been added to 
output_extensions.go
 to enable this once the API supports it.
Fixes session-cookie-expires: Updates the parsing logic to use the correct nginx.ingress.kubernetes.io/session-cookie-expires annotation key (previously session-cookie-max-age).
Maps this to cookieTtlSec in 
GCPBackendPolicy
 when affinity is set to cookie.

Which issue(s) this PR fixes: Fixes #

Does this PR introduce a user-facing change?:

release-note
Added support for parsing `nginx.ingress.kubernetes.io/session-cookie-name` and `nginx.ingress.kubernetes.io/session-cookie-expires` annotations in the ingress-nginx provider. The `session-cookie-expires` annotation is now correctly mapped to `cookieTtlSec` in GKE `GCPBackendPolicy`.

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/session_affinity.go
+++ b/pkg/i2gw/providers/ingressnginx/session_affinity.go
@@ -105,7 +105,7 @@ func sessionAffinityFeature(notify notifications.NotifyFunc, _ []networkingv1.In
 					svc.SessionAffinity.Type = affinityType
 					svc.SessionAffinity.CookieTTLSec = cookieTTL
 					svc.SessionAffinity.CookieName = cookieName
-					svc.SessionAffinity.Metadata = emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations")}, "")
+					svc.SessionAffinity.Metadata = emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations", "nginx.ingress.kubernetes.io/affinity")}, "")
 
 					// Update the map
 					ir.Services[svcKey] = svc
@@ -113,7 +113,7 @@ func sessionAffinityFeature(notify notifications.NotifyFunc, _ []networkingv1.In
 					// Service doesn't exist yet, create it
 					svc = providerir.ProviderSpecificServiceIR{
 						SessionAffinity: &emitterir.SessionAffinity{
-							Metadata:     emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations")}, ""),
+							Metadata:     emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations", "nginx.ingress.kubernetes.io/affinity")}, ""),
 							Type:         affinityType,
 							CookieTTLSec: cookieTTL,
 							CookieName:   cookieName,
--- a/pkg/i2gw/providers/ingressnginx/session_affinity_test.go
+++ b/pkg/i2gw/providers/ingressnginx/session_affinity_test.go
@@ -56,7 +56,7 @@ func TestGCEFeature(t *testing.T) {
 				},
 			},
 			expectedSessionAffinity: &emitterir.SessionAffinity{
-				Metadata: emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations")}, ""),
+				Metadata: emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations", "nginx.ingress.kubernetes.io/affinity")}, ""),
 				Type:     "Cookie",
 			},
 		},
@@ -72,7 +72,7 @@ func TestGCEFeature(t *testing.T) {
 				},
 			},
 			expectedSessionAffinity: &emitterir.SessionAffinity{
-				Metadata:     emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations")}, ""),
+				Metadata:     emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations", "nginx.ingress.kubernetes.io/affinity")}, ""),
 				Type:         "Cookie",
 				CookieTTLSec: ptr.To(int64(3600)),
 			},
@@ -89,7 +89,7 @@ func TestGCEFeature(t *testing.T) {
 				},
 			},
 			expectedSessionAffinity: &emitterir.SessionAffinity{
-				Metadata:   emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations")}, ""),
+				Metadata:   emitterir.NewExtensionFeatureMetadata("ingress-nginx", []*field.Path{field.NewPath("annotations", "nginx.ingress.kubernetes.io/affinity")}, ""),
 				Type:       "Cookie",
 				CookieName: "MY_COOKIE",
 			},
--- a/pkg/i2gw/emitters/utils/utils.go
+++ b/pkg/i2gw/emitters/utils/utils.go
@@ -150,4 +150,55 @@ func LogUnparsedErrors(ir emitterir.EmitterIR, notify notifications.NotifyFunc) 
 			)
 		}
 	}
+
+	for svcKey, serviceContext := range ir.Services {
+		if serviceContext.SessionAffinity != nil {
+			meta := serviceContext.SessionAffinity.Metadata
+			source := meta.Source()
+			paths := strings.Builder{}
+			for _, p := range meta.Paths() {
+				paths.WriteString(p.String())
+				paths.WriteString(", ")
+			}
+
+			message := meta.FailureMessage()
+			if message == "" {
+				message = "Session Affinity is not supported by this emitter"
+			}
+
+			var associatedHR *gatewayv1.HTTPRoute
+			for _, hrContext := range ir.HTTPRoutes {
+				for _, rule := range hrContext.HTTPRoute.Spec.Rules {
+					for _, br := range rule.BackendRefs {
+						if string(br.Name) == svcKey.Name {
+							associatedHR = &hrContext.HTTPRoute
+							break
+						}
+					}
+					if associatedHR != nil {
+						break
+					}
+				}
+				if associatedHR != nil {
+					break
+				}
+			}
+
+			var warnObj *gatewayv1.HTTPRoute
+			if associatedHR != nil {
+				// Create a copy to modify type meta safely
+				copyHR := associatedHR.DeepCopy()
+				if copyHR.TypeMeta.Kind == "" {
+					copyHR.TypeMeta.Kind = "HTTPRoute"
+					copyHR.TypeMeta.APIVersion = "gateway.networking.k8s.io/v1"
+				}
+				warnObj = copyHR
+			}
+
+			notify(notifications.WarningNotification,
+				fmt.Sprintf("failed to parse %s from Ingress %s: %s", source, strings.TrimSuffix(paths.String(), ", "), message),
+				warnObj,
+			)
+		}
+	}
 }
--- a/pkg/i2gw/notifications/notifications.go
+++ b/pkg/i2gw/notifications/notifications.go
@@ -37,10 +37,18 @@ type Notification struct {
 }
 
 func objectsToStr(ob []client.Object) string {
-	strs := make([]string, len(ob))
+	strs := make([]string, 0, len(ob))
 
-	for i, o := range ob {
-		strs[i] = o.GetObjectKind().GroupVersionKind().Kind + ": " + client.ObjectKeyFromObject(o).String()
+	for _, o := range ob {
+		if o == nil {
+			strs = append(strs, "Unknown")
+			continue
+		}
+		kind := o.GetObjectKind().GroupVersionKind().Kind
+		if kind == "" {
+			kind = "Object"
+		}
+		strs = append(strs, kind+": "+client.ObjectKeyFromObject(o).String())
 	}
 
 	return strings.Join(strs, ", ")


rajashish:
@Stevenjin8 @spencerhance please review

rajashish:
> Please sign CLA first

Done

Stevenjin8 (changes requested):
I know we have some gce stuff in emitter IR, but that's only there for backcompatibility reasons and to make the code diff in 273. Any new changes to the gce emitter should not make any changes to GCE/GKE related code. (Ideally it would get rid of it)

rajashish:
> I know we have some gce stuff in emitter IR, but that's only there for backcompatibility reasons and to make the code diff in 273. Any new changes to the gce emitter should not make any changes to GCE/GKE related code. (Ideally it would get rid of it)

Thanks for the feedback! I've refactored the implementation to keep the Emitter IR provider-neutral:

Generic IR: Added a generic 

SessionAffinity
 struct to the native 

EmitterIR
 to hold provider-neutral intent (Type: "Cookie", CookieName, etc.).
Nginx Parser: Updated the ingress-nginx parser to populate this generic struct instead of parsing into GCE-specific fields.
GCE Emitter: Updated the output extensions to translate the generic "Cookie" intent into the native GENERATED_COOKIE enum expected by the 

GCPBackendPolicy
 CRD.
Reverted GCE Structs: Rolled back the PR changes to the legacy gce.SessionAffinityConfig so it remains strictly for back-compatibility.
Tested end-to-end against a GKE cluster and the generated YAML output matches the previous behavior perfectly! Let me know if this looks good.

rajashish:
@Stevenjin8 Please review

rajashish:
Hi @Stevenjin8 @spencerhance @LiorLieberman Please review this pr

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -74,6 +85,8 @@ type HTTPRouteContext struct {
 	// IPRangeControlByRuleIdx maps HTTPRoute rule indices to IP range control intent.
 	// This is provider-neutral and applied by each custom emitter.
 	IPRangeControlByRuleIdx map[int]*IPRangeControl
+
+	Gce *gce.HTTPRouteIR
```
Stevenjin8: Any way we can have EmitterIR not have any GCE specific fields ( or at the very least not add any)

rajashish: Good catch, thanks! I completely agree.

I've pushed an update that removes the newly added (and empty) gce.HTTPRouteIR field from both the EmitterIR and ProviderSpecificHTTPRouteIR. This ensures we adhere to the 'at the very least not add any new ones' principle.

For the older legacy fields (like GceServices and GceGatewayIR), genericizing those would require a massive rewrite of the old GCE Ingress Provider parser. Since this PR is focused on ingress-nginx Session Affinity, I propose we tackle removing those older fields in a separate refactoring PR.

(Side note: I noticed the common converter was hardcoding invalid name: rule-%d tags onto HTTPRoutes which breaks Gateway API validation, so I also cleaned that up across the codebase!)

Let me know if you're good with this approach!

pkg/i2gw/providers/ingressnginx/session_affinity_test.go
```diff
@@ -0,0 +1,153 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
bexxmodd: Remove year

```suggestion
Copyright The Kubernetes Authors.
```

Per https://github.com/kubernetes/steering/issues/299, we no longer require copyright years in the file boilerplate for Kubernetes projects.

pkg/i2gw/providers/ingressnginx/session_affinity.go
```diff
@@ -0,0 +1,114 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
bexxmodd: Remove year

```suggestion
Copyright The Kubernetes Authors.
```

pkg/i2gw/emitters/gce/gce.go
```diff
@@ -74,9 +74,33 @@ func (c *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.Err
...
+
+	upsellGatewayClass(&gatewayResources)
+	removeHTTPRouteRuleNames(&gatewayResources)
 	return gatewayResources, nil
 }
 
+func upsellGatewayClass(gatewayResources *i2gw.GatewayResources) {
+	for i, gw := range gatewayResources.Gateways {
+		if gw.Spec.GatewayClassName == "nginx" {
+			gw.Spec.GatewayClassName = "gke-l7-global-external-managed"
```
bexxmodd: I'm not following why nginx is automatically converted to GXM?

rajashish: Good catch! That was originally an old hack to seamlessly output external LB YAML for test inputs, but I agree it's an anti-pattern to silently mutate the user's intent. I have gone ahead and deleted upsellGatewayClass out of 

gce.go
 so it retains the original class explicitly.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -84,7 +96,6 @@ type HTTPRouteContext struct {
 	gatewayv1.HTTPRoute
 	// TCPTimeoutsByRuleIdx holds provider TCP-level timeouts by HTTPRoute rule index.
 	TCPTimeoutsByRuleIdx map[int]*TCPTimeouts
-
```
Stevenjin8, line 87: add this back

pkg/i2gw/providers/common/converter.go
```diff
@@ -457,11 +457,6 @@ func (rg *ingressRuleGroup) toHTTPRoute(servicePorts map[types.NamespacedName]ma
 		httpRoute.Spec.Rules = append(httpRoute.Spec.Rules, hrRule)
 		allRuleBackendSources = append(allRuleBackendSources, sources)
 	}
-
-	for idx := range httpRoute.Spec.Rules {
-		httpRoute.Spec.Rules[idx].Name = ptr.To(gatewayv1.SectionName(fmt.Sprintf("rule-%d", idx)))
-	}
-
```
Stevenjin8, line 464: why?

rajashish: I removed those lines because assigning an arbitrary name like rule-%d to HTTPRouteRule.Name was actually a blocker: it was breaking strict Gateway API schema validation on Kubernetes clusters.

While this did functionally change the auto-generated output (names are no longer assigned automatically), it was an intentional and necessary fix to adhere to strict validation rules on the cluster and keep the manifests valid.

However, I would definitely like to hear your perspective on this!

pkg/i2gw/provider_intermediate/conversion.go
```diff
@@ -32,13 +33,20 @@ func ToEmitterIR(pIR ProviderIR) emitterir.EmitterIR {
...
+		ctx := emitterir.GatewayContext{Gateway: v.Gateway}
+		if v.ProviderSpecificIR.Gce != nil {
+			ctx.Gce = v.ProviderSpecificIR.Gce
+		}
+		eIR.Gateways[k] = ctx
 	}
 	for k, v := range pIR.HTTPRoutes {
-		eIR.HTTPRoutes[k] = emitterir.HTTPRouteContext{HTTPRoute: v.HTTPRoute}
+		ctx := emitterir.HTTPRouteContext{HTTPRoute: v.HTTPRoute}
+		eIR.HTTPRoutes[k] = ctx
```
Stevenjin8, line 49: This diff doesn't do anything right?

pkg/i2gw/providers/common/converter.go
Stevenjin8: Can you give me an example? downstream emitters rely on this behaviour. Adding a name to a route shouldnt cause an error

pkg/i2gw/providers/ingressnginx/annotations.go
```diff
@@ -61,7 +61,6 @@ const (
 
 	// SSL Redirect annotation
 	SSLRedirectAnnotation = "nginx.ingress.kubernetes.io/ssl-redirect"
-
```
Stevenjin8, line 64: add this back

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -70,9 +70,21 @@ type EmitterIR struct {
 	BackendTLSPolicies map[types.NamespacedName]BackendTLSPolicyContext
 	ReferenceGrants    map[types.NamespacedName]ReferenceGrantContext
 
+	Services map[types.NamespacedName]ServiceContext
+
 	GceServices map[types.NamespacedName]gce.ServiceIR
 }
 
+type SessionAffinity struct {
```
Stevenjin8, line 78:
```suggestion
type SessionAffinity struct {
    Metadata        ExtensionFeatureMetadata
```

pkg/i2gw/providers/common/converter.go
Stevenjin8: Also, if you are going to explicitly remove the route names later, then we can keep this right?

Stevenjin8:
My main concern is that unparsed behavior isn't reported right now.

For example, for the following

```yaml
# ing.yaml
kind: Ingress
metadata:
  annotations:
    argocd.argoproj.io/tracking-id: loki-distributed:networking.k8s.io/Ingress:logging/loki-distributed-gateway
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
    nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
    nginx.ingress.kubernetes.io/denylist-source-range: ""
    nginx.ingress.kubernetes.io/affinity: cookie
spec:
  ingressClassName: nginx
  rules:
    - host: abc
      http:
        paths:
          - backend:
              service:
                name: abc
                port:
                  number: 80
            path: /
            pathType: Prefix
```

I get 

```bash
14:24:31 (kind-kind) sj/i2gwstuff/feat-gke-session-cookie-stacked/ingress2gateway $ go run . print  --providers ingress-nginx  --input-file ing.yaml > /dev/null
┌─ WARN  ────────────────────────────────────────
│  Ingress NGINX serves TLS traffic for host "abc" with a self-signed certificate. This behavior will not be translated and the host will not be accessible via HTTPS.
│  source: INGRESS-NGINX
└─
┌─ WARN  ────────────────────────────────────────
│  Unsupported annotation nginx.ingress.kubernetes.io/force-ssl-redirect
│  source: INGRESS-NGINX
│  object: Ingress: /
└─
┌─ WARN  ────────────────────────────────────────
│  ingress-nginx only supports TCP-level timeouts; i2gw has made a best-effort translation to Gateway API timeouts.request. Please verify that this meets your needs. See documentation: https://gateway-api.sigs.k8s.io/guides/http-timeouts/
│  source: INGRESS-NGINX
│  object: HTTPRoute: /-abc
└─
┌─ WARN  ────────────────────────────────────────
│  Failed to apply [][].metadata.annotations."nginx.ingress.kubernetes.io/denylist-source-range" from /: IP-based authorization is not supported
│  source: STANDARD_EMITTER
│  object: HTTPRoute: /-abc
└─
┌─ WARN  ────────────────────────────────────────
│  Gateway API does not support configuring URL normalization (RFC 3986, Section 6). Please check if this matters for your use case and consult implementation-specific details.
│  source: STANDARD_EMITTER
└─
```

Note that there is a warning about `denylist-source-range`, but nothing about the `affinity` annotation.

rajashish:
I've updated the PR with the following changes:
1. Added Metadata field to SessionAffinity inside intermediate_representation.go for better tracking.
2. 2. Restored rule naming in the common converter (including for default backends).
3. 3. Added strict warnings to the Nginx session affinity parser for unsupported fields (unsupported cookies).
4. 4. Updated tests to match.
Verified these changes against a sample Ingress file to ensure correct warning emission and rule naming.

pkg/i2gw/providers/common/converter.go
rajashish: Restored the rule naming to ensure consistency for all providers. Done.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
rajashish: Added the `Metadata` field to `SessionAffinity` as suggested. Done.

pkg/i2gw/providers/common/converter.go
rajashish: Done! I have restored the rule naming loop in the common converter ToHTTPRoute (and also ensured rule naming for default backends). This adheres to standard Gateway API requirements for rules while leaving it up to individual emitters (like GCE) to strip them if they are not supported in their specific target controller implementation.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
rajashish: Done! Added a Metadata ExtensionFeatureMetadata field to the SessionAffinity struct inside intermediate_representation.go. Updated the Nginx provider to populate this with accuracy (pointing specifically back to the accurate annotation: annotations.nginx.ingress.kubernetes.io/affinity).

rajashish:
> You're going to need to add to `LogUnparsedErrors` for the sessions affinity.
> 
> Also, if we are not parsing the cookie name, we should emit some sort of warning rather than just a comment in the code (users aren't going to read the code)

Done! The reason it wasn't warning was that utils.LogUnparsedErrors() only inspected unparsed extensions tied to HTTPRouteContext.

I have updated LogUnparsedErrors in pkg/i2gw/emitters/utils/utils.go to also traverse ir.Services. Now, structured intermediate features placed on services—like SessionAffinity—safely surface warning notifications if the current emitter proceeds without consuming them (which is the case for the Standard Emitter). I also patched notifications.go to prevent any nil crashes during list assemblies

rajashish:
Hey @Stevenjin8  resolved all comments. Please review again

Stevenjin8:
@rajashish this only takes into account, nginx.ingress.kubernetes.io/affinity, but not affinity-mode, and session-cookie-*

Stevenjin8:
honestly, if something is parsed but not emitted anywhere, I think its best left unparsed.

Stevenjin8:
closing in favor of #394

PR #330, opened 2026-02-02
Test TLS termination, SSL-redirect, Canary, and CORS

Stevenjin8:
**What type of PR is this?**

/kind test

Fixes #

```release-note
NONE
```

### DIFF
--- a/e2e/gwapi.go
+++ b/e2e/gwapi.go
@@ -170,7 +170,7 @@ func createGatewayClasses(ctx context.Context, l logger, client *gwclientset.Cli
 			&gc,
 			metav1.CreateOptions{},
 		)
-		if errors.IsAlreadyExists(err) { // False when err is nil
+		if errors.IsAlreadyExists(err) {
 			_, err = client.GatewayV1().GatewayClasses().Update(ctx, &gc, metav1.UpdateOptions{})
 		}
 		if err != nil {
--- a/e2e/gwapi_kong.go
+++ b/e2e/gwapi_kong.go
@@ -93,7 +93,7 @@ func deployGatewayAPIKong(
 
 	l.Logf("Creating GatewayClass %s", kongGatewayClass)
 	_, err := gwClient.GatewayV1().GatewayClasses().Create(ctx, gc, metav1.CreateOptions{})
-	if errors.IsAlreadyExists(err) { // false if err is nil
+	if errors.IsAlreadyExists(err) {
 		_, err = gwClient.GatewayV1().GatewayClasses().Update(ctx, gc, metav1.UpdateOptions{})
 	}
 	if err != nil {
--- a/e2e/util.go
+++ b/e2e/util.go
@@ -22,7 +22,7 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
-// Generates a cryptographically random alphanumeric string of length n. Uses crypto/rand to ensure
+// Generates a cryptographically random alphanumeric string of length 5. Uses crypto/rand to ensure
 // uniqueness even when called from parallel tests.
 func randString() (string, error) {
 	n := 5


johananl:
Right now `make e2e` is failing. This is caused by the addition of a new package `./e2e/ingressnginx`.

The e2e tests are executed on a single Kind cluster in parallel. To avoid races, we use a resource manager singleton:

https://github.com/kubernetes-sigs/ingress2gateway/blob/2802595be53f437116630f9d7aa38a8ea47f69af/e2e/resource_manager.go#L23-L25

This assumes all e2e tests are under the same package. The addition of a new sub-package creates **two instances** of the test framework which in effect duplicates the resource manager. This makes the two e2e package race against the Kind cluster.

To address that, we could do one of the following:

1. Manage all e2e tests in a single package (we could still separate them by files).
2. Change the locking mechanism to use e.g. **file-based locks** which would work across processes.
3. Disable parallelism.

I suggest we do 1.

johananl:
I think you need to rebase. I'm currently seeing many commits already in main in your feature branch.

e2e/testutil.go
```diff
@@ -286,7 +309,13 @@ func setUpIngressPortForwarding(
 			pf, addr, err := startPortForwardToService(ctx, k8sClient, restConfig, svc.Namespace, svc.Name, 80)
 			require.NoError(t, err, "starting port forward to ingress-nginx")
 			pfs = append(pfs, pf)
-			addresses[ingressnginx.NginxIngressClass] = addr
+			addresses[ingressnginx.NginxIngressClass] = Addresses{HTTP: addr}
+			if useHTTPS {
```
johananl: This switch is becoming big and seems to be largely duplicated code. I wonder if we should make it generic and extract only the provider-specific values to some helper data structure. Please treat as optional since this isn't 100% in scope for this specific PR.

Stevenjin8:
@johananl That makes sense. I put everything in one package.

I also changed the GWAPI crds to be a monthly release just so we can redirect to 308. We can change it back once 1.5 is release, or maybe we want to do that for our release branches?

e2e/crd.go
```diff
@@ -37,8 +37,9 @@ import (
 )
 
 const (
-	gatewayAPIVersion    = "v1.4.1"
-	gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/" + gatewayAPIVersion + "/standard-install.yaml"
+	// gatewayAPIVersion    = "v1.4.1"
```
johananl, line 40: Forgot to clean up?

e2e/ingress_nginx_canary_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"regexp"
+	"testing"
+
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/ingressnginx"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/istio"
+	"github.com/stretchr/testify/require"
+	networkingv1 "k8s.io/api/networking/v1"
+)
+
+func TestCanary(t *testing.T) {
```
johananl: I think we might want to include "nginx" in the function name. Right now we see the following when the tests are run:

```
TestCanary/to_Istio/base_canary
```

This doesn't tell us which provider gets tested.

My initial thought with the naming was to organize the hierarchy by "from", "to" and "test case". This way you get e.g. `TestIngressNginx/to_Istio/with_host_field`. It's clear what exactly is being tested.

We should also think about **filtering** tests when choosing the naming, e.g. maybe I want to do "run all tests to Istio" or "run basic conversions for all providers".

I'm happy to change the naming scheme but IMO we want to be deliberate about it.

WDYT?

e2e/ingress_nginx_canary_test.go
```diff
@@ -0,0 +1,74 @@
...
+	t.Parallel()
+	t.Run("to Istio", func(t *testing.T) {
+		t.Parallel()
+		t.Run("base canary", func(t *testing.T) {
+			suffix, err := randString(6)
```
johananl: I'm curious - why 6? So far we used 5. We should probably standardize. Maybe we can hardcode a default in the helper to avoid inconsistencies.

e2e/ingress_nginx_canary_test.go
```diff
@@ -0,0 +1,74 @@
...
+			require.NoError(t, err)
+			host := fmt.Sprintf("canary-%s.com", suffix)
```
johananl, line 37: I assume you set the host explicitly because both ingresses need to share the same host, right? ICYMI, we generate random hosts by default when not specified in the test cases.

e2e/verifiers.go
```diff
@@ -18,59 +18,212 @@ package e2e
...
+	minSuccesses float64
+	maxSuccesses float64
+	runs         int
+}
+
+func (v *canaryVerifier) verify(ctx context.Context, log logger, addr addresses, ingress *networkingv1.Ingress) error {
+	successes := 0
+	for i := 0; i < v.runs; i++ {
+		err := v.verifier.verify(ctx, log, addr, ingress)
+		if err != nil {
```
johananl: This looks like a mistake: We're incrementing `successes` when `err` is NOT nil, namely we're counting `verify()` failures as successes.

e2e/ingress_nginx_canary_test.go
```diff
@@ -0,0 +1,74 @@
...
+						WithAnnotation("nginx.ingress.kubernetes.io/canary-weight", "20").
+						WithBackend(DummyAppName2).
+						Build(),
+				},
+				verifiers: map[string][]verifier{
+					"foo1": {
+						&canaryVerifier{
+							verifier:     &httpGetVerifier{host: host, path: "/hostname", bodyRegex: regexp.MustCompile("^dummy-app2")},
+							runs:         200,
+							minSuccesses: 0.7,
```
johananl: I'm confused by the logic here. We're routing 20% of the traffic to dummy app 2, then we're expecting to see **~80%** of requests on dummy app 2, and this seems to work because we have a wrong boolean in the verifier (see comment below) which makes us count failures as successes?

IIUC we have two logical errors which "cancel" one another out, i.e. we want to set the min and max successes to 0.1 and 0.3 and we want to fix the bug in the verifier.

e2e/verifiers.go
```diff
@@ -18,59 +18,212 @@ package e2e
...
+		allowedCodes = []int{http.StatusOK}
+	}
+	allowed := false
+	for _, code := range allowedCodes {
+		if res.StatusCode == code {
+			allowed = true
+			break
+		}
+	}
+	if !allowed {
```
johananl:
```suggestion
	if !slices.Contains(allowedCodes, res.StatusCode) {
```

e2e/ingress_nginx_cors_test.go
```diff
@@ -0,0 +1,378 @@
...
+)
+
+func TestCORS(t *testing.T) {
+	t.Parallel()
+	t.Run("to Istio", func(t *testing.T) {
+		t.Parallel()
+		t.Run("typical cors annotations", func(t *testing.T) {
+			suffix, err := randString(6)
+			require.NoError(t, err)
+			host := fmt.Sprintf("cors-%s.com", suffix)
```
johananl: You don't need to explicitly set hosts unless there is some special case like canary. `runTestCase()` generates per-ingress hosts with random strings by default.

e2e/verifiers.go
```diff
@@ -18,59 +18,212 @@ package e2e
...
+	return nil
 }
 
 type httpGetVerifier struct {
-	host string
-	path string
+	host           string
+	path           string
+	method         string
+	requestHeaders map[string]string
```
johananl, line 79: We probably want to document these fields. I had to read the code to understand what some of them do. How does the verifier behave when I set field X?

e2e/ingress_nginx_tls_test.go
```diff
@@ -0,0 +1,161 @@
...
+	t.Parallel()
+	t.Run("to Istio", func(t *testing.T) {
+		t.Parallel()
+		t.Run("tls ingress and gateway", func(t *testing.T) {
+			suffix, err := randString(6)
+			if err != nil {
+				t.Fatalf("creating host suffix: %v", err)
+			}
+			host := "tls-" + suffix + ".example.com"
+			tlsSecret, err := GenerateTLSTestSecret("tls-cert-"+suffix, host)
```
johananl: We can probably generate the secrets transparently in `runTestCase` like we do for hostnames (see [here](https://github.com/kubernetes-sigs/ingress2gateway/pull/330/changes#r2799009750)).

The idea is to keep the test case definitions as **declarative** as possible, i.e. specify only what's strictly necessary for this specific case and keep imperative logic in helpers.

Makefile
```diff
@@ -134,7 +134,7 @@ e2e: build ## Run end-to-end tests.
 	fi; \
 	set -x; \
 	I2GW_BINARY_PATH=$(REPO_ROOT)/ingress2gateway KUBECONFIG=$${kubeconfig} \
-		go test $(I2GW_GO_TEST_ARGS) $(REPO_ROOT)/e2e; \
+		go test $(I2GW_GO_TEST_ARGS) $(REPO_ROOT)/e2e/...; \
```
johananl: We don't need this if all tests are under `./e2e` (and it's probably safer to remove this to avoid accidental execution of unrelated packages in case someone adds a sub-package in the future).

e2e/scenarios_test.go
```diff
@@ -49,15 +50,18 @@ func TestIngressNginx(t *testing.T) {
...
+						Build(),
 				},
 				verifiers: map[string][]verifier{
 					"foo": {&httpGetVerifier{path: "/"}},
 				},
 			})
 		})
+		suffix, err := randString(6)
+		require.NoError(t, err)
+		host := "foo.example.com" + suffix
```
johananl, line 64: Do we need this? Is it to avoid having an "unsiffixed" `foo.example.com`? Anyway, this test case mainly tests that the framework handles explicit hostnames in test cases. Maybe we don't need this test case at all.

e2e/testutil.go
```diff
@@ -576,29 +634,71 @@ type ingressBuilder struct {
 	*networkingv1.Ingress
 }
 
-func (b *ingressBuilder) withName(name string) *ingressBuilder {
+func (b *ingressBuilder) WithName(name string) *ingressBuilder {
```
johananl: We no longer need to unexport symbols since everything is under `./e2e`.

e2e/scenarios_test.go
johananl: On second thought, do you think we even need this test file at all? Since we're defining test cases in multiple files now, maybe `scenarios_test.go` doesn't make much sense.

e2e/secrets.go
```diff
@@ -0,0 +1,140 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type TLSTestSecret struct {
+	Secret *corev1.Secret
+	CACert []byte
+}
+
+func GenerateTLSTestSecret(name, commonName string) (*TLSTestSecret, error) {
```
johananl: Do we need this wrapper function? If so, why is it exported whereas the other functions (e.g. `createSecrets()`) aren't?

e2e/testutil.go
```diff
@@ -52,13 +52,17 @@ import (
 
 // A prefix for all namespaces used in the e2e tests.
 const e2ePrefix = "i2gw"
+const DummyAppName1 = "dummy-app1"
```
johananl, line 55: Are we assuming many test cases would need two dummy apps? I'm wondering if we need to define these consts and do so **here** rather than defining plain text directly in the test cases.

e2e/testutil.go
```diff
@@ -114,7 +118,7 @@ func runTestCase(t *testing.T, tc *testCase) {
 	require.NoError(t, crdResource.wait(), "Gateway API CRDs installation failed")
 
 	providers := deployProviders(ctx, t, k8sClient, kubeconfig, tc.providers, tc.gatewayImplementation, skipCleanup)
-	gwImpl := deployGatewayImplementation(ctx, t, k8sClient, gwClient, kubeconfig, tc.gatewayImplementation, skipCleanup)
+	gwImpl := deploygatewayImplementation(ctx, t, k8sClient, gwClient, kubeconfig, tc.gatewayImplementation, skipCleanup)
```
johananl: Why this rename? It breaks the Go naming convention.

e2e/verifiers.go
```diff
@@ -18,59 +18,212 @@ package e2e
...
 	"io"
 	"net/http"
+	"regexp"
+	"strings"
 	"time"
 
 	networkingv1 "k8s.io/api/networking/v1"
 )
 
-// Validates that a service is accessible and working correctly. The addr parameter is a
```
johananl, line 29: IMO we want _some_ documentation here, especially if we have many types of verifiers (which we do).

e2e/verifiers.go
```diff
@@ -18,59 +18,212 @@ package e2e
...
+	caCertPEM      []byte
+	bodyRegex      *regexp.Regexp
+}
+
+type headerMatch struct {
+	name     string
+	patterns []*regexp.Regexp
+}
+
+type headerExclude struct {
```
johananl: AFAICT headerMatch can be used negatively. I assume you also wanted headerExclude for convenience? If so, can't we use the same code for both cases with a `negate bool` flag or something to keep things DRY?

e2e/verifiers.go
```diff
@@ -18,59 +18,212 @@ package e2e
...
+	method := v.method
+	if method == "" {
+		method = http.MethodGet
+	}
+	req, err := http.NewRequestWithContext(ctx, method, fmt.Sprintf("%s://%s%s", scheme, targetAddr, v.path), nil)
 	if err != nil {
 		return fmt.Errorf("constructing HTTP request: %w", err)
 	}
 
-	// If the Host field is specified in the test case, use that. Otherwise, default to deriving
```
johananl, line 46: Why remove this comment?

e2e/ingress_nginx_canary_test.go
Stevenjin8: is it the same random host?

Stevenjin8: I can check

e2e/ingress_nginx_tls_test.go
Stevenjin8: I was thinking about this. But then sometimes we want to generate secrets sometimes we don't. If we generate multiple different ones, we want to tell the verifier which on to use. Or what if we want to share or not share secrets for different hostnames? We could look at the gateway tls s and generate them in `runTestCase`, but then what should be the CN? do we want it to have SANs?

On the topic of certificate generation, we're going to have to generate them for frontend and backend tls. Do we also want that to happen in `runTestCase`? It seemed much easier to have the secret generation logic in one place, rather than plumbing a bunch of options around.

e2e/ingress_nginx_canary_test.go
johananl: No, each ingress gets a unique name:

```go
	for _, ing := range tc.ingresses {
		for i := range ing.Spec.Rules {
			if ing.Spec.Rules[i].Host == "" {
				ing.Spec.Rules[i].Host = fmt.Sprintf("%s.%s.%s.test", ing.Name, randPrefix, e2ePrefix)
			}
		}
	}
```

johananl: AFAICT canary is currently the only place where we might need to set hostnames explicitly (because we have a requirement that all ingresses in the test case share the same host IIUC).

e2e/scenarios_test.go
Stevenjin8: Ingress NGINX doesn't let you create multiple ingress with the same path and hostname. So when if I run the tests with no cleanup, then I would need to delete the ingresses every time, which is kinda annoying.

e2e/crd.go
Stevenjin8: not really. I think we want to target gwapi 1.5, which isn't released yet (rc coming this week) because we really want support for status code 308 in ReqestRedirects. Since 308 isn't available in 1.4, but is in 1.5, I figured I'd use  a monthly release and just update it to 1.5/1.5-rc whenever that comes out.

e2e/ingress_nginx_canary_test.go
Stevenjin8: +1. I can either throw IngressNGINX into the test name, or I can make each of these functions called in scenarios_test.go. Up to you

e2e/scenarios_test.go
Stevenjin8: Yeah id rather have one provider/ingress per file

e2e/secrets.go
Stevenjin8: vestage of when stuff was in a different module

e2e/testutil.go
Stevenjin8: (Most of what I'm doing is based on what istio does). I think its common enough to do it, but definitely worth thinking about if we are going to test timeouts or something

e2e/ingress_nginx_canary_test.go
Stevenjin8: yeah different hosts are generated

e2e/ingress_nginx_canary_test.go
johananl: >or I can make each of these functions called in scenarios_test.go

I don't understand this ☝️ 

I don't have a strong opinion on the naming, just wanted to make sure you're aware of my attempt to group things in a logical way.

e2e/ingress_nginx_canary_test.go
johananl: Then I guess we do need the hostname logic in the test case itself. That's fine.

e2e/ingress_nginx_tls_test.go
johananl: Fair enough. I don't understand the details here but sounds like you thought about it and chose this deliberately so I'm good with it.

e2e/scenarios_test.go
johananl: > Yeah id rather have one provider/ingress per file

So what should we do with `scenarios_test.go`? Break it up into multiple files?

e2e/gwapi.go
```diff
@@ -169,6 +170,9 @@ func createGatewayClasses(ctx context.Context, l logger, client *gwclientset.Cli
 			&gc,
 			metav1.CreateOptions{},
 		)
+		if !errors.IsAlreadyExists(err) {
```
johananl: This conditional is true when err is nil which looks like a bug: It runs an update when the gateway class **doesn't** exist, which makes some tests fail (run `make e2e` to reproduce this).

How about the following (which also gives a separate error message when an _update_ fails):

```go
		if errors.IsAlreadyExists(err) { // False when err is nil
			if _, err = client.GatewayV1().GatewayClasses().Update(ctx, &gc, metav1.UpdateOptions{}); err != nil {
				return nil, fmt.Errorf("updating GatewayClass %s: %w", name.String(), err)
			}
		}
		if err != nil {
			return nil, fmt.Errorf("creating GatewayClass %s: %w", name.String(), err)
		}
```

e2e/gwapi_kong.go
```diff
@@ -92,6 +93,9 @@ func deployGatewayAPIKong(
 
 	l.Logf("Creating GatewayClass %s", kongGatewayClass)
 	_, err := gwClient.GatewayV1().GatewayClasses().Create(ctx, gc, metav1.CreateOptions{})
+	if !errors.IsAlreadyExists(err) {
```
johananl: Same bug as [this](https://github.com/kubernetes-sigs/ingress2gateway/pull/330/changes#r2804264810).

e2e/ingress_nginx_canary_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/stretchr/testify/require"
+	networkingv1 "k8s.io/api/networking/v1"
+)
+
+func TestIngressNGINXCanary(t *testing.T) {
+	t.Parallel()
+	t.Run("to Istio", func(t *testing.T) {
+		t.Parallel()
+		t.Run("base canary", func(t *testing.T) {
+			suffix, err := randString(6)
```
johananl: Why 6 and not 5? We could use either but let's be consistent (run a grep on `randString`).

e2e/ingress_nginx_cors_test.go
```diff
@@ -0,0 +1,409 @@
...
+							headerMatches: []headerMatch{
+								{
+									name: "Access-Control-Allow-Origin",
+									patterns: []*maybeNegativePattern{
+										{pattern: regexp.MustCompile("^" + regexp.QuoteMeta(origin) + "$"), negate: false},
+									},
+								},
+								{
+									name: "Access-Control-Allow-Credentials",
+									// patterns: []*regexp.Regexp{regexp.MustCompile(`(?i)^true$`)},
```
johananl: Leftovers?

e2e/ingress_nginx_cors_test.go
```diff
@@ -0,0 +1,409 @@
...
+									patterns: []*maybeNegativePattern{
+										{pattern: regexp.MustCompile(`(?i)^true$`), negate: false},
+									},
+								},
+								{
+									name: "Access-Control-Expose-Headers",
+									// patterns: []*regexp.Regexp{
```
johananl: Leftovers?

e2e/scenarios_test.go
johananl: I don't know if we should do it in this PR but IMO we want to remove `scenarios_test.go` and relocate any relevant tests to more specific `_test.go` files. This file contains some initial smoke tests I've included with the e2e framework. Since we moved to a multi-file setup this file is IMO obsolete.

e2e/ingress_nginx_tls_test.go
```diff
@@ -0,0 +1,165 @@
...
+						&httpRequestVerifier{
+							host:         host,
+							path:         "/",
+							allowedCodes: []int{308},
+							useTLS:       false,
+							headerMatches: []headerMatch{
+								{
+									name: "Location",
+									patterns: []*maybeNegativePattern{
+										{pattern: regexp.MustCompile("^https://" + host + "/?$"), negate: false},
```
johananl: Just in case you aren't aware: You can omit `negate` when you want `false` since that's the zero value.

e2e/verifiers.go
```diff
@@ -29,40 +34,190 @@ import (
...
+	pattern *regexp.Regexp
+	negate  bool
+}
+
+func (m maybeNegativePattern) matches(s string) bool {
+	match := m.pattern.MatchString(s)
+	if m.negate {
+		return !match
 	}
+	return match
```
johananl: Nit: This is logical XOR. We can simplify. But maybe clear is better than clever.


```suggestion
    return m.pattern.MatchString(s) != m.negate
```

e2e/ingress_nginx_canary_test.go
Stevenjin8: oops, thought I changed that one

e2e/gwapi.go
Stevenjin8: I think I just have to remove the `!`

e2e/util.go
```diff
@@ -24,7 +24,8 @@ import (
 
 // Generates a cryptographically random alphanumeric string of length n. Uses crypto/rand to ensure
```
johananl: Might want to update the comment since n is an implementation detail now:

```go
// Generates a cryptographically random 5-character alphanumeric string.
```

e2e/scenarios_test.go
Stevenjin8: Makes sense, I'll do it in a follow up PR and have `runTestCase` take one provider and multiple gwapi implementations

Makefile
Stevenjin8: idk why this is still showing as modified

e2e/gwapi.go
```diff
@@ -169,6 +170,9 @@ func createGatewayClasses(ctx context.Context, l logger, client *gwclientset.Cli
 			&gc,
 			metav1.CreateOptions{},
 		)
+		if errors.IsAlreadyExists(err) { // False when err is nil
```
johananl: The comment was meant for you 🙂 I didn't mean that we need to add it. In any case, there is case mismatch with the other comment in `e2e/gwapi_kong.go` but this is really nitpicking now. I'd just drop these two comments.

e2e/scenarios_test.go
johananl: OK. I don't understand why multiple implementations but let's discuss in the other PR.

e2e/util.go
johananl: @Stevenjin8 did you see this?

johananl:
Thanks!
/lgtm

PR #336, opened 2026-02-05
Feature/gce infra

chakravardhan:
**What type of PR is this?**
/kind feature

This PR adds support for a new provider-specific flag `--gce-gateway-class-name` to the GCE emitter.

Currently, the GCE provider defaults to standard GatewayClasses (e.g., `gke-l7-global-external-managed` or `gke-l7-rilb`) based on the input Ingress class. This change allows users to override the [GatewayClassName](cci:1://file:///usr/local/google/home/chakravardhan/ingress2gateway/pkg/i2gw/emitters/gce/gce_test.go:616:0-667:1) field in all generated Gateways when using the `gce` emitter, which is useful for migration testing or custom Gateway deployments.

If the flag is unset, the emitter preserves the default classes determined by the provider.

Fixes #

```release-note
Added `--gce-gateway-class-name` flag to GCE emitter to allow overriding the output GatewayClassName.

### DIFF
--- a/pkg/i2gw/providers/gce/ir_converter_test.go
+++ b/pkg/i2gw/providers/gce/ir_converter_test.go
@@ -942,6 +942,7 @@ func Test_convertToIR(t *testing.T) {
 								Hostnames: []gatewayv1.Hostname{gatewayv1.Hostname(testHost)},
 								Rules: []gatewayv1.HTTPRouteRule{
 									{
+										Name: common.PtrTo(gatewayv1.SectionName("rule-0")),
 										Matches: []gatewayv1.HTTPRouteMatch{
 											{
 												Path: &gatewayv1.HTTPPathMatch{
@@ -961,6 +962,11 @@ func Test_convertToIR(t *testing.T) {
 											},
 										},
 									},
+								},
+							},
+							Status: gatewayv1.HTTPRouteStatus{
+								RouteStatus: gatewayv1.RouteStatus{
+									Parents: []gatewayv1.RouteParentStatus{},
 								},
 							},
 						},

### CODE REVIEW

File: pkg/i2gw/emitter.go [hunk 1]
```diff
@@ -61,4 +61,5 @@ type EmitterConstructor func(conf *EmitterConf) Emitter
 
 type EmitterConf struct {
 	// TODO: add fields as needed.
+	ProviderSpecificFlags map[string]map[string]string
```
Stevenjin8: I realize that a lot of the flags weren't built with emitters in mind, but is there any way to not have ProviderSpecificFlags in the Emitter configuration.

Stevenjin8: otherwise this PR lgtm

chakravardhan: I've completely removed `ProviderSpecificFlags` from the EmitterConf.

To solve the` --gce-gateway-class-name` issue, I moved the flag registration from the GCE Emitter to the GCE Provider. The provider now extracts this flag directly from its own configuration (ProviderConf) and applies the GatewayClass override into the Intermediate Representation (IR) during the initial `convertToIR` translation phase instead.

This keeps the Emitters perfectly clean and completely agnostic of any provider-specific CLI flags

pkg/i2gw/providers/gce/ir_converter.go
```diff
@@ -80,7 +80,17 @@ func (c *resourcesToIRConverter) convertToIR(storage *storage) (providerir.Provi
 		return providerir.ProviderIR{}, errs
 	}
 
-	errs = setGCEGatewayClasses(ingressList, ir.Gateways)
+	// Extract gatewayClassName from provider-specific flags
+	var gatewayClassName string
+	if c.conf != nil && c.conf.ProviderSpecificFlags != nil {
+		if flags, ok := c.conf.ProviderSpecificFlags["gce"]; ok {
+			if val, ok := flags[GatewayClassNameFlag]; ok && val != "" {
```
bexxmodd: So if no value is supplied for the flag it will be a no-op?

pkg/i2gw/providers/gce/utils.go
```diff
@@ -55,7 +55,12 @@ func setGCEGatewayClasses(ingresses []networkingv1.Ingress, gatewayContexts map[
 }
 
 // setGCEGatewayClass sets the Gateway to the corresponding GCE GatewayClass.
-func setGCEGatewayClass(ingress networkingv1.Ingress, gateway gatewayv1.Gateway) (gatewayv1.Gateway, *field.Error) {
+func setGCEGatewayClass(ingress networkingv1.Ingress, gateway gatewayv1.Gateway, gatewayClassName string) (gatewayv1.Gateway, *field.Error) {
+	if gatewayClassName != "" {
+		gateway.Spec.GatewayClassName = gatewayv1.ObjectName(gatewayClassName)
```
bexxmodd, line 60: I see this accepts any non-empty string as the gw class name. Shouldn't we do some sort of validation for it? What if user supplies non-existent class name?

pkg/i2gw/providers/gce/ir_converter.go
spencerhance: I agree with Beka, can we do some validation here to make sure that the flag value is valid?

If the value is non-nil, it should be one of the valid class names here: https://docs.cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities (not an empty string either).

I'm not totally sure if we should be doing validation further up the stack for ProviderSpecificFlags, but it doesn't look like any other provider is doing that right now.

pkg/i2gw/providers/gce/utils.go
spencerhance: I agree, but I think pkg/i2gw/providers/gce/ir_converter.go is probably a better place than doing it here.

pkg/i2gw/providers/gce/ir_converter.go
chakravardhan: Agreed, Added a validation logic in `ir_converter.go` right where we first extract the flag payload from the ProviderSpecificFlags map.

pkg/i2gw/providers/gce/utils.go
chakravardhan: Already added the validation logic upstream

pkg/i2gw/providers/gce/ir_converter.go
```diff
@@ -80,7 +80,21 @@ func (c *resourcesToIRConverter) convertToIR(storage *storage) (providerir.Provi
...
+				if !SupportedGKEGatewayClasses[val] {
+					errs = append(errs, field.Invalid(field.NewPath("provider-specific-flags"), val, "invalid GCE gateway class name"))
+					return providerir.ProviderIR{}, errs
+				}
+				gatewayClassName = val
+			}
+		}
+	}
```
bexxmodd, line 92: Let's simplify nested if statements:

```Go
	// Extract gatewayClassName from provider-specific flags
	var gatewayClassName string
	if c.conf != nil && c.conf.ProviderSpecificFlags != nil {
		gatewayClassName = c.conf.ProviderSpecificFlags["gce"][GatewayClassNameFlag]
	}

	if gatewayClassName != "" && !SupportedGKEGatewayClasses[gatewayClassName] {
		errs = append(errs, field.Invalid(field.NewPath("provider-specific-flags"), gatewayClassName, "invalid GCE gateway class name"))
		return providerir.ProviderIR{}, errs
	}
```

bexxmodd (changes requested):

bexxmodd (approved):

pkg/i2gw/providers/gce/ir_converter.go
spencerhance: not totally sure if this use of "field.Invalid()" is correct here, since IIUC that library is designed for referencing k8s object fields.  However, there doesn't appear to be a good mechanism for a non-field error currently.  Seems okay to me for now but cc @Stevenjin8 .

chakravardhan:
> /lgtm

@LiorLieberman , Can I get approval on this PR? Thanks!

PR #341, opened 2026-02-06
Redesign notification mechanism

### LINKED ISSUE #340: Notifications should be logs
Right now we have the concept of "notifications". We define a set of types and methods under `pkg/i2gw/notifications/notifications.go`, then we use these types in package-specific `notify()` functions which we call whenever we want to alert the user to special situations during ingress->GWAPI conversion. Finally we present the notifications to the user formatted as a table which gets printed to stderr.

I see the following problems with the current design:

1. We rely on **global state**.
2. We **duplicate** our logging infrastructure.

Let me elaborate on each problem.

## Global State

In order to format notifications as a table, we must print all notifications for a provider together. To do _that_, we need to store all notifications as **state** while the binary is running and construct the table at the end.

Right now, this state is stored in a global variable:

https://github.com/kubernetes-sigs/ingress2gateway/blob/2802595be53f437116630f9d7aa38a8ea47f69af/pkg/i2gw/notifications/notifications.go#L29-L31

As long as i2gw is run as a binary there is no problem. However, when trying to convert resources **concurrently** (which we do, for example, in our e2e tests) the global variable becomes a source of problems. For example, right now there is no mechanism for **clearing** stored notifications, so the global state is an ever-growing map of notifications which e.g. makes notifications from provider A to be printed for provider B.

In addition, although we do use a mutex for the global map, under certain conditions we still have races (which I've encountered while working on the e2e tests).

Lastly, AFAICT the only reason we need to store a global map is so that we can **format notifications as a table**. After all, if instead we printed notifications **ad hoc** as they occurred we wouldn't need to store them at all. From the user's perspective there is no difference between stored state and ad-hoc printing: In both cases notifications show up on stderr while GWAPI resources are shown on stdout. However, a table output requires state storage under the hood while ad hoc printing doesn't.

## Duplicate Logging Infrastructure

In addition to notifications, we also have **logs**. In fact, we often have the two _together_ with the same data. Example:

https://github.com/kubernetes-sigs/ingress2gateway/blob/2802595be53f437116630f9d7aa38a8ea47f69af/pkg/i2gw/providers/istio/converter.go#L203-L214

If we look at the types we define for notifications, we notice that this package is essentially a logging package. We implement log levels:

https://github.com/kubernetes-sigs/ingress2gateway/blob/2802595be53f437116630f9d7aa38a8ea47f69af/pkg/i2gw/notifications/notifications.go#L33-L37

We have a basic logging function (which stores the event as state rather than print immediately):

https://github.com/kubernetes-sigs/ingress2gateway/blob/2802595be53f437116630f9d7aa38a8ea47f69af/pkg/i2gw/notifications/notifications.go#L54-L59

And we have a "handler" which reads state, applies formatting and emits events for printing:

https://github.com/kubernetes-sigs/ingress2gateway/blob/2802595be53f437116630f9d7aa38a8ea47f69af/pkg/i2gw/notifications/notifications.go#L61-L88

In effect, we created **our own custom, stateful logging package**, which we then use to instrument the entire code base.

## Proposed Solution

I want propose that **notifications are logs**. If we accept that, we can:

- Leverage existing Go patterns and libraries for logging.
- Get rid of global state which IMO creates many problems for almost no benefit.
- Deduplicate our code and make it easier to test.
- Make the code less fragile.
- Allow consuming the i2gw code as a library with no extra cost (think controllers which migrate ingresses automatically for example).

Specifically, I want to suggest we leverage `log/slog`. We can have two handlers: A pretty handler which formats events for human consumption and a JSON handler for processing by machines.

I'll soon open a draft PR with a proposed solution we can evaluate.

## Related Issues:

https://github.com/kubernetes-sigs/ingress2gateway/issues/325
https://github.com/kubernetes-sigs/ingress2gateway/issues/314
https://github.com/kubernetes-sigs/ingress2gateway/issues/276

johananl:
**What type of PR is this?**

/kind feature

This PR proposes a refactor for the "notifications" concept by treating notifications as logs -- see #340 for background and a detailed rationale.

I've implemented a pretty handler which can serve as an alternative to the text tables we currently use without requiring global state storage:

image

There is also a JSON handler (`--log-format json`).

Fixes #340

```release-note
Pretty-print notifications as logs and deduplicate events
```

### DIFF
--- a/docs/emitters.md
+++ b/docs/emitters.md
@@ -21,7 +21,7 @@ The common emitter will output `EmitterIR`.
 
 Implementation-specific Emitters will read the `EmitterIR` from the common emitter and output Gateway API resources along with implementation-specific resources.
 Both providers and emitters MUST log any information that is lost in translation.
-Ideally, when there is new IR and an emitter does not implement it, ingress2gateway should automatically emit notifications.
+Ideally, when there is new IR and an emitter does not implement it, ingress2gateway should automatically emit logs.
 
 Note that the emitters MAY modify the Gateway API resources as needed.
 That said, emitters MUST NOT output non-Gateway API resources unless absolutely necessary.


johananl:
Note: The PR is big mainly because the entire code base is "instrumented" with the notifications infra. This PR replaces this instrumentation with the standard `log/slog`.

cmd/root.go
```diff
@@ -20,26 +20,50 @@ import (
...
 		PersistentPreRun: func(_ *cobra.Command, _ []string) {
 			getKubeconfig()
+			initLogging()
 		},
 	}
 
 	rootCmd.PersistentFlags().StringVar(&kubeconfig, "kubeconfig", "",
 		`The kubeconfig file to use when talking to the cluster. If the flag is not set, a set of standard locations can be searched for an existing kubeconfig file.`)
+
+	rootCmd.PersistentFlags().StringVar(&logFormat, "log-format", "text",
```
johananl, line 52: We already have `--output` for controlling the format of printed **resources**. This new flag is for controlling the format of **logs**. The naming inconsistency isn't ideal (why one has "format" and the other doesn't?). OTOH I don't know if we want to change `--output` to something like `--output-format`.

pkg/i2gw/providers/gce/implementation_specific.go
```diff
@@ -39,23 +38,31 @@ import (
 // | /v1                                   | /v1 Exact                              |
 // | /v1/                                  | /v1/ Exact                             |
 // | /v1/*                                 | /v1 Prefix                             |
-func implementationSpecificHTTPPathTypeMatch(path *gatewayv1.HTTPPathMatch) {
-	pmExact := gatewayv1.PathMatchExact
-	pmPrefix := gatewayv1.PathMatchPathPrefix
+func implementationSpecificHTTPPathTypeMatch(log *slog.Logger) func(path *gatewayv1.HTTPPathMatch) {
```
johananl, line 41: We change this function to return a closure so that we can pass a logger while avoiding changing the general `ImplementationSpecificHTTPPathTypeMatchConverter()` signature since most instances of that type don't log.

johananl:
- [x] Ensure we can easily generate a "report" from all the logs.
  - [x] Add a flag to disable color output.
  - [x] Ensure the formatting makes sense when writing logs to a file.

pkg/i2gw/logging/handler.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+// PrettyHandler is an slog.Handler that outputs human-readable colored text.
+type PrettyHandler struct {
+	opts   *slog.HandlerOptions
+	output io.Writer
+	mu     *sync.Mutex
+	attrs  []slog.Attr
+	groups []string
+}
```
Stevenjin8, line 50: Yeah we should detect if the output detects colors.

Stevenjin8:
I also think a lot of these "info" should be debug, but that's for another PR.

johananl:
> I also think a lot of these "info" should be debug, but that's for another PR.

Yes, I didn't touch the "content" of the logs at all here.

pkg/i2gw/logging/handler.go
johananl: What do you mean? I've added `--no-color`. Think we should do anything else?

Stevenjin8: there are ways to detect if the output terminal supports color, but adding a cli flag works too

johananl: I think that would add complexity/fragility, plus we might want to have an explicit flag anyway in case people want to use the tool to generate a "report" by piping stderr to a file.

cmd/root.go
Stevenjin8: I think its good as is

johananl:
I'm discussing this design with @LiorLieberman. Putting back on hold until we reach a conclusion.
/hold

johananl:
Superseded by #360.

PR #342, opened 2026-02-07
Feat cors eg

kkk777-7:
**What type of PR is this?**

/kind feature

Support CORS in the Envoy Gateway emitter.

Currently, Envoy Gateway GA supports CORS via `CRD/SecurityPolicy`.
So, when a user selects the Envoy Gateway emitter, it should generate CRD without requiring `--allow-experimental-gw-api`.
when user specify `--allow-experimental-gw-api` while selecting EG emitter, it will generate CORS filter. https://github.com/kubernetes-sigs/ingress2gateway/pull/303
Also In the future, once Gateway API GA supports CORS, the common emitter should take precedence and generate a CORS filter.

Fixes #

```release-note
support CORS feature in Envoy Gateway Emitter.
```

### DIFF
--- a/pkg/i2gw/emitters/envoygateway/buffer.go
+++ b/pkg/i2gw/emitters/envoygateway/buffer.go
@@ -33,6 +33,11 @@ func (e *Emitter) EmitBuffer(ir emitterir.EmitterIR, gwResources *i2gw.GatewayRe
 		MergeBodySizeIR(&ctx)
 
 		for idx, bs := range ctx.BodySizeByRuleIdx {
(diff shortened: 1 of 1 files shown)


Stevenjin8:
@kkk777-7 

CORS just became GA in Gateway API :)

https://gateway-api.sigs.k8s.io/reference/spec/#httpcorsfilter and gateway api 1.5 is coming out soon, so I think, if anything, we should have it always emit gwapi CORS Filter regardless of --allow-experimental.

kkk777-7:
@Stevenjin8  thanks for comments!
that makes sense, so I close this PR :)

PR #344, opened 2026-02-11
Fix regex and add some tests

Stevenjin8:
**What type of PR is this?**

/kind bug

Fixes #

```release-note
NONE
```
Apparently regex is per host not per ingress.... Also regex are prefix matches, but gateway api is exact

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/regex_test.go
+++ b/pkg/i2gw/providers/ingressnginx/regex_test.go
@@ -76,7 +76,7 @@ func TestRegexFeature(t *testing.T) {
 				{
 					Path: &gatewayv1.HTTPPathMatch{
 						Type:  &regexType,
-						Value: ptr.To("/users/.*/profile.*"),
+						Value: ptr.To("(?i)/users/.*/profile.*"),
 					},
 				},
 			},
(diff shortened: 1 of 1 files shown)


Stevenjin8:
need to investigatew some more

pkg/i2gw/providers/ingressnginx/regex.go
```diff
@@ -23,50 +23,64 @@ import (
...
-					if path != nil {
-						val := gatewayv1.PathMatchRegularExpression
-						path.Type = &val
-					}
+		for i, rule := range httpRouteCtx.Spec.Rules {
+			for j, path := range rule.Matches {
+				if path.Path != nil {
+					// Ingress nginx regex path matches are prefix matches by default
+					httpRouteCtx.Spec.Rules[i].Matches[j].Path.Type = ptr.To[gatewayv1.PathMatchType](gatewayv1.PathMatchRegularExpression)
+					httpRouteCtx.Spec.Rules[i].Matches[j].Path.Value = ptr.To(*httpRouteCtx.Spec.Rules[i].Matches[j].Path.Value + ".*")
```
bexxmodd: what if the regexes that use anchors or already have trailing wildcards? Maybe check for `.*` and `$` before appending?

Stevenjin8: I thought about this and decided it wasn't worth the effort.
* Performance wise, we're not doing pathological things with matches that could cause quadratic behaviour I think. Regardless, RE2 is always linear.
* Patterns that end with `.*.*` and `$.*` are valid (but kinda weird looking) regexes
* its not as simple as checking whether patterns end with `$` or `.*`, because there might be escapes or capture groups.

PR #345, opened 2026-02-12
feat: ip range control

kkk777-7:
**What type of PR is this?**

/kind feature

Support IP Range Control annotation in ingress nginx.

- nginx.ingress.kubernetes.io/whitelist-source-range
- nginx.ingress.kubernetes.io/denylist-source-range

currently, Envoy Gateway GA supports IP allowlist/denylist.
https://gateway.envoyproxy.io/docs/tasks/security/restrict-ip-access/

Fixes #

```release-note
Implemented support for Nginx IP source range annotations (whitelist-source-range, denylist-source-range) and Envoy Gateway SecurityPolicy output.
```

### DIFF
--- a/e2e/crd.go
+++ b/e2e/crd.go
@@ -37,8 +37,9 @@ import (
 )
 
 const (
-	gatewayAPIVersion    = "v1.4.1"
-	gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/" + gatewayAPIVersion + "/standard-install.yaml"
+	// gatewayAPIVersion    = "v1.4.1"
+	// gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/" + gatewayAPIVersion + "/experimental-install.yaml"
+	gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/monthly-2026.01/monthly-2026.01-install.yaml"
 )
 
 func deployCRDs(ctx context.Context, l logger, client *apiextensionsclientset.Clientset, skipCleanup bool) (func(), error) {
--- a/e2e/dummy_app.go
+++ b/e2e/dummy_app.go
@@ -31,21 +31,20 @@ import (
 )
 
 const (
-	name    = "dummy-app"
 	image   = "registry.k8s.io/e2e-test-images/agnhost"
 	version = "2.39"
 )
 
-func deployDummyApp(ctx context.Context, l logger, client *kubernetes.Clientset, namespace string, skipCleanup bool) (func(), error) {
-	if err := createDummyAppDeployment(ctx, l, client, namespace); err != nil {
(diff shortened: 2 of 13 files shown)


pkg/i2gw/emitters/envoygateway/iprange.go
```diff
@@ -0,0 +1,110 @@
...
+							Action: egapiv1a1.AuthorizationActionDeny,
+							Principal: egapiv1a1.Principal{
+								ClientCIDRs: lo.Map(ipCon.DenyList, func(a string, _ int) egapiv1a1.CIDR {
+									return egapiv1a1.CIDR(a)
+								}),
+							},
+						},
+					},
+				}
+			}
```
Stevenjin8: This syntax will probably wrong, but you get the point

```suggestion
securityPolicy.Spec.Authorization = &egapiv1a1.Authorization{
					DefaultAction: ptr.To(egapiv1a1.AuthorizationActionDeny),
	}
	if len(ipCon.AllowList) > 0 {
				securityPolicy.Spec.Authorization.Rules = append(securityPolicy.Spec.Authorization.Rules, 
				egapiv1a1.AuthorizationRule{
						{
							Action: egapiv1a1.AuthorizationActionAllow,
							Principal: egapiv1a1.Principal{
								ClientCIDRs: lo.Map(ipCon.AllowList, func(a string, _ int) egapiv1a1.CIDR {
									return egapiv1a1.CIDR(a)
								}),

				}
			} 
			if len(ipCon.DenyList) > 0 {
				// same append as above
				}
			}
```

Stevenjin8: also do we worry about conflicts?

pkg/i2gw/emitters/envoygateway/iprange.go
```diff
@@ -0,0 +1,110 @@
...
+
+		for idx, ipCon := range ctx.IPRangeControlByRuleIdx {
+			var sectionName *gwapiv1.SectionName
+			if idx != RouteRuleAllIndex && idx < len(ctx.Spec.Rules) {
+				sectionName = ctx.Spec.Rules[idx].Name
+			}
+
+			securityPolicy := e.getOrBuildSecurityPolicy(ctx, sectionName, idx)
+
+			if len(ipCon.AllowList) > 0 && len(ipCon.DenyList) > 0 {
```
Stevenjin8: do we worry about conflicts here?

Stevenjin8 (commented):
@kkk777-7 seems good at a high level. Do you think you could add eg to the integration test framework? Doesn't have to be in this PR, but I think it would be good to have integration tests for every feature.

pkg/i2gw/emitters/envoygateway/iprange.go
kkk777-7: maybe, If both deny lists and allow lists exist, they can be able to coexist, with the deny list being evaluated first.
https://github.com/kubernetes/ingress-nginx/blob/main/rootfs/etc/nginx/template/nginx.tmpl#L1206-L1214

kkk777-7:
@Stevenjin8 thanks for review!
I'll add EG integration test in follow-up PRs.

PR #360, opened 2026-02-23
Refactor notification mechanism

### LINKED ISSUE #340: Notifications should be logs

johananl:
**What type of PR is this?**

/kind feature

This PR refactors the notification mechanism. Here is what it does specifically:

- Replace the global notifications map with a new `Report` type.
- One local `Report` instance is created in `ToGatewayAPIResources()`.
- A `*Report` pointer is passed to each provider/emitter constructor.
- Each provider/emitter calls `Report.Notifier()` to create its own provider/emitter-scoped `NotifyFunc` and stores this function as a property.
- Conversion code can now emit notifications by calling a package-scoped `notify()` function.

There are no user-facing changes -- this is an implementation change only. The size of the PR is largely due to repetitive replacements of the "instrumentation" everywhere.

Fixes #340 

```release-note
NONE
```


pkg/i2gw/notifications/report_test.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/stretchr/testify/require"
+
+	istioclientv1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+func TestReportRender(t *testing.T) {
```
LiorLieberman, line 33: Question: wouldn't this test be flaky since we are doing non guaranteed order in cmd/print.go

```go
for _, table := range report.Render() {
    fmt.Fprintln(os.Stderr, table)
}
```

Should we have the order fixed by having Render() returning a sorted slice of strings?

pkg/i2gw/notifications/notifications.go
```diff
@@ -44,61 +36,14 @@ type Notification struct {
...
+func objectsToStr(ob []client.Object) string {
+	strs := make([]string, len(ob))
 
 	for i, o := range ob {
-		if i > 0 {
-			sb.WriteString(", ")
-		}
-		object := o.GetObjectKind().GroupVersionKind().Kind + ": " + client.ObjectKeyFromObject(o).String()
-		sb.WriteString(object)
+		strs[i] = o.GetObjectKind().GroupVersionKind().Kind + ": " + client.ObjectKeyFromObject(o).String()
```
LiorLieberman, line 43: would we have any scenario where o is nil? If so, it will panic

pkg/i2gw/notifications/report_test.go
johananl: No. The test calls `Render()` directly. The snippet you mentioned is from `PrintGatewayAPIObjects()` which isn't part of the test.

But you _are_ right that the order of the **sources** (i.e. providers/emitters) is random, i.e. there is no guarantee to the order at which the various notification tables get printed. This is the case even before this refactor. I can add a sort there if you think that's important.

pkg/i2gw/notifications/notifications.go
johananl: AFAICT a programmer would have to explicitly insert a nil value as a slice element for `o` to be nil. I can add an extra defensive nil check if you want (but this logic isn't changed by this PR, i.e. we aren't introducing new fragility here).

Stevenjin8:
@johananl Implementation wise, this seems fine to me, but is there any way we can use the formatting from the other PR? I like it so much better. The current one is almost unreadable.

image

LiorLieberman:
> @johananl Implementation wise, this seems fine to me, but is there any way we can use the formatting from the other PR? I like it so much better. The current one is almost unreadable.

I love the table format more. But I should not insist here since you became the experts now. 

LGTM from my side, feel free to align with @Stevenjin8 feedback in case you both decide. 

/approve
/hold - for you to control merging - thanks!

johananl:
We have no character limit on the notification message which can mess up the tables when the terminal isn't big enough.

I'm fine with either format. The colored box design was done using an slog handler. I can look into "porting" the visuals without reintroducing the logging library (since we've agreed to keep notifications as distinct from logs).

Stevenjin8:
@johananl makes sense.
/lgtm

johananl:
I'm porting the design from #341.

PR #361, opened 2026-02-23
Log unparsed IR

Stevenjin8:
**What type of PR is this?**

Fixes #

```release-note
NONE
```

cc @kkk777-7

### DIFF
--- a/pkg/i2gw/emitters/common_emitter/emitter_test.go
+++ b/pkg/i2gw/emitters/common_emitter/emitter_test.go
@@ -161,10 +161,10 @@ func TestEmitCORSFiltering(t *testing.T) {
 			expectedFiltersCount: 1,
 		},
 		{
-			name:                 "experimental denied + cors in sidecar -> cors NOT added",
+			name:                 "experimental denied + cors in sidecar -> cors added",
 			allowExperimental:    false,
 			corsInSidecar:        &emitterir.CORSConfig{},
-			expectedFiltersCount: 0,
+			expectedFiltersCount: 1,
 		},
 		{
 			name:              "other filters preserved regardless of flag",
@@ -173,7 +173,7 @@ func TestEmitCORSFiltering(t *testing.T) {
 				{Type: gatewayv1.HTTPRouteFilterRequestHeaderModifier},
 			},
 			corsInSidecar:        &emitterir.CORSConfig{},
-			expectedFiltersCount: 1, // only header modifier
+			expectedFiltersCount: 2, // only header modifier
 		},
 	}
 
--- a/pkg/i2gw/providers/ingressnginx/rewrite.go
+++ b/pkg/i2gw/providers/ingressnginx/rewrite.go
@@ -17,8 +17,8 @@ limitations under the License.
 package ingressnginx
 
 import (
-	"strings"
 	"fmt"
+	"strings"
 
 	emitterir "github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/emitter_intermediate"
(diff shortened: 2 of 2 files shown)


pkg/i2gw/emitters/standard/standard.go
```diff
@@ -37,5 +37,6 @@ func NewEmitter(_ *i2gw.EmitterConf) i2gw.Emitter {
 
 // Emit converts the provider intermediate representation to Gateway API resources.
 func (e *Emitter) Emit(ir emitterir.EmitterIR) (i2gw.GatewayResources, field.ErrorList) {
+	utils.LogUnparsedErrors(ir, notify)
```
kkk777-7, line 40: what do you think about moving this function to ingress2gateway.go around L104?

That way, none of the emitters need to be aware of this function, which helps prevent missing output from some emitters.
But, there’s also the consideration that structurally we’d like to handle emitterIR within the emitter package.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -69,13 +97,28 @@ type HTTPRouteContext struct {
...
+	for _, x := range h.BodySizeByRuleIdx {
+		if x != nil {
+			unparsedExtensions = append(unparsedExtensions, &x.Metadata)
+		}
+	}
+	for _, x := range h.IPRangeControlByRuleIdx {
+		if x != nil {
+			unparsedExtensions = append(unparsedExtensions, &x.Metadata)
+		}
+	}
```
kkk777-7, line 118: should we add `CorsPolicyByRuleIdx` so that users notice they need to specify allow-experimental-gw-api?

pkg/i2gw/emitter_intermediate/intermediate_representation.go
```diff
@@ -94,16 +138,22 @@ type PathRewrite struct {
...
 
 // IPRangeControl represents provider-neutral IP range control intent.
 type IPRangeControl struct {
+	Metadata  ExtensionFeatureMetadata
 	AllowList []string
 	DenyList  []string
 }
 
+type CORSConfig struct {
+	gatewayv1.HTTPCORSFilter
```
kkk777-7, line 159: should we add Metadata here?

pkg/i2gw/emitter_intermediate/intermediate_representation.go
kkk777-7: currently, we've release GW v1.5.0, so this comment is wrong :) 
https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.5.0

and I'll create PR that support CORS without flag.

pkg/i2gw/emitter_intermediate/intermediate_representation.go
kkk777-7: this comment is solved (I can't resolve myself in GH)

pkg/i2gw/emitters/standard/standard.go
Stevenjin8: This is what I originally had in mind. I ran into two problems

1. Emitters output `GatewayResources` which do not have IR. So for this to work, Emitters would need to output both `GatewayResources` and emitter IR, which seems a bit awkward
2. I like the idea that emitters have control over what they output.

kkk777-7: it makes sense! thanks

kkk777-7:
LGTM, thanks!
could you rebase?

kkk777-7:
LGTM, thanks!
/lgtm

PR #366, opened 2026-02-26
Reorganize e2e tests

johananl:
**What type of PR is this?**

/kind test

This PR reorganizes the e2e tests. It introduces a clean separation between the e2e "framework" code, helper code for deploying ingress providers and GWAPI implementations and files containing test case definitions, thus making the package easier to understand and more maintainable:

- Framework code now lives under `e2e/framework`.
- GWAPI implementation deployment helpers now live under `e2e/implementation`.
- Ingress provider deployment helpers now live under `e2e/provider`.
- Test cases now live directly under `e2e` in `*_test.go` files.

See the updated `e2e/README.md` in this PR for more info on the directory structure and testing approach.

None

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,5 @@ go.work
 .vscode/
 
 kind-kubeconfig
+e2e/kind.yaml
 bin
--- a/Makefile
+++ b/Makefile
@@ -102,10 +102,25 @@ ensure-kind:
 		echo "kind installed successfully to $(LOCAL_BIN)/kind"; \
 	fi
 
+KIND_CONFIG := $(REPO_ROOT)/e2e/kind.yaml
+
+.PHONY: generate-kind-config
+generate-kind-config:
+	@mkdir -p $(dir $(KIND_CONFIG))
+	@printf '%s\n' \
+		'kind: Cluster' \
+		'apiVersion: kind.x-k8s.io/v1alpha4' \
+		'nodes:' \
+		'- role: control-plane' \
+		'- role: worker' \
+		'- role: worker' \
+		'- role: worker' \
+		> $(KIND_CONFIG)
+
 .PHONY: kind
-kind: ensure-kind
+kind: ensure-kind generate-kind-config
 	@if ! $(KIND) get clusters | grep -q i2gw-e2e; then \
-		$(KIND) create cluster -n i2gw-e2e --kubeconfig $(REPO_ROOT)/kind-kubeconfig --config ./e2e/kind.yaml; \
+		$(KIND) create cluster -n i2gw-e2e --kubeconfig $(REPO_ROOT)/kind-kubeconfig --config $(KIND_CONFIG); \
 	else \
 		echo "Cluster i2gw-e2e already exists. Reusing it."; \
 		$(KIND) get kubeconfig --name i2gw-e2e > $(REPO_ROOT)/kind-kubeconfig; \
@@ -146,3 +161,4 @@ e2e: build ## Run end-to-end tests.
 clean-kind:
 	$(KIND) delete cluster -n i2gw-e2e
 	rm -f $(REPO_ROOT)/kind-kubeconfig
+	rm -f $(KIND_CONFIG)
--- a/e2e/kind.yaml
+++ b/e2e/kind.yaml
@@ -1,7 +1,0 @@
-kind: Cluster
-apiVersion: kind.x-k8s.io/v1alpha4
-nodes:
-- role: control-plane
-- role: worker
-- role: worker
-- role: worker


e2e/provider_ingressnginx_test.go
```diff
@@ -0,0 +1,1093 @@
...
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/ingressnginx"
+	"github.com/kubernetes-sigs/ingress2gateway/pkg/i2gw/providers/istio"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+)
+
+// ingress-nginx provider features: One test per feature, all using Istio + standard emitter.
+
+func TestIngressNGINXCanary(t *testing.T) {
```
Stevenjin8, line 35: We should also add kgateway here and envoygateway when it merges. I'll let you decide the timing on that

Stevenjin8:
/approve

@kkk777-7 @johananl can you coordinate with this merge with #372? At first glance,  the conflicts are minimal or easy to fix.

e2e/README.md
```diff
@@ -50,18 +49,95 @@ I2GW_GO_TEST_ARGS="-race -v -count=1 -timeout=10m -run TestIngressNginx" make e2
...
+
+One test per provider, all using the Istio implementation and the standard emitter. Validates
+that each provider's core ingress → Gateway API conversion works.
+
+#### Category 2 — provider features (`provider_<name>_test.go`)
+
+One test per provider-specific annotation or feature using the Istio implementation and the
+standard emitter. Only the provider that owns the feature is used. Validates that the feature is
+correctly converted to Gateway API resources.
+
```
Stevenjin8, line 98: I think here we may want to also use the other gateway api implementations.

Stevenjin8: eh or maybe not, I see it both ways.

johananl: AFAICT, since provider-specific features get translated to an IR, the GWAPI implementation doesn't matter. Another way to put it: You can't tell which provider is the source just from looking at an IR object (right?) and therefore there is no meaning to (provider X implementation) pairs in the tests.

By separating the tests to two "legs" -- (ingress -> IR) and (IR -> GWAPI), we make the testing matrix simpler without reducing coverage. AFAICT this decoupling is precisely why you introduced emitters in the first place. We're just building on top of that decoupling here in the tests, too.

Am I missing anything? Are you noticing something specific which we aren't covering by this structure?

Stevenjin8: There is still a lot of implementation specific behavior even if we are only using standard gateway API. For example, does the `(?i)` regex flag work? Are regex matches full, or do they require explicit `^` and `$` anchors? Do redirects preserve trailing slashes? As long as we have some way of testing this I'm good. And even if we don't its probably fine.

johananl: Is this implementation-specific behavior dependent on a specific **provider**? In other words, is there any functionality which we can **only** test using e.g. ingress-nginx + Istio? If so, looks like we need a new test category, maybe something like "provider-implementation combinations". This is what I was hoping to avoid but maybe we do have a coverage gap here.

e2e/provider_ingressnginx_test.go
jgreeer, line 1: is it possible to have each annotation group test as a seperate file under a folder like tests/ingress-nginx? i get why we don't want all the files at the root e2e folder but having it under one really big file is a bit less convenient

johananl: Sub-dirs under e2e would break our concurrency because in Go the tests for each package run in a separate process, which means duplicating the ResourceManager singleton, which means trouble 🙂

Maybe you can think of a way to use multiple sub-packages and still have safe concurrent test cases. I'd love to see such a solution working.

johananl: BTW, we can definitely break down big files into smaller *files*, the issue is just with sub-directories because that equals sub-packages in Go.

johananl:
> /approve
> 
> @kkk777-7 @johananl can you coordinate with this merge with #372? At first glance, the conflicts are minimal or easy to fix.

I just reviewed #372. I think we can safely merge #372 as is and I'll rebase this PR and do the necessary adaptations.

e2e/provider_ingressnginx_test.go
jgreeer: ah i see, yeah that's more of my personal preference anyways. i don't really like the idea of having all the seperate files in the root directory so we can stick with this for now

johananl: Yeah, I'd argue that keeping the parallelism is worth way more than a somewhat nicer dir structure. Anyway, thanks for the feedback and we can refactor later in a new PR if we find some better arrangement.

johananl:
Found some small issues I need to fix.
/hold

e2e/provider_ingressnginx_test.go
```diff
@@ -0,0 +1,1211 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
kkk777-7: nits:
we should remove copyright years
Ref : https://github.com/kubernetes-sigs/ingress2gateway/pull/389

kkk777-7:
thanks for working on this excellent refactor!!
I only left a nits comment, but overall LGTM!

e2e/provider_ingressnginx_test.go
johananl: Done. Thanks.

johananl:
Ready for a re-review.
/unhold

Stevenjin8:
/approve

Will lgtm tomorrow unless someone objects

kkk777-7:
LGTM, thanks!
I'll leave the timing of the merge to Steven :)

PR #372, opened 2026-02-28
add envoy gateway e2e

kkk777-7:
**What type of PR is this?**

/kind test

Add Envoy Gateway E2E test.

Fixes #

```release-note
NONE
```

### DIFF
--- a/e2e/crd.go
+++ b/e2e/crd.go
@@ -36,11 +36,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
-const (
-	gatewayAPIVersion    = "v1.5.0"
-	gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/" + gatewayAPIVersion + "/experimental-install.yaml"
-)
-
 func deployCRDs(ctx context.Context, l logger, client *apiextensionsclientset.Clientset, url string, skipCleanup bool) (func(), error) {
 	l.Logf("Fetching manifests from %s", url)
 	yamlData, err := fetchManifests(ctx, l, url)
--- a/e2e/gwapi.go
+++ b/e2e/gwapi.go
@@ -32,6 +32,11 @@ import (
 	gwclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
 )
 
+const (
+	gatewayAPIVersion    = "v1.5.0"
+	gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/" + gatewayAPIVersion + "/experimental-install.yaml"
+)
+
 func createGatewayResources(
 	ctx context.Context,
 	l logger,


Stevenjin8:
/approve

cc @johananl

e2e/gwapi_envoygateway.go
```diff
@@ -0,0 +1,130 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gwclientset "sigs.k8s.io/gateway-api/pkg/client/clientset/versioned"
+)
+
+const (
+	envoyGatewayName          = "envoy-gateway"
+	envoyGatewayVersion       = "v0.0.0-latest"
```
Stevenjin8: is there a reason we are using latest instead of an actual release?

kkk777-7: Envoy Gateway Helm also installs the Gateway API CRD, which caused CRD version conflict issue.
There’s also an installation method where we apply the CRDs individually, so I’ll handle it using that approach.
https://gateway.envoyproxy.io/docs/install/install-helm/#installing-crds-separately

Stevenjin8: that's so annoying...

Stevenjin8:
c @johananl could you take  alook

e2e/gwapi_envoygateway.go
kkk777-7: I resolved it by installing only the EG CRDs separately :)

e2e/crd.go
```diff
@@ -41,11 +41,11 @@ const (
 	gatewayAPIInstallURL = "https://github.com/kubernetes-sigs/gateway-api/releases/download/" + gatewayAPIVersion + "/experimental-install.yaml"
 )
 
-func deployCRDs(ctx context.Context, l logger, client *apiextensionsclientset.Clientset, skipCleanup bool) (func(), error) {
-	l.Logf("Fetching manifests from %s", gatewayAPIInstallURL)
-	yamlData, err := fetchManifests(ctx, l)
+func deployCRDs(ctx context.Context, l logger, client *apiextensionsclientset.Clientset, url string, skipCleanup bool) (func(), error) {
```
johananl, line 39: I understand this change, there is a subtlety here: Before the change, this function was meant to be a helper fro installing the *GWAPI* CRDs specifically whereas now it's a *generic* helper for installing arbitrary CRDs.

What I suggest we do is move the `gatewayAPIVersion` and `gatewayAPIInstallURL` consts to `gwapi.go`. This way we turn `crd.go` into a truly generic helper file while keeping GWAPI-specific stuff in one place.

e2e/gwapi_envoygateway.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+	}
+
+	l.Logf("Creating GatewayClass %s", envoyGatewayName)
+	_, err = gwClient.GatewayV1().GatewayClasses().Create(ctx, gc, metav1.CreateOptions{})
+	if errors.IsAlreadyExists(err) {
+		_, err = gwClient.GatewayV1().GatewayClasses().Update(ctx, gc, metav1.UpdateOptions{})
+	}
+	if err != nil {
+		cleanupCRDs()
```
johananl, line 115: We're leaking the Helm chart here. This isn't a huge deal since we seem to have a *general* problem with this pattern (see #386) so please treat this comment as an FYI.

e2e/crd.go
kkk777-7: thanks, that makes sense! I've fixed.

e2e/gwapi_envoygateway.go
kkk777-7: thanks, good catch!
After https://github.com/kubernetes-sigs/ingress2gateway/pull/366 merged, I'll tackle this issue when I have time.

johananl: >I'll tackle this issue when I have time

Does "this issue" mean #386? Can we merge this (#372) PR?

kkk777-7: yeah, I meant https://github.com/kubernetes-sigs/ingress2gateway/issues/386. I’d appreciate it if we could merge PR #372 first.

johananl:
Merged! Thanks for the contribution @kkk777-7 🙏

PR #375, opened 2026-03-02
fix: use binary suffixes (Ki/Mi/Gi) for body size conversion

Stevenjin8:
For ingress nginx, `M` is `2^20` bytes while for kubernetes `M` is 10^6 while `Mi` is 2^20`.

**What type of PR is this?**
/kind bug

/kind bug

Fixes #

```release-note

```


kkk777-7:
LGTM, thanks!

/lgtm

PR #381, opened 2026-03-04
E2E for Backend TLS + Backend TLS fix

### LINKED ISSUE #328: E2E tests for backend TLS
Turns out, its quite common for pods to terminate TLS. We should have E2E tests for this.

**What happened**:

jgreeer:
**What type of PR is this?**

Adds E2E tests for backend TLS

Fixes #328 

```release-note
NONE
```

### DIFF
--- a/e2e/framework/dummy_app.go
+++ b/e2e/framework/dummy_app.go
@@ -84,13 +84,13 @@ func createDummyAppDeployment(ctx context.Context, l Logger, client *kubernetes.
 	l.Logf("Creating dummy app %s", name)
 
 	containerArgs := []string{"netexec", "--http-port=8080"}
-	portname := "http"
+	portName := "http"
 	if serverSecretName != "" {
 		containerArgs = append(containerArgs,
 			"--tls-cert-file=/etc/tls/tls.crt",
 			"--tls-private-key-file=/etc/tls/tls.key",
 		)
-		portname = "https"
+		portName = "https"
 	}
 
 	volumeMount := corev1.VolumeMount{
@@ -120,7 +120,7 @@ func createDummyAppDeployment(ctx context.Context, l Logger, client *kubernetes.
 							Args:  containerArgs,
 							Ports: []corev1.ContainerPort{
 								{
-									Name:          portname,
+									Name:          portName,
 									ContainerPort: 8080,
 								},
 							},
--- a/e2e/framework/secrets.go
+++ b/e2e/framework/secrets.go
@@ -41,13 +41,13 @@ type TLSTestSecret struct {
 }
 
 // BackendTLSSecrets holds the secrets needed for backend TLS authentication testing.
-// ServerSecret contains the TLS cert+key for the HTTPS backend pod.
-// CASecret contains the CA certificate used to verify the backend, referenced by the
-// proxy-ssl-secret annotation and later by BackendTLSPolicy.
 type BackendTLSSecrets struct {
-	ServerSecret *corev1.Secret // tls.crt + tls.key for the backend pod
-	CASecret     *corev1.Secret // ca.crt for the proxy-ssl-secret / BackendTLSPolicy
-	CACertPEM    []byte
+	// ServerSecret contains the TLS cert+key for the HTTPS backend pod.
+	ServerSecret *corev1.Secret
+	// CASecret contains the CA certificate used to verify the backend, referenced by the
+	// proxy-ssl-secret annotation and later by BackendTLSPolicy.
+	CASecret  *corev1.Secret // ca.crt for the proxy-ssl-secret / BackendTLSPolicy
+	CACertPEM []byte
 }
 
 // GenerateSelfSignedTLSSecret creates a self-signed TLS secret for testing.
@@ -153,12 +153,8 @@ const (
 	BackendCASecretName     = "tls-backend-ca"          //nolint:gosec // Not a credential, just a resource name.
 )
 
-// GenerateBackendTLSSecrets creates a CA and a server certificate signed by that CA.
-// It returns:
-//   - ServerSecret: a TLS secret (tls.crt + tls.key) to mount in the HTTPS backend pod
-//   - CASecret: an Opaque secret with ca.crt for the proxy-ssl-secret annotation / BackendTLSPolicy
-//
-// The serverHostname is the DNS name the server cert is valid for (used as SAN and for proxy-ssl-name).
+// GenerateBackendTLSSecrets creates a self-signed CA and a server certificate
+// signed by that CA, returning them as Kubernetes TLS and Opaque secrets.
 func GenerateBackendTLSSecrets(serverSecretName, caSecretName, namespace, serverHostname string) (*BackendTLSSecrets, error) {
 	// Generate CA key and self-signed CA cert.
 	caKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)


jgreeer:
This PR also contains a change to backend_tls.go. Instead of using a Secret we use a ConfigMap, Secrets in the `policy.Spec.Validation.CACertificateRefs` field are not supported by istio and are not in core support. From [the reference:](https://gateway-api.sigs.k8s.io/reference/spec/#backendtlspolicyvalidation)
```
Support: Core - An optional single reference to a Kubernetes ConfigMap,
with the CA certificate in a key named ca.crt.
```

e2e/framework/dummy_app.go
```diff
@@ -161,3 +161,133 @@ func waitForDummyApp(ctx context.Context, l logger, client *kubernetes.Clientset
 
 	return nil
 }
+
+// deployDummyTLSApp deploys an agnhost netexec app that serves HTTPS using the provided TLS
+// secret. The secret must contain tls.crt and tls.key. The service exposes port 443 targeting
+// the container's HTTPS port (8443).
+func deployDummyTLSApp(ctx context.Context, l logger, client *kubernetes.Clientset, name, namespace, tlsSecretName string, skipCleanup bool) (func(), error) {
```
johananl: This adds a lot of duplicate code. Have you thought about reusing `deployDummyApp()` and adding a new argument which controls the TLS logic?

e2e/ingress_nginx_backend_tls_test.go
```diff
@@ -0,0 +1,260 @@
...
+// backendTLSSetup deploys an HTTPS backend app using a CA-signed server certificate
+// and patches the given ingresses so that namespace-dependent annotations (proxy-ssl-secret
+// and proxy-ssl-name) contain the correct dynamic values.
+//
+// ingress-nginx requires proxy-ssl-secret in "namespace/secretName" format and
+// proxy-ssl-name must match the server certificate SAN, which uses the dynamic test namespace.
+// Because the namespace is only known at runtime (inside runTestCase), the ingresses are built
+// with placeholder values and patched here before they are created in the cluster.
+func backendTLSSetup(ingresses ...*networkingv1.Ingress) func(ctx context.Context, t *testing.T, client *kubernetes.Clientset, namespace string, skipCleanup bool) {
+	return func(ctx context.Context, t *testing.T, client *kubernetes.Clientset, namespace string, skipCleanup bool) {
```
johananl: I'm not a fan of this approach. My intention was to keep the e2e test case definition **declarative** and minimal, with the bulk of the tricky logic living as **generic code** under functions such as `runTestCase()`. I'd try to avoid adding a `setup()` function which can contain arbitrary code in the test config as I think it will make the tests less maintainable.

I'm trying to figure out if there's some **generic framework functionality** that you're missing here. If so, we should extend the framework rather than add "special setup magic" for one test case.

johananl: OK, I think I see what the issue is: You need access to intermediate objects created inside `runTestCase()` but `runTestCase()` is a monolith which doesn't expose what you need. I think we can solve this problem without resorting to arbitrary setup functions. I'll add a general comment on the PR.

e2e/framework/dummy_app.go
Stevenjin8: What about always creating doing this for dummy-app1 and dummy-app2?

Stevenjin8: or just always have dummy app have tls.

e2e/ingress_nginx_backend_tls_test.go
Stevenjin8: If we hardcode the config map and always create them, we should be fine then? much like we hardcode the name of dummy-app services.

e2e/ingress_nginx_backend_tls_test.go
```diff
@@ -0,0 +1,260 @@
...
+		// Gateway API implementations like Istio). The ConfigMap name matches the
+		// CA secret name so proxy-ssl-secret annotation value maps directly.
+		caCM := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      backendCASecretName,
+				Namespace: namespace,
+			},
+			Data: map[string]string{
+				"ca.crt": string(secrets.CACertPEM),
+			},
```
Stevenjin8: sigh... I didn't realize that Ingress-NGINX uses a secret while gateway api uses a config map. Generally, I don't want to touch secrets.

e2e/framework/dummy_app.go
jgreeer: if dummy app always has tls then the normal http tests wouldn't be able to connect to it?

e2e/ingress_nginx_backend_tls_test.go
jgreeer: yeah don't like it at all but only configMap is core support

e2e/framework/dummy_app.go
jgreeer: oh wait have both ports open right

pkg/i2gw/providers/ingressnginx/backend_tls.go
```diff
@@ -174,8 +174,8 @@ func backendTLSFeature(ingresses []networkingv1.Ingress, _ map[types.NamespacedN
 				// We know proxySSLVerify is "on" and proxySSLSecret is not empty due to strict validation above.
 				policy.Spec.Validation.CACertificateRefs = []gatewayv1.LocalObjectReference{{
 					Group: "",
-					Kind:  "Secret",
-					Name:  gatewayv1.ObjectName(secretName),
+					Kind:  "ConfigMap",
+					Name:  gatewayv1.ObjectName(caRefName),
 				}}
 				policy.Spec.Validation.WellKnownCACertificates = nil
```
Stevenjin8, line 193: @jgreeer can you add logs here that say

1. mTLS will not be configured 
2. you need to create a config map with the same name and the `ca.crt` key (but in better english)

Stevenjin8:
@jgreer I'm guessing this is ready for review?

jgreeer:
> @jgreer I'm guessing this is ready for review?

trying to refactor it so we don't have the setup function

Stevenjin8:
@jgreeer sorry about that.

Stevenjin8:
@jgreer any chance we could have this ready to review by today?

jgreeer:
> @jgreer any chance we could have this ready to review by today?

yep, just finishing some things up should be ready by today

jgreeer:
Made some changes to the framework

- Added a configMap argument to `framework.testCase`
- dummyApp now has a useTLS argument instead of duplicate functions
- Broke `runTestCase` into two functions: `SetupTestEnv` and `env.Run()`, following the suggestion by @johananl.
- Since this division really isn't needed in other parts of the e2e so far, `runTestCase` is now a wrapper that calls both of those functions, so no changes for other e2e tests

e2e/framework/dummy_app.go
```diff
@@ -72,11 +85,27 @@ func deployDummyApp(ctx context.Context, l Logger, client *kubernetes.Clientset,
...
 	l.Logf("Creating dummy app %s", name)
 
+	containerArgs := []string{"netexec", "--http-port=8080"}
+	portname := "http"
+	if useTLS {
+		containerArgs = append(containerArgs,
+			"--tls-cert-file=/etc/tls/tls.crt",
+			"--tls-private-key-file=/etc/tls/tls.key",
+		)
+		portname = "https"
```
johananl: Nit: Doesn't follow the Go style for local vars (should be `portName`).

e2e/framework/secrets.go
```diff
@@ -40,6 +40,16 @@ type TLSTestSecret struct {
 	CACert []byte
 }
 
+// BackendTLSSecrets holds the secrets needed for backend TLS authentication testing.
+// ServerSecret contains the TLS cert+key for the HTTPS backend pod.
```
johananl: Nit: Isn't it better to put field-specific comments directly on the relevant fields?

e2e/framework/secrets.go
```diff
@@ -137,3 +147,101 @@ func createSecrets(ctx context.Context, l Logger, client *kubernetes.Clientset,
 		}
 	}, nil
 }
+
+// GenerateBackendTLSSecrets creates a CA and a server certificate signed by that CA.
+// It returns:
```
johananl: I find this misleading. The function returns a `*BackendTLSSecrets`. IMO info about the sub-fields should be documented on `BackendTLSSecrets` rather than here.

johananl:
Thanks @jgreeer! IMO this looks a lot better. However, we're mixing TLS bits with the general framework. I think we should keep the framework as generic as possible and keep details belonging to a small number of test cases in the individual tests.

Here is a (vibe-coded) example of how we can use TLS without "leaking" test-specific stuff into central pieces such as `TestEnv`: https://github.com/johananl/ingress2gateway/tree/ingress-nginx/backend-tls-e2e-tests-mod1

It continues the refactor you started but takes the separation of concerns further. It does so by allowing each test case to specify an arbitrary set of **backends**, with or without TLS. The default behavior is 2 non-TLS backends like we had before this PR so no added boilerplate.

Let me know what you think and feel free to use anything from this branch if you find it helpful.

P. S. I'm aware it can be disorienting that I'm proposing complete change sets rather than commenting in the PR. I'm doing this because it's easier for me to show my intention in code rather than describe it verbally, plus vibe-coding is so easy now that I can actually verify what I'm suggesting works before sending you on a wild goose chase just to hit a problem I overlooked.

jgreeer:
> Thanks @jgreeer! IMO this looks a lot better. However, we're mixing TLS bits with the general framework. I think we should keep the framework as generic as possible and keep details belonging to a small number of test cases in the individual tests.
> 
> Here is a (vibe-coded) example of how we can use TLS without "leaking" test-specific stuff into central pieces such as `TestEnv`: https://github.com/johananl/ingress2gateway/tree/ingress-nginx/backend-tls-e2e-tests-mod1
> 
> It continues the refactor you started but takes the separation of concerns further. It does so by allowing each test case to specify an arbitrary set of **backends**, with or without TLS. The default behavior is 2 non-TLS backends like we had before this PR so no added boilerplate.
> 
> Let me know what you think and feel free to use anything from this branch if you find it helpful.
> 
> P. S. I'm aware it can be disorienting that I'm proposing complete change sets rather than commenting in the PR. I'm doing this because it's easier for me to show my intention in code rather than describe it verbally, plus vibe-coding is so easy now that I can actually verify what I'm suggesting works before sending you on a wild goose chase just to hit a problem I overlooked.

Thanks for the feedback, I agree with the backend suggestion and added your changes to the branch. I only changed the default backend to be one instead of two since only the Canary E2E needs two. Additionally, I decided to remove `runTestCase` and require all tests to create and environment and call `.run()` on it. This is because `framework.TestCase`  still had two variables (providers and gatewayImplementation) that are actually used in `setupTestEnv()`. Now everything is cleanly seperated into the two structs. If we want to keep runTestCase, I think we could maybe have it take two arguments (envrionment config and a testCase). Let me know if you think we should do that instead.

johananl:
> Thanks for the feedback, I agree with the backend suggestion and added your changes to the branch. I only changed the default backend to be one instead of two since only the Canary E2E needs two. Additionally, I decided to remove `runTestCase` and require all tests to create and environment and call `.run()` on it. This is because `framework.TestCase` still had two variables (providers and gatewayImplementation) that are actually used in `setupTestEnv()`. Now everything is cleanly seperated into the two structs. If we want to keep runTestCase, I think we could maybe have it take two arguments (envrionment config and a testCase). Let me know if you think we should do that instead.

IMO providers and the gateway implementation should remain under the `TestCase` struct because they are part of the input for each test case and are a critical part of what we're testing (i.e. "given provider X and implementation Y, does this conversion work?").

Also, it certainly doesn't make sense to me to separate `Providers` and `ProviderFlags` to two different structs since they are closely related.

I would do the following:

```go
// TestCase defines the configuration for a single e2e test case.
type TestCase struct {
	// Backends to deploy. When nil, defaults to DummyAppName1 + DummyAppName2
	// (both plain HTTP).
	Backends               []Backend
	Ingresses              []*networkingv1.Ingress
	Secrets                []*corev1.Secret
	ConfigMaps             []*corev1.ConfigMap
	Providers              []string
	ProviderFlags          map[string]map[string]string
	GatewayImplementation  string
	AllowExperimentalGWAPI bool
	Emitter                string
	Verifiers              map[string][]Verifier
}
```

You can then keep `runTestCase()` as an optional convenience function that's used by default unless the test case needs more granularity:

```go
// runTestCase is a convenience wrapper for tests that don't need custom setup between environment
// creation and test execution.
func runTestCase(t *testing.T, tc *framework.TestCase) {
	env := setupTestEnv(t, tc.Providers, tc.GatewayImplementation)
	env.Run(tc)
}
```

This keeps the test cases more concise:

```go
				t.Run("basic conversion", func(t *testing.T) {
					runTestCase(t, &framework.TestCase{
						Providers:             []string{prov.name},
						GatewayImplementation: implementation.IstioName,
						ProviderFlags:         prov.providerFlags,
						Ingresses: []*networkingv1.Ingress{
							framework.BasicIngress().
								WithName("foo").
								WithIngressClass(prov.ingressClass).
								Build(),
						},
						Verifiers: map[string][]framework.Verifier{
							"foo": {&framework.HTTPRequestVerifier{Path: "/"}},
						},
					})
				})
```

And it keeps `TestCase` as a formal structure for **all** inputs to a test case while keeping `TestEnv` as a more low-level plumbing element that _can_ be manipulated directly if necessary but doesn't _have_ to be.

Another way to look at it: `TestCase` is the formal definition of a test case. We may **copy** some data from `TestCase` to `TestEnv` (e.g. which providers or implementation to deploy), but `TestCase` remains authoritative. `TestCase` defines the test whereas `TestEnv` is an "infra" struct.

So, if we accept the above, then for example when calling the i2gw binary in `env.Run()` we should use `tc.Providers` as an argument to i2gw, **not** `env.Providers` even if the two values are the same. This is to keep the logical separation of concerns.

Does that make sense?

PR #384, opened 2026-03-05
Clean up notifications

Stevenjin8:
I think we have a lot of notifications that are really just debug logs that we should not present to the user.

Also, most of our errors are recoverable (say, by falling back to a default value), so we should just make an Error notification instead and continue on.

This also adds some other notifications

* GRPC isn't fully supported
* We wont serve traffic using a self-signed cert
* regex matches are case insenstive and prefix
* path normalization isn't super well defined.
 
 ~

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/iprange.go
+++ b/pkg/i2gw/providers/ingressnginx/iprange.go
@@ -73,7 +73,7 @@ func (p *Provider) applyIPRangeControlToEmitterIR(pIR providerir.ProviderIR, eIR
 				DenyList:  denyList,
 			}
 			{
-				source := fmt.Sprintf("%s and %s from %s/%s", WhiteListSourceRangeAnnotation, DenyListSourceRangeAnnotation, ing.Namespace, ing.Name)
+				source := fmt.Sprintf("%s/%s", ing.Namespace, ing.Name)
 				message := "IP-based authorization is not supported"
 				paths := make([]*field.Path, 0, 2)
 				if len(allowList) > 0 {


pkg/i2gw/providers/ingressnginx/canary.go
```diff
@@ -126,15 +136,13 @@ func canaryFeature(notify notifications.NotifyFunc, ingresses []networkingv1.Ing
...
 				existingMatches := httpRouteContext.HTTPRoute.Spec.Rules[ruleIdx].Matches
-				canaryWeight, nonCanaryWeight, config, canarySourceIngress, canaryBackendIdx, nonCanaryBackendIdx, parseErrs := getCanaryInfo(notify, backendSources, "httproute", httpRouteContext.HTTPRoute.Name, ruleIdx)
+				canaryWeight, nonCanaryWeight, config, canaryBackendIdx, nonCanaryBackendIdx, parseErrs := getCanaryInfo(notify, backendSources, "httproute", httpRouteContext.HTTPRoute.Name, ruleIdx)
 				errList = append(errList, parseErrs...)
 				if canaryBackendIdx != -1 && nonCanaryBackendIdx != -1 {
 					// Set weights if isWeight is true or both header and weight are not set (all traffic should go to non-canary)
 					if config.isWeight || !config.isHeader {
 						httpRouteContext.HTTPRoute.Spec.Rules[ruleIdx].BackendRefs[canaryBackendIdx].Weight = &canaryWeight
 						httpRouteContext.HTTPRoute.Spec.Rules[ruleIdx].BackendRefs[nonCanaryBackendIdx].Weight = &nonCanaryWeight
-						notify(notifications.InfoNotification, fmt.Sprintf("parsed canary annotations of ingress %s/%s and set weights (canary: %d, non-canary: %d, total: %d)",
```
Stevenjin8, line 136: IMO notifications liek these ones don't really add anything. The user already expects ingress2gateway to do this.

pkg/i2gw/providers/ingressnginx/headers.go
```diff
@@ -37,7 +37,6 @@ func headerModifierFeature(notify notifications.NotifyFunc, _ []networkingv1.Ing
 
 			ingress := getNonCanaryIngress(sources)
 			if ingress == nil {
-				notify(notifications.InfoNotification, "Found canary ingress rule without non-canary ingress rule", &httpRouteContext.HTTPRoute)
```
Stevenjin8, line 40: maybe this is an error? In theory we should never get to this code path.

pkg/i2gw/providers/ingressnginx/canary.go
jgreeer: i agree, we should make sure the README is update to date though in that case. otherwise the user would have to look through the translated output and confirm the annotation was parsed. i guess they would know by the fact that there wasn't an unsupported annotation notification but that's a bit less intuitive.

pkg/i2gw/emitters/standard/standard.go
```diff
@@ -42,8 +42,33 @@ func NewEmitter(conf *i2gw.EmitterConf) i2gw.Emitter {
...
+				}
+			}
+		}
+		if hasRegex && hasNonRegex {
+			e.notify(notifications.WarningNotification, "HTTPRoute contains both regex and non-regex path matches, which have implementation-specific priorities", &route)
+		}
+	}
+
+}
+
```
Stevenjin8:
```suggestion
This cannot be true for a host.
```

Stevenjin8: or at least not with ingress nginx

pkg/i2gw/providers/ingressnginx/iprange.go
```diff
@@ -74,14 +73,14 @@ func applyIPRangeControlToEmitterIR(notify notifications.NotifyFunc, pIR provide
 				DenyList:  denyList,
 			}
 			{
-				source := fmt.Sprintf("%s/%s", ing.Namespace, ing.Name)
+				source := fmt.Sprintf("%s and %s from %s/%s", WhiteListSourceRangeAnnotation, DenyListSourceRangeAnnotation, ing.Namespace, ing.Name)
```
kkk777-7: Do we need annotation info in this? setting paths contains annotation info.

output contains duplicate information.
```
┌─ WARN  ────────────────────────────────────────
│  Failed to apply namespace1.iprange-ingress.metadata.annotations."nginx.ingress.kubernetes.io/whitelist-source-range" from nginx.ingress.kubernetes.io/whitelist-source-range and nginx.ingress.kubernetes.io/denylist-source-range from namespace1/iprange-ingress: IP-based authorization is not supported
│  source: STANDARD_EMITTER
│  object: HTTPRoute: namespace1/iprange-ingress-example-com
```

pkg/i2gw/providers/ingressnginx/redirect.go
```diff
@@ -240,15 +225,7 @@ func addDefaultSSLRedirect(pir *providerir.ProviderIR, eir *emitterir.EmitterIR)
...
-				parsed, err := strconv.ParseBool(val)
-				if err != nil {
-					return field.ErrorList{field.Invalid(
-						field.NewPath("ingress", ingress.Namespace, ingress.Name, "metadata", "annotations"),
-						ingress.Annotations,
-						fmt.Sprintf("failed to parse canary configuration: %v", err),
-					)}
-				}
-				enableRedirect = parsed
+				enableRedirect, _ = strconv.ParseBool(val)
```
kkk777-7, line 228: should we notify invalid values with a warning here?

kkk777-7:
thanks @Stevenjin8 ! I left some comments.

pkg/i2gw/providers/ingressnginx/redirect.go
Stevenjin8: Its something I wrestle with. In our existing code, we dont

https://github.com/Stevenjin8/ingress2gateway/blob/e26d9c1f1b986f202440e228c5b0575a6ae6f1d9/pkg/i2gw/providers/ingressnginx/regex.go#L36

https://github.com/Stevenjin8/ingress2gateway/blob/e26d9c1f1b986f202440e228c5b0575a6ae6f1d9/pkg/i2gw/providers/ingressnginx/utils.go#L32

and in general, I think we can assume that the configuration we receive is valid ingress-nginx. In either case, we are falling back to the default value.

kkk777-7: > I think we can assume that the configuration we receive is valid ingress-nginx.

That makes sense!

kkk777-7 (approved):

kkk777-7:
LGTM, thanks!
/lgtm

PR #397, opened 2026-03-24
Move http route rule naming from common provider to emitters

jgreeer:
**What type of PR is this?**

We were adding names to HTTPRoute rules in the common provider despite it being an Extended support field. This PR fixes this by moving the logic to a helper function that emitters can call. This also fixes an issue with the new rules created by the ingress-nginx feature parsers not having names, since we now add the names to the rules after at the emitter stage.

Fixes #

```release-note
Adding names to HTTPRoute rules only happens in supported emitters
```


kkk777-7 (approved):

kkk777-7:
LGTM, thanks!
wait for Steven’s confirmation before merging

Stevenjin8 (commented):
/lgtm

jgreeer:
@Stevenjin8 should be good now

PR #401, opened 2026-03-31
fix: keep PathPrefix when rewrite-target is static path

puertomontt:
**What type of PR is this?**

/kind bug

fix bug in regex logic. if rewrite-target is static path should not treat as regex.

Fixes #

```release-note
fix: keep PathPrefix when rewrite-target is static path
```

### DIFF
--- a/pkg/i2gw/providers/ingressnginx/regex.go
+++ b/pkg/i2gw/providers/ingressnginx/regex.go
@@ -28,6 +28,18 @@ import (
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// pathContainsRegex returns true if the path contains regex metacharacters,
+// indicating it should be treated as a regular expression.
+func pathContainsRegex(path string) bool {
+	for _, c := range path {
+		switch c {
+		case '.', '+', '*', '?', '^', '$', '{', '}', '(', ')', '[', ']', '|', '\\':
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


pkg/i2gw/providers/ingressnginx/regex.go
```diff
@@ -28,6 +28,18 @@ import (
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// pathContainsRegex returns true if the path contains regex metacharacters,
+// indicating it should be treated as a regular expression.
+func pathContainsRegex(path string) bool {
```
jgreeer, line 33: do we need this function? couldn't we try to compile as regex and use prefix match if it fails?

puertomontt: a plain path wouldn't fail to compile

jgreeer:
the example you linked is from NGINX ingress not ingress-nginx. from the [ingress-nginx documentation:](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#use-regex)

> Additionally, if the [rewrite-target annotation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#rewrite) is used on any Ingress for a given host, then the case insensitive regular expression [location modifier](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) will be enforced on ALL paths for a given host regardless of what Ingress they are defined on.

so i think the original behaviour here is correct

Stevenjin8:
+1 to @jgreeer

PR #402, opened 2026-03-31
feat: ingress-nginx ssl passthrough annotation

jgreeer:
**What type of PR is this?**

Fixes #396 

```release-note
add support for `nginx.ingress.kubernetes.io/ssl-passthrough`
```


jgreeer:
@Stevenjin8 could you ok to test

jgreeer:
`scenarios_test.go` made it in from the www redirect PR & had a failing test. it's not mentioned anywhere in the e2e README + had duplicate tests so i deleted it in this PR

johananl:
> `scenarios_test.go` made it in from the www redirect PR & had a failing test. it's not mentioned anywhere in the e2e README + had duplicate tests so i deleted it in this PR

Yes, this file is outdated and should be removed. We have a more granular test hierarchy now.

e2e/implementation/istio.go
```diff
@@ -74,6 +74,12 @@ func DeployIstio(ctx context.Context,
 		"global": map[string]interface{}{
 			"istioNamespace": namespace,
 		},
+		// istio currently blocks attaching tlsroute to a gateway unless we have this env var set
```
johananl: Nit: Go comment style.

e2e/provider_ingressnginx_test.go
```diff
@@ -1451,3 +1451,128 @@ func TestIngressNGINXRegex(t *testing.T) {
 		})
 	})
 }
+
+func TestIngressNGINXSSLPassthrough(t *testing.T) {
+	t.Parallel()
+	t.Run("to Istio", func(t *testing.T) {
```
johananl: "to Istio" is a remnant from the older e2e setup. We should remove it since it doesn't add meaningful context regarding the test.

If you remove this, please also remove `t.Parallel()` on the following line.

johananl (commented):
Thanks. Left a comment and a nit.

pkg/i2gw/providers/ingressnginx/ssl_passthrough.go
```diff
@@ -0,0 +1,264 @@
...
+				}
+				ingKey := types.NamespacedName{Namespace: src.Ingress.Namespace, Name: src.Ingress.Name}
+				if _, ok := passthroughIngresses[ingKey]; ok {
+					hasPassthrough = true
+				} else {
+					hasNonPassthrough = true
+				}
+			}
+		}
+
```
Stevenjin8, line 80: I think around here, it would be much cleaner if we iterated over the ingresses and found all the hosts with ssl-passthrough. But i guess here you can let people know of rules that might unexpectedly get dropped. Both work.

Stevenjin8:
@jgreeer sorry it took me a sec to get to this. Can you add a test where you have both both ssl passthrough hosts and non-ssl passthrough host and check that if the mixed termination. I'm guessing the mixed termination modes works in ingress nginx, but unclear whether it works with istio. If that's the case, we also need to avoid outputting mixed termination gwapi.

jgreeer:
> @jgreeer sorry it took me a sec to get to this. Can you add a test where you have both both ssl passthrough hosts and non-ssl passthrough host and check that if the mixed termination. I'm guessing the mixed termination modes works in ingress nginx, but unclear whether it works with istio. If that's the case, we also need to avoid outputting mixed termination gwapi.

added it, looks like it does work with istio

pkg/i2gw/providers/ingressnginx/ssl_passthrough.go
jgreeer: do you mean checking if a httproute is attached to a gateway with a "passthrough" hostname instead of iterating through its ruleBackendSources to decide whether we delete it or not?

jgreeer:
We actually shouldn't delete the httpRoute, we can pick the rules sourced from the passthrough ingress off the original httpRoute and keep all of the L7 rules at port 80

pkg/i2gw/providers/ingressnginx/ssl_passthrough.go
Stevenjin8: yeah, but this is more of a minor nit. NBD

Stevenjin8:
/approve
plan to lgtm by lazy concensus at the end of the week.

PR #409, opened 2026-05-01
Istio - add support for HTTPCORSFilter

### LINKED ISSUE #408: [Istio] Support for `HTTPCORSFilter`
Istio `VirtualService.http.route` has a [corsPolicy](https://istio.io/latest/docs/reference/config/networking/virtual-service/#CorsPolicy) field that maps to [HTTPCORSFilter](https://gateway-api.sigs.k8s.io/reference/1.5/spec/#httpcorsfilter).

`ingress2gateway` should be able to convert a `VirtualService` with `http.route.corsPolicy` to an `HTTPRoute` with `HTTPCORSFilter`.

**Why this is needed**:
Istio supports `HTTPCORSFilter` but currently this field is ignored by `ingress2gateway`:
```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: bookinfo-vs
  namespace: bookinfo
spec:
  gateways:
  - bookinfo-gateway
  hosts:
  - bookinfo.example.com
  http:
  - name: productpage
    corsPolicy:
      allowOrigins:
      - exact: https://bookinfo.example.com
      allowMethods:
      - GET
      - POST
      allowHeaders:
      - x-custom-header
      maxAge: 24h
    match:
    - uri:
        prefix: /productpage
    route:
    - destination:
        host: productpage.bookinfo.svc.cluster.local
        port:
          number: 9080
```
```
ingress2gateway print --input-file vs.yaml --providers istio
I0430 15:09:29.805640  379018 converter.go:600] ignoring field: istio.VirtualService[/cors].Http[productpage].CorsPolicy
┌─ INFO  ────────────────────────────────────────
│  ignoring field: istio.VirtualService[/cors].Http[productpage].CorsPolicy
│  source: ISTIO
│  object: VirtualService: /cors
└─
...
kind: HTTPRoute
metadata:
  annotations:
  name: cors-productpage
spec:
  rules:
  - backendRefs:
    - name: productpage
      namespace: bookinfo
      port: 9080
      weight: 0
    matches:
    - path:
        type: PathPrefix
        value: /productpage
status:
  parents: null
```

nrfox:
**What type of PR is this?**
/kind feature

Adds support for converting VirtualService corsPolicy --> HTTPCORSFilter.

Fixes #408 

```release-note
Istio - Add support for converting `corsPolicy` to `HTTPCORSFilter` for `VirtualService`.
```

### DIFF
--- a/pkg/i2gw/providers/istio/converter.go
+++ b/pkg/i2gw/providers/istio/converter.go
@@ -19,6 +19,7 @@ package istio
 import (
 	"context"
 	"fmt"
+	"math"
 	"net"
 	"regexp"
 	"strings"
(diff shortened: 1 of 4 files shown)


jgreeer:
I think it might be a good idea to add some e2e tests for this by creating a new `provider_istio_test.go` file under the `e2e` folder. Right now we're not running it through gateway api validation, when i tried to apply the translated yaml from the fixtures i got this:
```
The HTTPRoute "cors-all-fields" is invalid: spec.rules[0].filters[0].cors.allowOrigins[1]: Invalid value: "https://dashboard.*": spec.rules[0].filters[0].cors.allowOrigins[1] in body should match '(^\*$)|(^(http(s)?):\/\/(((\*\.)?([a-zA-Z0-9\-]+\.)*[a-zA-Z0-9-]+|\*)(:([0-9]{1,5}))?)$)'
```

pkg/i2gw/providers/istio/converter.go
```diff
@@ -595,9 +596,56 @@ func (c *resourcesToIRConverter) convertVsHTTPRoutes(virtualService metav1.Objec
...
+			}
+
+			for _, header := range corsPolicy.GetExposeHeaders() {
+				corsFilter.ExposeHeaders = append(corsFilter.ExposeHeaders, gatewayv1.HTTPHeaderName(header))
+			}
+
+			if maxAge := corsPolicy.GetMaxAge(); maxAge != nil {
+				// maxAge in corsPolicy is an int64 but int32 in corsFilter.
+				// Just in case it's too large, clamp it to math.MaxInt32.
+				corsFilter.MaxAge = int32(min(maxAge.Seconds, math.MaxInt32))
```
jgreeer, line 633: should we check if this is less than or equal to zero too?

nrfox: Yes on the Istio side this is a `time.Duration` which can be negative. There's [some validation on the VirtualService CRD](https://github.com/istio/istio/blob/3fc2be5a9cd6117f2344cd803e498bee773123f8/manifests/charts/base/files/crd-all.gen.yaml#L11017-L11023) that would prevent setting this to a negative value but it wouldn't hurt to check. What does `ingress2gateway` typically do for invalid values of fields? Log a warning? Should we also log an info message if we clamped the value?

nrfox:
> I think it might be a good idea to add some e2e tests for this by creating a new `provider_istio_test.go` file under the `e2e` folder. Right now we're not running it through gateway api validation, when i tried to apply the translated yaml from the fixtures i got this:
> 
> ```
> The HTTPRoute "cors-all-fields" is invalid: spec.rules[0].filters[0].cors.allowOrigins[1]: Invalid value: "https://dashboard.*": spec.rules[0].filters[0].cors.allowOrigins[1] in body should match '(^\*$)|(^(http(s)?):\/\/(((\*\.)?([a-zA-Z0-9\-]+\.)*[a-zA-Z0-9-]+|\*)(:([0-9]{1,5}))?)$)'
> ```

Sorry I should've tested all the output fixtures myself. Looking at this more carefully I wonder if translating `prefix` based matching makes any sense. I originally based it off of how [Istio converts](https://github.com/istio/istio/blob/33becd758cda14196f56c89f79fa90790d02271f/pilot/pkg/config/kube/gateway/conversion.go#L1324-L1333) from GW API --> VS but just adding a `*` to the end won't work here. I'll have another look at the [testdata](https://github.com/istio/istio/blob/279796ffaa2294d8b2ca511deafafc979e4bbb70/pilot/pkg/config/kube/gateway/testdata/http.yaml.golden#L141) and see what makes sense.

PR #410, opened 2026-05-13
feat(notifications): adds Report.Notifications() accessor grouped by source

marshallford:
**What type of PR is this?**
/kind feature

Adds `notifications.Report.Notifications()`. 

Currently the only way to read notifications off a Report is `Render()`, which returns a human-formatted string for CLI use. Go programs importing ingress2gateway as a module (calling `provider.ToIR()` or `i2gw.ToGatewayAPIResources()` directly) have no way to inspect notifications without parsing that string.

The accessor returns a defensive copy and is safe to call concurrently with Add. No changes to existing public API; `Render()` output is unchanged.

An alternative was to add a `Source` field to `Notification` itself for flat iteration, but that left `Add(source, n)` with two ways to set the source -- fixing that cleanly would have meant unexporting `Add` and forcing all producers through `Notifier(source)`. The grouped map keeps `Notification` source-agnostic and avoids the question.

N/A

No CLI-facing changes. Open to feedback on whether changes to the Go API surface warrant a release note.

### DIFF
--- a/pkg/i2gw/notifications/report.go
+++ b/pkg/i2gw/notifications/report.go
@@ -107,14 +107,12 @@ func (r *Report) Notifier(source string) NotifyFunc {
 // after conversion completes. Notifications are sorted by source name for deterministic output.
 // Returns "" when r is nil or there are no notifications.
 func (r *Report) Render() string {
-	if r == nil {
+	notifications := r.Notifications()
+	if len(notifications) == 0 {
 		return ""
 	}
 
-	r.mu.Lock()
-	defer r.mu.Unlock()
-
-	sources := slices.Sorted(maps.Keys(r.notifications))
+	sources := slices.Sorted(maps.Keys(notifications))
 
 	// Returns the ANSI code, or "" when color is disabled.
 	c := func(code string) string {
@@ -127,7 +125,7 @@ func (r *Report) Render() string {
 	var buf strings.Builder
 
 	for _, source := range sources {
-		for _, n := range r.notifications[source] {
+		for _, n := range notifications[source] {
 			label, lcolor := levelLabel(n.Type)
 
 			// Top border with level.
--- a/pkg/i2gw/notifications/report.go
+++ b/pkg/i2gw/notifications/report.go
@@ -68,6 +68,29 @@ func (r *Report) Add(source string, n Notification) {
 	r.mu.Unlock()
 }
 
+func (r *Report) Notifications() map[string][]Notification {
+	if r == nil {
+		return nil
+	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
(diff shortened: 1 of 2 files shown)


johananl:
Thanks @marshallford. Code looks OK to me. I'm curious about the use case though: Since you're adding `Notifications()` without using it anywhere, I assume you already are consuming ingress2gateway as a library?

johananl:
Another comment: Since we now have `Notifications()` as a way to get notifications safely, we might as well use it in `Render()` instead of having bespoke retrieval logic in `Render()`:

```diff
diff --git a/pkg/i2gw/notifications/report.go b/pkg/i2gw/notifications/report.go
index eb9f8d3..dc59464 100644
--- a/pkg/i2gw/notifications/report.go
+++ b/pkg/i2gw/notifications/report.go
@@ -107,14 +107,12 @@ func (r *Report) Notifier(source string) NotifyFunc {
 // after conversion completes. Notifications are sorted by source name for deterministic output.
 // Returns "" when r is nil or there are no notifications.
 func (r *Report) Render() string {
-       if r == nil {
+       notifications := r.Notifications()
+       if len(notifications) == 0 {
                return ""
        }

-       r.mu.Lock()
-       defer r.mu.Unlock()
-
-       sources := slices.Sorted(maps.Keys(r.notifications))
+       sources := slices.Sorted(maps.Keys(notifications))

        // Returns the ANSI code, or "" when color is disabled.
        c := func(code string) string {
@@ -127,7 +125,7 @@ func (r *Report) Render() string {
        var buf strings.Builder

        for _, source := range sources {
-               for _, n := range r.notifications[source] {
+               for _, n := range notifications[source] {
                        label, lcolor := levelLabel(n.Type)

                        // Top border with level.
```

marshallford:
> I assume you already are consuming ingress2gateway as a library?

Correct -- consuming it as a module and found this to be the missing piece for programmatic access to notifications.

> Since we now have Notifications() as a way to get notifications safely, we might as well use it in Render() instead of having bespoke retrieval logic in Render():

Ah, yes -- nice. Implemented.

johananl:
/lgtm

@LiorLieberman WDYT?

marshallford:
@johananl Thoughts?

johananl:
@marshallford I've already reviewed and gave an LGTM. We need another review and an approval.
Maybe @robscott or @kkk777-7 ?

youngnick:
The approach (passing a typed object rather than having to parse a string) makes total sense to me, but I haven't looked at the ingress2gateway code for a while, so I will leave the actual approval for one of the ingress2gateway approvers.

LiorLieberman:
Sorry for the delay - did not get notifications on it.
Thanks @marshallford @johananl !
/lgtm
/approve

PR #412, opened 2026-06-15
fix regex rewrite warning

Stevenjin8:
When users have regex rewrites configured, we should actually warn them that htis is not supported.


Stevenjin8:
@johananl ptal

johananl:
/hold
/lgtm

Holding to allow others to review. Feel free to unhold at your discretion.

PR #415, opened 2026-06-30
1.1.1 -> 1.2.0 changelog

Stevenjin8:
We should do a minor release instead of a patch since we added airlock microgateway support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


Stevenjin8:
@tyxeron If you want to polish the release note/changelog

tyxeron:
@Stevenjin8 Maybe it would be more precise to say 'Added airlock-micrograteway emitter support'?

